PR #11, opened 2020-02-25
UIDS-10 remove prepare script and precompile lib files

### LINKED ISSUE #10: Fix installation issue resulting from prepare script
Having the `prepare` script run on installation causes triggers this yarn [bug](https://github.com/yarnpkg/yarn/issues/6312). Pre-compiling the lib folder will allow us to remove that script.

davidmferris:
Closes #10 

Running the `prepare` script on installation triggers this yarn [bug](https://github.com/yarnpkg/yarn/issues/6312). Precompiling the lib files will make it so we can install this package without error.


.gitignore
```diff
@@ -5,9 +5,6 @@
 /.pnp
 .pnp.js
 
-# distribution
-lib
```
rsaris: I assume we already ignore this in eslint?

PR #17, opened 2020-03-07
UIDS-16 create Form and related components

### LINKED ISSUE #16: Add Form and FormGroup components
In an effort to standardize our form styles, we should add a Form and FormGroup component which can be used to import the styles outlined in `app/javascript/common/layout/bootstrap_extensions/ui_form.scss`. Hopefully these can handle a lot of what Simple Form is doing on the FE and break out the label, input, and errors into sane defaults.

Some things to note:

* We currently use these styles in ERB files directly (in `app/views/participant_populations/opt_in/participants/new.html.erb` as well as expanded use in https://github.com/user-interviews/rails-server/pull/2352) so we should ensure we have the ability to use these styles outside of these React components (namely ensuring they are in the default pack that we import).
* We currently use these styles in `app/javascript/researcher/opt_in_forms/form/form.jsx` so we may want to look into the structure there and update that to use these new components.
* Additionally -- we should cutover the use of the old `FormGroup` components in `app/javascript/researcher/saml/saml_configuration_form.jsx` to use these new components. Hopefully we can keep the API reasonably similar and this should be a good experiment into how much of a cutover other usages of the old form / form group will be.

davidmferris:
Closes #16 

I created some stories for these form components. If you run `yarn storybook` you can see them in the interactive storybook explorer. It's kind of a hodgepodge of the API from the old `FormGroup` component and the newer styles of the opt in form.

Screen Shot 2020-03-06 at 4 11 11 PM
Screen Shot 2020-03-06 at 4 11 19 PM


lib/Form/Form.js
```diff
@@ -0,0 +1,27 @@
+"use strict";
```
rsaris: I wonder if there is a way to get this to be compiled after merge to ensure it's always up to date and not in the diff.

src/Form/Form.jsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+import './Form.scss';
+
+export default function Form(props) {
+  return (
+    <form
+      className={classNames('Form', props.inline ? 'inline' : '')}
```
rsaris: Should this be:
```suggestion
      className={classNames('Form', props.inline ? 'Form--inline' : '')}
```
Also, using classNames this can be:
```suggestion
      className={classNames('Form', { 'Form--inline': props.inline })}
```

src/FormControl/FormControl.jsx
```diff
@@ -0,0 +1,76 @@
+import React, { Children, cloneElement } from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+import './FormControl.scss';
```
rsaris: Probably a bigger discussion but do we want to use PascalCase in this or snake_case like we do in Rails?

src/FormControl/FormControl.jsx
```diff
@@ -0,0 +1,76 @@
...
+    </ol>
+  );
+}
+
+export default function FormControl(props) {
+  const { errors, inputKey } = props;
+  const hasErrors = errors[inputKey] && errors[inputKey].length > 0;
+
+  return (
+    <div className={classNames('FormControl', props.inputClassName)}>
```
rsaris: The goal if this form was to mimic the markup we did in the opt in form, which does not include this div, it just has inputs and labels as siblings. The one thing is that the inputs do have the `form-control` class attached by default, which we may want to do, though we can also require that the user of this declare that. just so we don't get too cute.

src/FormControl/FormControl.jsx
```diff
@@ -0,0 +1,76 @@
+import React, { Children, cloneElement } from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+import './FormControl.scss';
+
+function addErrorClass(child) {
```
rsaris: I wonder if we should just update the styles to have this on the `FormGroup` so we don't have to do this crazy logic, I would require a bit more styling vs. using Bootstrap defaults but I'm kinda OK with that?

src/FormControl/FormControl.jsx
```diff
@@ -0,0 +1,76 @@
...
+      }
+      {
+        (props.displayErrorText && hasErrors) && (
+          <div className="invalid-feedback">
+            {renderErrors(errors[inputKey])}
+          </div>
+        )
+      }
+      {
+        props.description && (
```
rsaris: I think we can skip description now, it's still up in the air in https://github.com/user-interviews/rails-server/pull/2352 how we want this to work and I don't think we use it anywhere.

src/FormLabel/FormLabel.jsx
```diff
@@ -0,0 +1,40 @@
...
+    labelHtmlFor,
+    legend,
+    required,
+    text,
+  } = props;
+
+  const label = labelHtmlFor ? <label htmlFor={labelHtmlFor}>{text}</label> : <span>{text}</span>;
+
+  return (
+    <div className={classNames('FormLabel', className)}>
```
rsaris: Similar with above -- I think the goal was to get rid of this parent div as we no longer need it with the newer styles. Similarly we can likely get rid of the legend prop here as we'll obviously have to really rethink the styles if we ever bring that over.

src/FormLabel/FormLabel.jsx
```diff
@@ -0,0 +1,40 @@
...
+export default function FormLabel(props) {
+  const {
+    className,
+    labelHtmlFor,
+    legend,
+    required,
+    text,
+  } = props;
+
+  const label = labelHtmlFor ? <label htmlFor={labelHtmlFor}>{text}</label> : <span>{text}</span>;
```
rsaris: Could we default these divs to have the class name `FormLabel` to apply the styles we want directly?

stories/FormGroup.stories.jsx
```diff
@@ -0,0 +1,62 @@
...
+
+const withPadding = (story) => <div style={{ padding: '1rem' }}>{story()}</div>;
+
+export default {
+  title: 'Design System/Form Group',
+  component: FormGroup,
+  decorators: [withA11y, withKnobs, withPadding],
+};
+
+export const Default = () => (
```
rsaris: Honestly -- the stories in this line up with my assumption as to how we'd organize these. I don't know that we actually need `FormControl` other than maybe to deal with the errors, but maybe even `FormGroup` could handle that?

lib/Form/Form.js
davidmferris: I think the preferred way is to compile this in a post-install script for that very reason. I had it set up that way originally but ran into some issues with that that seemed related to this being hosted on github instead of NPM. I can't remember if that was before we switched this to a public package so I'll play around with it and see if I can change that back.

src/FormControl/FormControl.jsx
davidmferris: I think I might have started it this way when I was following the Storybook tutorial and that's how they had it. TBH I kind of prefer the pascal case when dealing with components, which is how I've seen it the majority of the time. It's all personal preference though so I'm not married to it either way.

stories/FormGroup.stories.jsx
davidmferris: Yeah good call. I think it makes sense for `FormGroup` to handle the errors which would mean we could ditch `FormControl` altogether.

lib/Form/Form.js
davidmferris: Ok I think I got this working using the `prepublishOnly` script. It allows us to compile the distribution files right before we pack the archive so we won't have to check these lib files into git anymore 🎉

.gitignore
```diff
@@ -5,6 +5,9 @@
 /.pnp
 .pnp.js
 
+# distribution
+lib
```
rsaris: :fire:

src/FormControl/FormControl.jsx
rsaris: Makes sense. This is a JS library so let's keep it with JS conventions, especially since these file names don't really leak out so it's pretty contained to this library.

scss/typography.scss
```diff
@@ -1,3 +1,36 @@
 $font-family: Lato, Arial, sans-serif;
+$font-weight-light: 200;
+$font-weight-normal: 300;
 $font-size-base: .875rem;
 $font-weight-bold: 700;
+$headings-font-weight: 400;
+$headings-font-weight-bold: 700;
+
+@mixin font-type-100() {
```
rsaris: Not for this PR but going forward -- we likely want to bring in the new typography soon and start using that (and updating whatever styles we have to only use those) so we're not creating more work for us in the future w/ this.

src/FormGroup/FormGroup.scss
```diff
@@ -0,0 +1,21 @@
+@import '../../scss/theme';
+
+.FormGroup {
+  margin: 1rem 0 0.5rem;
+
+  &--is-invalid {
+    input, textarea {
```
rsaris: Could we include `select` on this too?

src/FormLabel/FormLabel.jsx
```diff
@@ -0,0 +1,30 @@
...
+
+export default function FormLabel(props) {
+  const {
+    labelHtmlFor,
+    required,
+    text,
+  } = props;
+
+  return (
+    <label className="FormLabel" htmlFor={labelHtmlFor}>
```
rsaris: `FormGroup` provides a className prop here right? Might as well pass it through.

src/FormLabel/FormLabel.jsx
```diff
@@ -0,0 +1,30 @@
...
+      {text}
+      {required && <span className="helper-text">&nbsp;(required)</span>}
```
rsaris: ??
```suggestion
      {required && <span className="FormLabel__helper-text">&nbsp;(required)</span>}
```

src/FormLabel/FormLabel.scss
```diff
@@ -0,0 +1,15 @@
...
+.FormLabel {
+  @include font-type-200;
+  font-weight: $headings-font-weight;
+  margin-bottom: 0.5rem;
+
+  .helper-text {
+    color: $ux-gray-400;
+  }
+
+  label {
```
rsaris: Do we need this? This would seem to style a label in the label, which seems off.

davidmferris: Good catch. I think this is a holdover from a previous iteration that no longer applies.

PR #19, opened 2020-03-11
UIDS 18 react select wrappers

### LINKED ISSUE #18: Add ReactSelect wrappers with our styles

rsaris:
This brings over the SingleSelect and AsyncSelect components referenced in https://github.com/user-interviews/rails-server/pull/2338/files

These should provide a good place for common styles / interactions within our usage of React Select. There was / is a bit of debate about declaring all props sent to ReactSelect but I think with Fullstory it definitely makes sense in this repo to be explicit about what we're passing through. That said, it may be worth a follow on task to really think through this vs. just including the ones I happened to need for these usages.

Something else to note -- this uses React Select v3 while our repo uses v2. The truth is the upgrade is relatively simple (the breaking change was the import path).

image

Closes #18


src/Select/AsyncSelect.jsx
```diff
@@ -0,0 +1,86 @@
...
+    isLoading={isLoading}
+    loadOptions={loadOptions}
+    menuPortalTarget={modal ? document.body : undefined}
+    name={name}
+    noOptionsMessage={noOptionsMessage}
+    placeholder={placeholder}
+    shouldShowValue
+    styles={{
+      ...defaultStyles,
+      menuPortal: (base) => (
```
rsaris: I'm not totally in love with this implementation so happy to take criticism here but I do think it's valuable to include the gains we made here: https://github.com/user-interviews/rails-server/pull/2413

src/Styles/colors.js
```diff
@@ -0,0 +1,28 @@
+// this can probably be generated... try to group the colors spatially plz
```
rsaris: Not sure about this `src/Styles` folder name or the fact that these files are camelCase vs. PascalCased. It was a little weird to PascalCase ZStack so went with camelCase.

src/Select/AsyncSelect.jsx
rsaris: I've thought on this a bit and I do think that this is a bit of a different case. The goal here (and w/ the `menuPortalTarget`) is really to accommodate the weird interaction we were seeing with modals where it wouldn't extend outside of the modal. I could see moving to allowing for outside styles but TBH my goal w/ this design system is to make it so we never pass in custom styles or anything and instead using props / CSS class variants keep all styling in this repo.

src/Select/SingleSelect.jsx
```diff
@@ -0,0 +1,69 @@
...
+  value,
+
+  onChange,
+}) => (
+  <Select
+    className={`${className || ''} SingleSelect`}
+    defaultValue={defaultValue}
+    isDisabled={isLoading}
+    isSearchable={isSearchable}
+    menuPortalTarget={modal ? document.body : undefined}
```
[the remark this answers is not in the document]
rsaris: See note above on why I think this specific piece of the implementation is worth keeping as it is.

package.json
```diff
@@ -15,6 +15,7 @@
     "react-transition-group": "^4.3.0",
     "react-copy-to-clipboard": "^5.0.2",
     "react-popper": "^1.3.7",
+    "react-select": "^3.0.8",
```
davidmferris: Should this be a `peerDependency`?

src/Select/AsyncSelect.jsx
davidmferris: Fair enough. That makes sense to me.

src/index.js
```diff
@@ -12,9 +12,12 @@ import Pill from './Pill/Pill';
 import Popper from './Popper/Popper';
 import RadioButton from './RadioButton/RadioButton';
 import RadioButtonGroup from './RadioButtonGroup/RadioButtonGroup';
+import AsyncSelect from './Select/AsyncSelect';
+import SingleSelect from './Select/SingleSelect';
```
davidmferris: Any thoughts on adding an index.js file to /Select and importing these in one statement like we do with `Flash`? I'm not sure on where I stand on this but I wonder if we should add them to all of the component directories, even if they just have one component. Also curious to hear @rroppolo's thoughts

package.json
rsaris: So we actually use a different version (2.x) in the Rails app. I was hoping to update to that once we get this library in and then can circle back?

src/index.js
rroppolo: Hmm I do like the consistency of everything having an index..  I was also wondering if at some point in the future, if the design system gets quite large, we may want to categorize these components further.  Like foundations/elements/modules or something to indicate whether it is a building block or a component made up of other components.  Have you guys thought about doing something like that, or other ways of categorizing?

davidmferris: I think that's a good thing to think about. TBH I had only really thought about housing the building blocks in here but I'm sure that will change as this library grows, so maybe something to start working on now while it's a small library.

src/Select/SingleSelect.jsx
```diff
@@ -0,0 +1,86 @@
...
+  onChange,
+}) => (
+  <Select
+    aria-label={ariaLabel}
+    aria-labelledby={ariaLabelledBy}
+    className={`${className || ''} SingleSelect`}
+    defaultValue={defaultValue}
+    id={id}
+    isDisabled={disabled || isLoading}
+    isSearchable={isSearchable}
```
rroppolo: Is it worth including isClearable ?  Right now I can't unset something once I set it.  For example to clear out an optional field.

rsaris: Added!

.eslintrc
```diff
@@ -5,13 +5,13 @@
       "files": [
         "stories/*.{js,jsx}"
       ],
-      rules: {
```
rsaris, line 8: @SMaxOwok I know you made a lot of these changes in your PR too so IT'S A RACE!

src/Styles/palette.js
```diff
@@ -1,6 +1,9 @@
 // This can probably be generated... try to group the colors spatially plz
 
 export default {
+  inputBorderColor: '#CED4DA',
+  inputDisabledBg: '#E9ECEF',
+
```
rsaris, line 6: I'm not how I feel about these but Jeff added them to the Rails app and I'd rather get this up and running sooner rather than later.

stories/Palette.stories.jsx
```diff
@@ -4,14 +4,16 @@ import palette from '../src/Styles/palette';
 
 import { withPadding } from './decorators';
 
-const renderColor = (colorName) => (
+const renderColor = (colorName) => () => (
```
rsaris, line 7: Apparently the decorator gets upset if you don't pass a function so had to make this curry.

scss/palette.scss
```diff
@@ -66,3 +66,6 @@ $ux-red-600: #C71F1F;
 $ux-red-700: #B11B1B;
 $ux-red-800: #9B1818;
 $ux-red-900: #851414;
+
+$input-border-color: $ux-gray-400; // Override bootstrap default
+$input-disabled-bg: $ux-gray-300; // Override bootstrap default
```
rsaris, line 71: Talked with Jeff this morning and he convinced me that having semantic colors is super useful so using some of the variable names that Bootstrap has already. Let me know if this is controversial but I kinda like this.

rsaris: And then not use them in the palette? They're still colors so it makes sense for them to be here?

davidmferris: I was thinking like how we handle stuff like box shadow and buttons. They define their colors in their .scss files and then they're imported where needed. My thought around palette was this is where we'd define all of the `$ux-` token values and then we could implement them at the component level.

davidmferris: I could also see a bootstrap_overrides.scss file if we want to keep all of our overrides in one easy to find spot. That file could import palette.scss and theme could import.scss could import bootstrap_overrides.scss

src/Select/AsyncSelect.jsx
```diff
@@ -0,0 +1,108 @@
+import React from 'react';
+import propTypes from 'prop-types';
+import Async from 'react-select/async';
+
+import zStack from '../Styles/zStack';
```
SMaxOwok, line 5: Dogggg

rsaris: GOD DAMN IT!

PR #21, opened 2020-03-12
UIDS 20 Add RadioButton component

### LINKED ISSUE #20: Add Radio Button component

davidmferris:
Closes #20 

This takes some styling from the old "ui-form" CSS class and extracts it to the component level. Tested this on the opt-in page:

Screen Shot 2020-03-12 at 1 04 13 PM

### DIFF
--- a/src/FormGroup/FormGroup.jsx
+++ b/src/FormGroup/FormGroup.jsx
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 
-import FormLabel from '../FormLabel/FormLabel';
+import InputLabel from '../InputLabel/InputLabel';
 
 import './FormGroup.scss';
 
@@ -31,7 +31,7 @@ export default function FormGroup(props) {
       id={props.id}
     >
       {props.label && (
-        <FormLabel
+        <InputLabel
           className={props.labelClassName}
           labelHtmlFor={props.labelHtmlFor}
           required={props.required}
--- a/src/RadioButton/RadioButton.jsx
+++ b/src/RadioButton/RadioButton.jsx
@@ -1,0 +1,28 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export default function RadioButton(props) {
+  return (
+    <input
+      className="RadioButton"
+      disabled={props.disabled}
+      id={props.id}
+      name={props.name}
+      type="radio"
+      {...props.radioButtonProps}
+    />
+  );
+}
+
+RadioButton.propTypes = {
+  disabled: PropTypes.bool,
+  id: PropTypes.string.isRequired,
+  name: PropTypes.string,
+  radioButtonProps: PropTypes.object,
+};
+
+RadioButton.defaultProps = {
+  disabled: false,
+  name: '',
+  radioButtonProps: {},
+};
--- a/stories/Card.stories.jsx
+++ b/stories/Card.stories.jsx
@@ -7,10 +7,12 @@ import Card from '../src/Card/Card';
 import 'bootstrap/dist/css/bootstrap.min.css';
 import '../scss/global.scss';
 
+import { withPadding } from './decorators';
+
 export default {
   title: 'Design System/Card',
   component: Card,
-  decorators: [withA11y, withKnobs],
+  decorators: [withA11y, withPadding, withKnobs],
 };
 
 export const allCards = () => (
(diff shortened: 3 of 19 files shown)


src/RadioButton/RadioButton.jsx
```diff
@@ -0,0 +1,30 @@
...
+import PropTypes from 'prop-types';
+
+import './RadioButton.scss';
+
+export default function RadioButton(props) {
+  return (
+    <label className="RadioButton" htmlFor={props.htmlFor}>
+      <input
+        disabled={props.disabled}
+        id={props.id}
```
rsaris: Should we add a `name` prop here? I could also see adding a `radioButtonProps` which allows for passing whatever we want to the radio button input since I assume there are a lot that we'll want to pass through.

src/RadioButton/RadioButton.jsx
```diff
@@ -0,0 +1,30 @@
...
+        type="radio"
+      />
+
+      {props.label}
+    </label>
+  );
+}
+
+RadioButton.propTypes = {
```
rsaris: Probably excessive but thoughts on calling this `LabeledRadioButton`?

stories/RadioButton.stories.jsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import { withA11y } from '@storybook/addon-a11y';
+
+import RadioButton from '../src/RadioButton/RadioButton';
+
+const withPadding = (story) => <div style={{ padding: '1rem' }}>{story()}</div>;
```
rsaris: Wanna pull this into a lib somewhere (as per your request in another PR)?

stories/Form.stories.jsx
```diff
@@ -39,6 +40,10 @@ export const Default = () => (
     >
       <input className="form-control" id="phone" type="text" />
     </FormGroup>
+    <div className="radio">
```
rsaris: Should we make this a  standard component to use? Like `RadioButtonGroup`? I imagine the padding etc here may be meaningful.

src/RadioButton/RadioButton.jsx
davidmferris: Did you one better. I decoupled the label and `RadioButton` altogether and created a new component that takes a `control` prop (`RadioButton` in this case, but will work with checkbox and switch when we add those too). I modeled it off of how [MUI](https://material-ui.com/api/form-control-label/) does it. If we ever want to update the label placement like top, bottom, whatever this will be easy to extend.

stories/Form.stories.jsx
davidmferris: Absolutely.

src/FormGroup/FormGroup.jsx
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 
-import FormLabel from '../FormLabel/FormLabel';
+import InputLabel from '../InputLabel/InputLabel';
```
davidmferris: Renamed this for clarity's sake now that there is `FormControlLabel` as well.

rsaris: The beauty of this being behind the design system.

stories/Card.stories.jsx
```diff
@@ -7,10 +7,12 @@ import Card from '../src/Card/Card';
 import 'bootstrap/dist/css/bootstrap.min.css';
 import '../scss/global.scss';
 
+import { withPadding } from './decorators';
```
rsaris: :+1:

stories/FormControlLabel.stories.jsx
```diff
@@ -0,0 +1,23 @@
...
+
+export default {
+  title: 'Design System/Form Control Label',
+  component: FormControlLabel,
+  decorators: [withA11y, withPadding],
+};
+
+export const Radio = () => (
+  <FormControlLabel
+    control={
```
rsaris: Seeing this here, any reason we couldn't just use `children` for this? In general I feel like passing a component as a prop makes me sad and find using `children` looks / feels a lot better.

davidmferris: `children` is fine. FWIW I do think there are valid reasons for wanting to pass a rendered component as a prop but probably not necessary here.

rsaris: I can see it though I think when you can mimic the normal HTML flow by using `children` (like we're doing here w/ labels and buttons) it makes more sense to do so.

Issue #24, opened 2020-03-21
Reusable flash message functionality


rroppolo:
Create new React hook for flash message functionality. 
Simplify state handling in a reusable way.
Create a pass-through hoc to use flash hook in class components

PR #25, opened 2020-03-23
UIDS-24 Flash message components

rroppolo:
For https://github.com/user-interviews/ui-design-system/issues/24
Move flash message component into design system and include hook (and pass through HOC that uses the hook too) for better reusability in app components.

### DIFF
--- a/scss/theme.scss
+++ b/scss/theme.scss
@@ -1,7 +1,6 @@
+@import './box_shadow';
 @import './lists';
+@import './navbar';
 @import './palette';
 @import './typography';
 @import './z_stack';
-
-$ui-navbar-height: 3rem;
-$ui-navbar-height-mobile: $ui-navbar-height;
--- a/scss/theme.scss
+++ b/scss/theme.scss
@@ -1,3 +1,6 @@
 @import './lists';
 @import './palette';
 @import './typography';
+
+$ui-navbar-height: 3rem;
+$ui-navbar-height-mobile: $ui-navbar-height;
--- a/spec/Flash/withFlash.test.jsx
+++ b/spec/Flash/withFlash.test.jsx
@@ -1,4 +1,4 @@
-import React, { Component } from 'react';
+import React from 'react';
 import { act } from '@testing-library/react-hooks';
 import { create } from 'react-test-renderer';
 
@@ -12,12 +12,7 @@ jest.mock('react-transition-group', () => (
   }
 ));
 
-// eslint-disable-next-line react/prefer-stateless-function
-class WrappedComponent extends Component {
-  render() {
-    return <div />;
-  }
-}
+const WrappedComponent = () => <div />;
 
 describe('test withFlash', () => {
   test('it can create a new flash message', async () => {
@@ -25,7 +20,6 @@ describe('test withFlash', () => {
     const ComponentWithFlash = withFlash(WrappedComponent);
     const flash = create(<ComponentWithFlash />);
     const component = flash.root.findByType(WrappedComponent);
-
 
     act(() => {
       component.props.setFlashMessage(MessageTypes.SUCCESS, newMessage);
--- a/spec/Flash/withFlash.test.jsx
+++ b/spec/Flash/withFlash.test.jsx
@@ -1,0 +1,36 @@
+import React, { Component } from 'react';
+import { act } from '@testing-library/react-hooks';
+import { create } from 'react-test-renderer';
+
+import withFlash from '../../src/Flash/withFlash';
+import { MessageTypes } from '../../src/Flash';
+
+jest.mock('react-transition-group', () => (
+  {
+    CSSTransition: ({ children }) => children,
+    TransitionGroup: ({ children }) => children,
+  }
+));
+
+// eslint-disable-next-line react/prefer-stateless-function
+class WrappedComponent extends Component {
+  render() {
+    return <div />;
+  }
+}
+
+describe('test withFlash', () => {
+  test('it can create a new flash message', async () => {
+    const newMessage = 'This is just a test...';
+    const ComponentWithFlash = withFlash(WrappedComponent);
+    const flash = create(<ComponentWithFlash />);
+    const component = flash.root.findByType(WrappedComponent);
+
+
+    act(() => {
+      component.props.setFlashMessage(MessageTypes.SUCCESS, newMessage);
+    });
+
+    expect(flash).toMatchSnapshot();
+  });
+});
--- a/src/FadeTransition/FadeTransition.scss
+++ b/src/FadeTransition/FadeTransition.scss
@@ -1,16 +1,16 @@
-.fade-transition-enter {
+.FadeTransition-enter {
   opacity: 0.01;
 
-  &.fade-transition-enter-active {
+  &.FadeTransition-enter-active {
     opacity: 1;
     transition: opacity 300ms ease-in;
   }
 }
 
-.fade-transition-exit {
+.FadeTransition-exit {
   opacity: 1;
 
-  &.fade-transition-exit-active {
+  &.FadeTransition-exit-active {
     opacity: 0.01;
     transition: opacity 200ms ease-in;
   }
--- a/src/Flash/Flash.scss
+++ b/src/Flash/Flash.scss
@@ -1,0 +1,54 @@
+@import '~bootstrap/scss/functions';
+@import '~bootstrap/scss/variables';
+@import '~bootstrap/scss/mixins';
+@import '~bootstrap/scss/alert';
+@import '../../scss/theme';
+
+$top-padding: 1.25rem;
+// Flash colors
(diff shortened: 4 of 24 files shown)


src/Flash/Flash.scss
```diff
@@ -0,0 +1,54 @@
...
+
+$ux-box-shadow-card: 0 2px 4px rgba(0,0,0,0.1);
+
+$ux-box-shadow-top: 0 -2px 5px $ux-navbar-shadow-color;
+$ux-box-shadow-top-light: 0 -1px 1px $ux-navbar-shadow-color;
+
+// This zindex is between the header (zindex-fixed) and the modal backdrop (zindex-modal-backdrop)
+$z-index-flash: 1045;
+
+.flash {
```
rsaris: So one thing we're trying to do with this library is to make the classes PascalCased like the component they're used on, so this would be `Flash`. (Note to self to update the change I just made in the codebase to reflect this).

src/Flash/Flash.scss
```diff
@@ -0,0 +1,54 @@
...
+$ux-box-shadow: 0 2px 5px $ux-navbar-shadow-color;
+$ux-box-shadow-light: 0 1px 1px $ux-navbar-shadow-color;
+
+$ux-box-shadow-card: 0 2px 4px rgba(0,0,0,0.1);
+
+$ux-box-shadow-top: 0 -2px 5px $ux-navbar-shadow-color;
+$ux-box-shadow-top-light: 0 -1px 1px $ux-navbar-shadow-color;
+
+// This zindex is between the header (zindex-fixed) and the modal backdrop (zindex-modal-backdrop)
+$z-index-flash: 1045;
```
rsaris: This should already exist in this repo.

src/Flash/Flash.scss
```diff
@@ -0,0 +1,54 @@
...
+@import '~bootstrap/scss/mixins';
+@import '~bootstrap/scss/alert';
+@import '../../scss/theme';
+
+$top-padding: 1.25rem;
+// Flash colors
+$ux-flash-error: $ux-red-100;
+$ux-flash-error-border: $ux-red-200;
+
+$ux-navbar-shadow-color: rgba( 0, 0, 0, 0.26 );
```
rsaris: I would expect all of this to live in a separate file, maybe `scss/box_shadow.scss`?

src/Flash/Flash.scss
```diff
@@ -0,0 +1,54 @@
+@import '~bootstrap/scss/functions';
```
rsaris: I wonder how hard it would be to just steal the styles and build our own styles vs. relying on bootstrap here.

src/Flash/fade_transition/FadeTransition.scss
```diff
@@ -0,0 +1,17 @@
+.fade-transition-enter {
```
rsaris: Is this worth having in the styles directory? I wonder how much this is specific to this vs. could be made generic. Again, assuming the `fade-transition` is because that's what we put in, let's see about making it `FadeTransition`.

src/Flash/useFlash.js
```diff
@@ -0,0 +1,38 @@
...
+  };
+
+  const dismissMessage = (messageId) => {
+    setMessages(
+      messages.filter((message) => message.id !== messageId),
+    );
+  };
+
+  return {
+    messages,
```
rsaris: These could be alphabetized

src/Flash/withFlash.jsx
```diff
@@ -0,0 +1,29 @@
...
+export const withFlashPropTypes = {
+  setFlashMessage: PropTypes.func,
+};
+
+export default function withFlash(WrappedComponent) {
+  return (props) => {
+    const { messages, setMessage, dismissMessage } = useFlash();
+
+    return (
+      <>
```
rsaris: Is this just shorthand for Fragment? I have seen this places but never understood it.

scss/theme.scss
```diff
@@ -1,3 +1,6 @@
 @import './lists';
 @import './palette';
 @import './typography';
+
+$ui-navbar-height: 3rem;
```
rsaris: I wonder if this should be in it's own file? Maybe `navbar`?

src/Flash/Flash.scss
davidmferris: I don't think that's been merged yet. [PR](https://github.com/user-interviews/ui-design-system/pull/19/files)

rsaris: Shoot... let me open a separate PR to do that then.

rsaris: PR here: https://github.com/user-interviews/ui-design-system/pull/26

rsaris: @rroppolo this has been merged if you wanna merge or rebase.

spec/Flash/withFlash.test.jsx
```diff
@@ -0,0 +1,36 @@
...
+}
+
+describe('test withFlash', () => {
+  test('it can create a new flash message', async () => {
+    const newMessage = 'This is just a test...';
+    const ComponentWithFlash = withFlash(WrappedComponent);
+    const flash = create(<ComponentWithFlash />);
+    const component = flash.root.findByType(WrappedComponent);
+
+
```
davidmferris: Think you got an extra line in here.

src/Flash/FlashMessage.jsx
```diff
@@ -0,0 +1,40 @@
...
+  if (!Object.values(MessageTypes).includes(type)) {
+    throw new TypeError(`Unexpected type ${type} used for a flash message.`);
+  }
+
+  return `alert alert-${type}`;
+};
+
+export default function FlashMessage(props) {
+  const handleDismissMessage = () => props.onDismiss(props.id);
+  return (
```
davidmferris: Style suggestion: add a blank line between `const` and `return`?

stories/Flash.stories.jsx
```diff
@@ -0,0 +1,32 @@
...
+    <p>Click the button to see a flash message.  Use the knobs to try different types!</p>
+    <button className="btn btn-primary" type="button" onClick={() => setFlashMessage(type, message)}>Submit</button>
+  </div>
+  );
+DummyComponent.propTypes = withFlashPropTypes;
+DummyComponent = withFlash(DummyComponent);
+
+export const flashStory = () => (
+  <DummyComponent
+    type={radios('Message Type', MessageTypes, MessageTypes.SUCCESS)}
```
davidmferris: Awesome, very nice setup.

stories/Flash.stories.jsx
```diff
@@ -0,0 +1,32 @@
...
+let DummyComponent = ({ type, message, setFlashMessage }) => (
+  <div>
+    <p>Click the button to see a flash message.  Use the knobs to try different types!</p>
+    <button className="btn btn-primary" type="button" onClick={() => setFlashMessage(type, message)}>Submit</button>
+  </div>
+  );
+DummyComponent.propTypes = withFlashPropTypes;
+DummyComponent = withFlash(DummyComponent);
+
+export const flashStory = () => (
```
davidmferris: Minor suggestion but could we call this `FlashMessage`? Then you can remove L24-26. I've been capitalizing these story names, though don't know if that's best practice or not.

src/Flash/withFlash.jsx
davidmferris: Hrmm, we can probably disable that rule too. I do like the `<>` syntax but probably makes sense to keep our lint rules consistent between the apps.

src/Flash/Flash.scss
rroppolo: Forgot to document slack discussion I had with Bob here:  we will keep this for now as importing only some bootstrap (from the 'required' variables, mixins and functions) is a hairy path that I did try and go down for a bit.  For now leaving as is.
Some other options for this problem could be to not use bootstrap classes at all in the design system, or use the bootstrap classes but don't override any of the styles here (i.e. do it in rails-server instead).

src/FadeTransition/FadeTransition.scss
```diff
@@ -0,0 +1,17 @@
...
+  &.FadeTransition-enter-active {
+    opacity: 1;
+    transition: opacity 300ms ease-in;
+  }
+}
+
+.FadeTransition-exit {
+  opacity: 1;
+
+  &.FadeTransition-exit-active {
```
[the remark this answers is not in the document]
rsaris: So these classes are added by the FadeTransition that we're using. I do desperately wish they used BEM but they do not. 

That said, may be worth a comment.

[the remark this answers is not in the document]
davidmferris: Ah, just one more dash!

rsaris:
Sorry this review took so long -- @davidmferris and @rroppolo do you want to get a release together and get this working in the Rails app?

PR #28, opened 2020-03-23
UIDS-27 Add CopyToClipboard and TrackedButton components

### LINKED ISSUE #27: Add CopyToClipboard component

davidmferris:
Closes #27 

Screen Shot 2020-03-23 at 4 05 59 PM

### DIFF
--- a/src/CopyToClipboard/CopyToClipboard.jsx
+++ b/src/CopyToClipboard/CopyToClipboard.jsx
@@ -1,0 +1,51 @@
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faCopy } from '@fortawesome/free-regular-svg-icons';
+import { CopyToClipboard as ReactCopyToClipboard } from 'react-copy-to-clipboard';
+
+import Popper from '../Popper/Popper';
+import TrackedButton from '../TrackedButton/TrackedButton';
+
+import './CopyToClipboard.scss';
+
+function CopyToClipboard(props) {
+  const [copied, setCopied] = useState(false);
+
+  const handleClickCopy = () => {
+    setCopied(true);
+    setTimeout(() => setCopied(false), 1000);
+  };
+
+  return (
+    <div className="CopyToClipboard">
+      <span className="CopyToClipboard__copy-text">{props.copyText}</span>
+      <ReactCopyToClipboard text={props.copyText} onCopy={handleClickCopy}>
+        <TrackedButton
+          className="btn btn-link btn-link--neutral"
+          event={props.trackingEvent}
+          type="button"
+        >
+          <Popper
+            dark
+            text="Copied!"
+            visible={copied}
+          >
+            <FontAwesomeIcon icon={faCopy} />
+          </Popper>
+        </TrackedButton>
+      </ReactCopyToClipboard>
+    </div>
+  );
+}
+
+CopyToClipboard.propTypes = {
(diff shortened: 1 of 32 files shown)


src/hoc/withTrackedClick.jsx
```diff
@@ -0,0 +1,38 @@
...
+
+    function handleClick(clickEvent) {
+      props.tracking.trackEvent({ ...eventData, event });
+
+      if (props.onClick) {
+        props.onClick(clickEvent);
+      }
+    }
+
+    return <Target {...rest} onClick={handleClick} />;
```
davidmferris: @rsaris - the current implementation passes the `event` prop to the target component too. Is that actually needed?

stories/CopyToClipboard.stories.jsx
```diff
@@ -0,0 +1,20 @@
...
+import { withKnobs, text } from '@storybook/addon-knobs';
+
+import CopyToClipboard from '../src/CopyToClipboard/CopyToClipboard';
+
+import { withPadding } from './decorators';
+
+export default {
+  title: 'Design System/Copy To Clipboard',
+  component: CopyToClipboard,
+  decorators: [withA11y, withKnobs, withPadding],
```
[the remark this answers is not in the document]
davidmferris: Nice catch! I update this to use `escapeHtml: false`

scss/popper.scss
```diff
@@ -0,0 +1,19 @@
+@import './borders';
+@import './box_shadow';
+@import './palette';
+@import './typography';
+
+.ui-popper {
```
rsaris: I'm kind into the fact that these files don't have any CSS classes (and thus can be imported infinite times without adjusting the bundle size). Is there any reason this couldn't be shipped with a `Popper` component?

If we do want to have classes, we may think about moving to a similar system to the one we have in the Rails app where files with the _ prefix mean "This has no CSS classes, so you can import as needed" while other files should only be imported at the top of the tree.

src/hoc/withTrackedClick.jsx
rsaris: It is not, I just didn't think to remove it.

src/CopyToClipboard/CopyToClipboard.jsx
```diff
@@ -0,0 +1,71 @@
...
+        <TrackedButton
+          className="btn btn-link btn-link--neutral"
+          event={props.trackingEvent}
+          type="button"
+        >
+          <Manager>
+            <Reference>
+              {({ ref }) => (
+                <div ref={ref}>
+                  <FontAwesomeIcon className="icon-right" icon={faCopy} />
```
rsaris: Do we have `icon-right` in this repo?

davidmferris: Ah nope. I thought that was something that `FontAwesomeIcon` handled so this is effectively not doing anything. FWIW even without the margin that `icon-right` would provide, I think the spacing looks good:

Screen Shot 2020-03-25 at 9 11 30 AM

scss/popper.scss
davidmferris: I added a `Popper` component because I think that'd be useful for things other than `CopyToClipboard` and I agree with your point about not adding CSS classes to these.

src/CopyToClipboard/CopyToClipboard.jsx
rsaris: Yup -- we added icon-right and icon-left. Unclear if we even always have access to it as we were expecting it to only be used in buttons.

PR #30, opened 2020-03-27
UIDS 29 Add all the colors and export them.

rsaris:
Also included exporting FadeTransition.

### DIFF
--- a/scss/palette.scss
+++ b/scss/palette.scss
@@ -2,7 +2,6 @@
 
 $ux-black: #000000;
 $ux-blue: #337ab7;
-$ux-dark-blue: #639BB8;
 $ux-gray: #A1A1A1;
 $ux-green: #6DBD63;
 $ux-light-blue: #7CCBF2;
--- a/src/index.js
+++ b/src/index.js
@@ -1,5 +1,6 @@
 import Card from './Card/Card';
 import CopyToClipboard from './CopyToClipboard/CopyToClipboard';
+import FadeTransition from './FadeTransition/FadeTransition';
 import {
  Flash, MessageTypes, withFlash, withFlashPropTypes, useFlash,
 } from './Flash';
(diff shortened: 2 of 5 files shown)


scss/palette.scss
```diff
@@ -2,7 +2,6 @@
 
 $ux-black: #000000;
 $ux-blue: #337ab7;
-$ux-dark-blue: #639BB8;
```
rsaris: We don't use this anywhere so removed it.

stories/Palette.stories.jsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+
+import palette from '../src/Styles/palette';
+
+import { withPadding } from './decorators';
+
+const renderColor = (colorName) => (
+  <div style={{ display: 'flex', height: '20rem', width: '100%' }}>
```
rsaris: Not sure how we feel about using styles like this in stories but I figured it was better than trying to interject scss files into the stories directory.

src/index.js
```diff
@@ -11,17 +12,21 @@ import Pill from './Pill/Pill';
 import Popper from './Popper/Popper';
 import RadioButton from './RadioButton/RadioButton';
 import RadioButtonGroup from './RadioButtonGroup/RadioButtonGroup';
+import palette from './Styles/palette';
```
davidmferris: Wondering if we should make a /Styles/index.js file and import like `import { palette, zStack } from 'ui-design-system/Styles` to differentiate it from components. What do you think? FYI as of v1.1.0 the /Styles directory is included in the package lib

stories/Palette.stories.jsx
davidmferris: This is cool with me. Do you think it'd be helpful to add the color names (e.g. `uxBlue200`) somewhere on or near each swath for anyone who needs to quickly cross reference?

src/index.js
rsaris: I'm into it.

stories/Palette.stories.jsx
rsaris: DONE!

PR #32, opened 2020-03-31
[FEAT] Add new props to radio button

SMaxOwok:
- Adds labels
- Adds children
- Adds styles

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -5,13 +5,13 @@
       "files": [
         "stories/*.{js,jsx}"
       ],
-      rules: {
-        "import/no-extraneous-dependencies": ["error", { "devDependencies": true }],
-      },
+      "rules": {
+        "import/no-extraneous-dependencies": ["error", { "devDependencies": true }]
+      }
     },
     {
       "files": [
-        "src/**/*.{js,jsx}",
+        "src/**/*.{js,jsx}"
       ],
       "env": {
         "browser": true
@@ -103,9 +103,16 @@
         "react-hooks/exhaustive-deps": "warn",
         "react-hooks/rules-of-hooks": "error",
         "semi": 0
-      },
+      }
     }
   ],
+  "settings": {
+    "import/resolver": {
+      "node": {
+        "paths": ["./"]
+      }
+    }
+  },
   "rules": {
     "class-methods-use-this": 0,
     "function-paren-newline": ["error", "consistent"],
--- a/src/RadioButton/RadioButton.jsx
+++ b/src/RadioButton/RadioButton.jsx
@@ -23,12 +23,19 @@ export default function RadioButton(props) {
       {props.label && (
         <span className="RadioButton__label">{props.label}</span>
       )}
+
+      {props.children && (
+        <div className="RadioButton__children">
+          {props.children}
+        </div>
+      )}
     </label>
   );
 }
 
 RadioButton.propTypes = {
   bordered: PropTypes.bool,
+  children: PropTypes.node,
   disabled: PropTypes.bool,
   id: PropTypes.string.isRequired,
   label: PropTypes.string.isRequired,
@@ -38,6 +45,7 @@ RadioButton.propTypes = {
 
 RadioButton.defaultProps = {
   bordered: false,
+  children: null,
   disabled: false,
   name: '',
   radioButtonProps: {},
--- a/src/RadioButton/RadioButton.jsx
+++ b/src/RadioButton/RadioButton.jsx
@@ -1,27 +1,43 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import classnames from 'classnames';
+
+import './RadioButton.scss';
 
 export default function RadioButton(props) {
+  const classNames = classnames('RadioButton', {
+    'RadioButton--bordered': props.bordered,
+  });
+
   return (
-    <input
-      className="RadioButton"
-      disabled={props.disabled}
-      id={props.id}
-      name={props.name}
-      type="radio"
-      {...props.radioButtonProps}
-    />
+    <label className={classNames}>
+      <input
+        className="RadioButton__input"
+        disabled={props.disabled}
+        id={props.id}
+        name={props.name}
+        type="radio"
+        {...props.radioButtonProps}
+      />
+
+      {props.label && (
+        <span className="RadioButton__label">{props.label}</span>
+      )}
+    </label>
   );
 }
 
 RadioButton.propTypes = {
+  bordered: PropTypes.bool,
   disabled: PropTypes.bool,
   id: PropTypes.string.isRequired,
+  label: PropTypes.string.isRequired,
   name: PropTypes.string,
   radioButtonProps: PropTypes.object,
 };
 
 RadioButton.defaultProps = {
+  bordered: false,
   disabled: false,
   name: '',
   radioButtonProps: {},
--- a/stories/RadioButton.stories.jsx
+++ b/stories/RadioButton.stories.jsx
@@ -1,25 +1,22 @@
 import React from 'react';
 import { withA11y } from '@storybook/addon-a11y';
+import { withKnobs, text, boolean } from '@storybook/addon-knobs';
 
 import RadioButton from '../src/RadioButton/RadioButton';
 
 import { withPadding } from './decorators';
 
 export default {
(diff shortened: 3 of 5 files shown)


stories/RadioButton.stories.jsx
```diff
@@ -1,25 +1,33 @@
 import React from 'react';
 import { withA11y } from '@storybook/addon-a11y';
+import { withKnobs, text, boolean } from '@storybook/addon-knobs';
 
-import RadioButton from '../src/RadioButton/RadioButton';
```
davidmferris: 🙌

stories/RadioButton.stories.jsx
```diff
@@ -1,25 +1,33 @@
...
+import RadioButton from 'src/RadioButton/RadioButton';
 
-import { withPadding } from './decorators';
+import { withPadding } from 'stories/decorators';
+import { palette } from 'src/Styles';
 
 export default {
-  title: 'Design System/Buttons/Radio Button',
+  title: 'Design System/Form Elements/Radio Button',
```
davidmferris: I'm a little hesitant to change the namespace here. Buttons are not necessarily always form elements. And as we add more type of buttons & form stuff it might get cluttered in this directory.

SMaxOwok: I don't necessarily disagree with you.  In this case though, do radio buttons ever get used outside of forms?

davidmferris: It's probably unlikely TBH. I guess radio "Button" is a little bit of a misnomer since it's actually an `input`. On my todo list is to start adding actual buttons to this library but I suppose those could live in a buttons directory and this could stay here.

SMaxOwok: That was my thought as well 👍

src/RadioButton/RadioButton.jsx
```diff
@@ -1,27 +1,51 @@
...
+        className="RadioButton__input"
+        disabled={props.disabled}
+        id={props.id}
+        name={props.name}
+        type="radio"
+        {...props.radioButtonProps}
+      />
+
+      {props.label && (
+        <span className="RadioButton__label">{props.label}</span>
```
rroppolo: I might be missing something, but when should this be used over:
```
  <FormControlLabel
    labelHtmlFor="radio"
    text="Labeled radio button"
  >
    <RadioButton id="radio" name="radio" />
  </FormControlLabel>
```

This is another story under "Form Control Label" that looks quite similar to the new Radio button stories

davidmferris: Ah yeah you're right. I forgot about FormControlLabel.

SMaxOwok: Oh, interesting.  Let me look at how the radio button changes would work with this as the wrapper.

SMaxOwok: So I'm not sure this really works... 

The order of the markup has to change for the radio button with children, since the children have to come _after_ the form label.  That said, I do think it's valid that `RadioButton` has its own markup and styles as it's a specific use of a label and not necessarily in line with a generic form label.

Thoughts?

rsaris: FWIW when I was doing the form styles I always kinda wondered why we had `FormControlLabel` and would generally think this would be part of the `RadioButton` styles. I don't know that flexing and padding are form specific.

stories/RadioButton.stories.jsx
rsaris: THIS
IS
HUGE

Wanna pull a Rachel and clean up the whole repo?

stories/RadioButton.stories.jsx
```diff
@@ -1,25 +1,33 @@
...
 
-export const Disabled = () => (
+export const WithChildren = () => (
   <RadioButton
-    disabled
-    id="disabled"
-  />
+    label="Option with child"
+    bordered
+    id="default"
```
rsaris: Should this be unique?

.eslintrc
```diff
@@ -5,13 +5,13 @@
       "files": [
         "stories/*.{js,jsx}"
       ],
-      rules: {
-        "import/no-extraneous-dependencies": ["error", { "devDependencies": true }],
-      },
+      "rules": {
+        "import/no-extraneous-dependencies": ["error", { "devDependencies": true }]
+      }
```
rsaris: I see you too use Intellij... my IDE also screamed about this.

stories/RadioButton.stories.jsx
SMaxOwok: Very much, but lemme do that on a separate PR, so this can get merged and I can keep trucking on my actual task...

src/RadioButton/RadioButton.jsx
davidmferris: I bit that from [MUI](https://material-ui.com/api/form-control-label/). I was really trying to keep the base components as generic as possible so that we wouldn't have to add this label markup for each type of form control.

stories/RadioButton.stories.jsx
SMaxOwok: I guess it could be 'with-children', but it's still the only element on the page in this story.

src/RadioButton/RadioButton.jsx
rsaris: This is separate from `InputLabel` right? I don't know that I fully understand the use case for this vs. adding a label to the radio button. Should this be use the `FormControlLabel`?

stories/RadioButton.stories.jsx
rsaris: Checks out. Ignore me.

src/RadioButton/RadioButton.jsx
```diff
@@ -1,27 +1,51 @@
...
   disabled: PropTypes.bool,
   id: PropTypes.string.isRequired,
+  label: PropTypes.string.isRequired,
   name: PropTypes.string,
   radioButtonProps: PropTypes.object,
 };
 
 RadioButton.defaultProps = {
+  bordered: false,
+  children: null,
```
rsaris, line 48: Should this be `undefined`?

src/RadioButton/RadioButton.jsx
```diff
@@ -1,27 +1,51 @@
...
+          {props.children}
+        </div>
+      )}
+    </label>
   );
 }
 
 RadioButton.propTypes = {
+  bordered: PropTypes.bool,
+  children: PropTypes.node,
```
rsaris, line 38: FWIW per discussions with Darien it's not uncommon to not declare the `children` propType when you're just going to render it. I think there is a convention that you only declare the children prop when you want to make it required or set some specific rules on it (e.g. only a string) since it's such a weird prop.

src/RadioButton/RadioButton.jsx
SMaxOwok: This works and is typically what I do.  Is there a reason to prefer `undefined` here?

src/RadioButton/RadioButton.jsx
SMaxOwok: I'm OK with it either way.  I think it's useful to know the component _can_ take children, since it's not necessarily obvious a radio button would, but I don't feel strongly.  Would you rather it be removed?

davidmferris: Interesting, for some reason I thought if you rendered children without declaring the prop type you'd get a warning, but if not I'm fine either way.

src/RadioButton/RadioButton.jsx
rsaris: I think it lines up more with if the prop weren't in this list at all. It seems weird to put a value in here (even if it's null) vs. leaving it with the true default of undefined.

src/RadioButton/RadioButton.jsx
rsaris: I don't think so, though I may be wrong. I don't have a strong opinion here but if no one else cares I'd say let's remove it.

src/RadioButton/RadioButton.jsx
davidmferris: Yeah `InputLabel` is meant for like text/number type inputs. `FormControlLabel` is for radios, switches, and checkboxes (the latter 2 which don't exist yet). Perhaps the naming could use some improvement since it's definitely not obvious

PR #34, opened 2020-03-31
UIDS-33 Allow links in flash messages

### LINKED ISSUE #33: Allow links in flash messages
Needed for https://github.com/user-interviews/rails-server/issues/2547
This flash message contains a link.  Right now we escape all messages.

rroppolo:
Fixes #33 

Needed for user-interviews/rails-server#2547
The flash message contains a link. Right now we escape all messages.

Screen Shot 2020-03-31 at 4 28 40 PM


rroppolo:
Ohh I was just looking at that component the other day and totally missed that!  Thanks, that did the trick!

PR #43, opened 2020-04-08
RS-42 refactor flash message component to exportable alert message component

angrobertsh:
Makes an alert message component to be used in places like in modals where you want colorful alert stuff.

See: AlertMessage in stories

Screen Shot 2020-04-08 at 2 18 22 PM


src/Flash/AlertMessage.jsx
```diff
@@ -0,0 +1,60 @@
...
+        props.showDismiss && (
+          <button
+            className="close"
+            type="button"
+            onClick={handleDismissMessage}
+          >
+          &times;
+          </button>
+        )
+      }
```
rsaris, line 36: Could we just have this check if `props.onDismiss` exists and if not don't show it?

src/Flash/AlertMessage.jsx
```diff
@@ -0,0 +1,60 @@
...
+  return (
+    <div className={alertMessageClassName(props.type)}>
+      {
+        props.showDismiss && (
+          <button
+            className="close"
+            type="button"
+            onClick={handleDismissMessage}
+          >
+          &times;
```
rsaris, line 33:
```suggestion
            &times;
```

stories/Alert.stories.jsx
```diff
@@ -0,0 +1,28 @@
...
+import '../scss/global.scss';
+
+export default {
+  title: 'Design System/Alert',
+  component: AlertMessage,
+  decorators: [withKnobs],
+};
+
+export const alertMessage = () => (
+  <div>
```
rsaris, line 16: Any reason for the `div`?

src/Flash/AlertMessage.jsx
angrobertsh: @rroppolo and I were talking about this, and we thought that having the close button without a `props.onDismiss` would still have a default behavior of closing the modal -- the `props.onDismiss` just adding extra functionality to that. This is corroborated by the optionality of the `onDismiss` on the higher level `Flash` component.

stories/Alert.stories.jsx
angrobertsh: No special reason. Removing.

src/Flash/AlertMessage.jsx
rsaris: Ah -- that makes sense!

src/Flash/AlertMessage.scss
```diff
@@ -0,0 +1,26 @@
...
+@import '~bootstrap/scss/mixins';
+@import '~bootstrap/scss/alert';
+
+@import '../../scss/theme';
+
+// Flash colors
+$ux-flash-error: $ux-red-100;
+$ux-flash-error-border: $ux-red-200;
+
+.alert {
```
davidmferris, line 12: Our convention for these components is to name the base class after the component name, so this would be `.AlertMessage`

stories/Alert.stories.jsx
```diff
@@ -0,0 +1,26 @@
...
+
+export default {
+  title: 'Design System/Alert',
+  component: AlertMessage,
+  decorators: [withKnobs],
+};
+
+export const alertMessage = () => (
+  <AlertMessage
+    message={text('Message', 'Default message')}
```
davidmferris, line 17: I think you need to pass an `id` prop to avoid a missing required prop warning.

src/Flash/AlertMessage.scss
```diff
@@ -0,0 +1,26 @@
+@import '~bootstrap/scss/functions';
+@import '~bootstrap/scss/variables';
+@import '~bootstrap/scss/mixins';
+@import '~bootstrap/scss/alert';
+
+@import '../../scss/theme';
```
davidmferris, line 6: Double check me on this, but I *think* you can just do `@import 'scss/theme';` now since we updated our resolve paths.

src/Flash/AlertMessage.scss
```diff
@@ -0,0 +1,26 @@
...
+
+// Flash colors
+$ux-flash-error: $ux-red-100;
```
davidmferris, line 9: I know this is existing code from a different component but I've tended towards defining these tokens outside of component stylesheets and instead putting them in the `scss` directory. The reason being that we want to be able to use certain styles in our non-react parts of the app and it makes them easier to import that way.

stories/Alert.stories.jsx
```diff
@@ -0,0 +1,26 @@
...
+import 'bootstrap/dist/css/bootstrap.min.css';
+import '../scss/global.scss';
+
+export default {
+  title: 'Design System/Alert',
+  component: AlertMessage,
+  decorators: [withKnobs],
+};
+
+export const alertMessage = () => (
```
davidmferris, line 15: could you add a story for a dismissable alert? Check out Form.stories.jsx on how to handle passing in actions.

src/Flash/AlertMessage.scss
rsaris: I think this relies on a Bootstrap class so we may have to special case this one though I guess we could do `.AlertMessage` here and put both on the component.

src/Flash/AlertMessage.scss
rsaris: You can not. This is what broke the build that Max fixed.

src/Flash/AlertMessage.jsx
```diff
@@ -0,0 +1,60 @@
...
+  );
+}
+
+AlertMessage.propTypes = {
+  id: PropTypes.string.isRequired,
+  message: PropTypes.oneOfType([PropTypes.string, PropTypes.element]).isRequired,
+  showDismiss: PropTypes.bool,
+  title: PropTypes.string,
+  type: PropTypes.string.isRequired,
+  onDismiss: PropTypes.func,
```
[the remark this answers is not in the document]
angrobertsh: Oops you're totally right, actually testing this now, it looks like this dismiss definitely doesn't have any default behavior and needs a function.

src/Flash/AlertMessage.jsx
angrobertsh: 180ing on this. onDismiss is definitely necessary to make the close button not blow up

src/Flash/AlertMessage.scss
angrobertsh: Just tried this `scss/theme` path and everything seems to work!

davidmferris: Yeah I forgot, this might actually still be a problem when we install the package on rails-server.

rsaris: Correct -- we should only be using relative paths in SCSS unless @SMaxOwok got it working right.

angrobertsh: I put it back

scss/colors.scss
```diff
@@ -1,2 +1,3 @@
 @import './colors/inputs';
 @import './colors/palette';
+@import './colors/components';
```
rsaris, line 3: Could be alphabetized.

scss/colors/components.scss
```diff
@@ -0,0 +1,3 @@
+// AlertMessage/Flash colors
```
rsaris, line 1: Could this be `colors/flash`? Components is a little generic.

src/Flash/AlertMessage.scss
```diff
@@ -0,0 +1,22 @@
...
+@import '~bootstrap/scss/variables';
+@import '~bootstrap/scss/mixins';
+@import '~bootstrap/scss/alert';
+
+@import '../../scss/theme';
+
+.AlertMessage {
+  font-weight: 500;
+
+  h2 {
```
rsaris, line 11: I wonder if this shouldn't just be `AlertMessage_title` I worry about relying on `h2` for styling as well as relying on styles that are already present for `h2` to create most of these styles (e.g. margins etc). I'd rather we just use a `p` and put the styles on that.

angrobertsh:
Agree with and have resolved Bob's comments

stories/Alert.stories.jsx
angrobertsh: (This has been done. The story doesn't do anything but it does fire an action)

stories/Alert.stories.jsx
```diff
@@ -0,0 +1,37 @@
...
+export default {
+  title: 'Design System/Alert',
+  component: AlertMessage,
+  decorators: [withKnobs],
+};
+
+export const WithoutDismiss = () => (
+  <AlertMessage
+    message={text('Message', 'Default message')}
+    id={text('ID', '1')}
```
davidmferris, line 19: I don't think you need a `text()` here. That's typically reserved for text that shows up in the component so you can use the knob to see how the component reacts to variable text.

stories/Alert.stories.jsx
```diff
@@ -0,0 +1,37 @@
...
+);
+
+const onDismiss = (id) => {
+  action('alert dismissed')(id);
+};
+
+export const WithDismiss = () => (
+  <AlertMessage
+    message={text('Message', 'Default message')}
+    id={text('ID', '1')}
```
davidmferris, line 32: Same here.

angrobertsh: I actually like the knobs here -- the dispatch on the dismiss sends the id of the component as the only param, and having this info is pretty useful. For consistency's sake I think it's worth keeping as a knob in both stories

angrobertsh: (if you don't think that it provides useful information i can remove just say the word and it'll go bye bye)

angrobertsh: @davidmferris the id for these elements is the only param that's passed into the `handleDismiss` and I think it might be worth having the knob for (you can see the id dispatched in the action). If you don't think so let me know and I can remove it.

davidmferris: I think it still passes the id to the action even without the knob right? I'm not totally sure what value the knob provides since if you update the ID it doesn't really change any behavior in the component itself. Doesn't hurt anything though so I don't mind if you want to keep it in.

angrobertsh: I removed the knob since we don't really do anything with it in the display which I'm gathering is what the storybook is about.

angrobertsh:
(Do I push merge now)

rsaris:
@davidmferris you're the owner of this so I'll leave merging to you.

Issue #15, opened 2020-02-28
Configure storybook to deploy to Github pages


rsaris:
It'd be pretty dope if we could get this set up to deploy to github pages, even if that was on some random branch. I know I've gotten something like this set up before w/ personal webpages using https://github.com/tschaub/gh-pages but I imagine Storybook may have something more specialized to it (that may already be installed in this repo?)

jhforster:
A real task?? For me?? Oh man, this is exciting! (seriously) Love me some github pages.

PR #45, opened 2020-04-13
Install storybook-deployer

jhforster:
This is related to #15 

So this [storybook deployer](https://github.com/storybookjs/storybook-deployer) thing creates a `package-lock.json` file. Is that okay?

I wanted to check on that before messing with it too much more.


jhforster:
I don't know if we need this PR or not? I'm confused.

davidmferris:
Doesn't hurt to keep our dependencies current, so I'm cool with having this PR.

PR #46, opened 2020-04-21
CHORE Add getOptionLabel,getOptionValue to SingleSelect

rroppolo:
Small chore to add two pass-through props to SingleSelect: getOptionLabel and getOptionValue


rsaris:
@davidmferris assigning to you to merge and release as you need it.

PR #47, opened 2020-04-22
CHORE Add defaultValue to AsyncSelect

rroppolo:
Another pass-through prop that I missed the first time to set a defaultValue on the AsyncSelect.

### DIFF
--- a/src/Select/AsyncSelect.jsx
+++ b/src/Select/AsyncSelect.jsx
@@ -27,8 +27,10 @@ const AsyncSelect = ({
   value,
 
   onChange,
+  ...props
 }) => (
   <Async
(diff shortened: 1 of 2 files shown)


rsaris:
Honestly at this point I may just call it that Jeff is a silly person and we should just pass any unknown prop to ReactSelect...

rroppolo:
I already found another prop I'm going to need to add, so that may be a good option...

rsaris:
Let's just pull the trigger and do it. I'm high on keeping it so we expose props we use commonly in propTypes to make it easier to know what is available (and suggested) vs. having to dig into the ReactSelect docs.

PR #49, opened 2020-05-14
UIDS-48 Make select heights consistent with other input controls

### LINKED ISSUE #48: Make height consistent between select components and form-control inputs.
Right now we have the following css in rails-server:
```
.FormGroup {
  .form-control {
    height: 2.25rem;
  }
}
```

Our design system components do not have this height set, which leads to a 1px difference in height between the two when you mix and match.  This can lead to situations like this when you try to switch to using the design system selects:

Screen Shot 2020-05-14 at 9 55 37 AM

You can see the heights do not quite match by 1px.  The process could be made much easier by consolidating height differences b/t rails-server and the ui-design-system.

rroppolo:
Fixes #48 

We noticed when trying to switch to the design system components in the project filtering task that the heights did not quite match up between the design system and the form controls in rails-server.  This was due to a piece of css overriding the height of .form-control in rails-server, but which did not apply to the DS selects.  See there is a small 1px difference:

Screen Shot 2020-05-14 at 9 55 37 AM

From the discussion in the design system slack, it sounds like the height in rails-server (```height: 2.25rem;```) is the correct one.  So here I attempt to make the default design system styles reflect this.  I added a size prop to the selects like we did for modals because I imagine there will be cases where we could want a larger select.  I didn't include any other sizes yet though b/c wasn't sure what we would want, just left the door open for it. 

Here it is with the new height:
Screen Shot 2020-05-14 at 12 44 46 PM

### DIFF
--- a/src/FormGroup/FormGroup.scss
+++ b/src/FormGroup/FormGroup.scss
@@ -22,4 +22,8 @@
       @include ui-remove-list-styles();
     }
   }
+
+  .form-control {
+    height: 2.25rem;
+  }
 }
--- a/src/Select/styles.js
+++ b/src/Select/styles.js
@@ -1,47 +1,60 @@
 import systemColors from 'src/Styles/colors';
 
+export const SELECT_SIZES = { SMALL: 'small' };
+
+const getHeightProps = (size) => {
+  if (size === SELECT_SIZES.SMALL) {
+    return {
+      height: '2.25rem',
+      minHeight: '2.25rem',
+    };
+  }
+  return null;
+};
+
 /*
  To set styles for your item, make sure your option object has a child `colors` object
  with a text and/or hover key defined to override the defaults
  */
-const defaultStyles = {
-  control: (styles, { isDisabled }) => ({
-    ...styles,
-    backgroundColor: isDisabled ? systemColors.inputDisabledBg : styles.backgroundColor,
-    borderColor: systemColors.inputBorderColor,
-  }),
-  indicatorSeparator: (styles) => ({ ...styles, display: 'none' }),
-  singleValue: (styles, { data }) => ({
-    ...styles,
-    color: (data.colors ? data.colors.text : systemColors.uxGray600) || systemColors.uxGray600,
-  }),
-  option: (styles, {
-    data,
-    isDisabled,
-    isFocused,
-    isSelected,
-  }) => {
-    const colors = data.colors || {};
-
-    return {
+const defaultStyles = ({ size }) => ({
+    control: (styles, { isDisabled }) => ({
       ...styles,
-      backgroundColor: isSelected || isFocused ? colors.hover : styles.backgroundColor,
-      color: colors.text,
-      cursor: 'pointer',
+      ...getHeightProps(size),
+      backgroundColor: isDisabled ? systemColors.inputDisabledBg : styles.backgroundColor,
+      borderColor: systemColors.inputBorderColor,
+    }),
+    indicatorSeparator: (styles) => ({ ...styles, display: 'none' }),
+    singleValue: (styles, { data }) => ({
+      ...styles,
+      color: (data.colors ? data.colors.text : systemColors.uxGray600) || systemColors.uxGray600,
+    }),
+    option: (styles, {
+      data,
+      isDisabled,
+      isFocused,
+      isSelected,
+    }) => {
+      const colors = data.colors || {};
 
-      ':active': {
-        ...styles[':active'],
-        backgroundColor:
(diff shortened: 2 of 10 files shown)


src/Select/Select.scss
```diff
@@ -0,0 +1,9 @@
+.SingleSelect,
+.AsyncSelect {
+  &--small {
+    > div[class$='-control'] {
```
rroppolo, line 4: I realized the min-height is on the control div.  If this is too weird we can alternatively do it in javascript.

davidmferris: I think I'd maybe lean towards javascript since that's how we're setting styles with react-select and *seems* to be their preferred method.

src/FormGroup/FormGroup.scss
```diff
@@ -22,4 +22,8 @@
       @include ui-remove-list-styles();
     }
   }
+
+  .form-control {
+    height: 2.25rem;
```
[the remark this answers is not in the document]
rroppolo: Yeah this was more to have the design system FormGroup styles in sync with rails-server and matching the new select heights of the 2.25rem (36px). But yeah non-FormGroup inputs are still an issue in rails-server, but I was planning to fix that in my rails-server branch when I pull in the new DS release, because right now there's not a concept of a bare input component in the design system.  Hopefully at some point we can just import css in from the DS though and not have any in rails-server, but sounds like Bob was saying that work is already underway elsewhere.  Does that approach make sense?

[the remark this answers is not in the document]
davidmferris: Totally! Makes perfect sense.

src/Select/Select.scss
rroppolo: Yeah makes sense - I liked this approach until I realized I actually had to style the internal div which did feel icky

src/Select/styles.js
```diff
@@ -1,21 +1,34 @@
...
-    ...styles,
-    backgroundColor: isDisabled ? systemColors.inputDisabledBg : styles.backgroundColor,
-    borderColor: systemColors.inputBorderColor,
-  }),
-  indicatorSeparator: (styles) => ({ ...styles, display: 'none' }),
-  singleValue: (styles, { data }) => ({
+const defaultStyles = ({ size }) => ({
+    control: (styles, { isDisabled }) => ({
+      ...styles,
+      ...getHeightProps(size),
```
rroppolo, line 22: @davidmferris Moved it in to Javascript and added a few specs - let me know what you think!

src/FormGroup/FormGroup.scss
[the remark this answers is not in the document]
rsaris: I'd say unless there are pages where we'll end up with some serious inconsistencies let's just leave this in the "modern styles" and accept that they are 1px off. Bootstrap has some annoyingly linked styles so I worry that if we put this height on the default `.form-control` we're going to end up messing up those assumptions. Let me know if this does not align with what y'all are seeing.

That said -- we may need to update the `ui_form.scss` file in `rails-server` to match these styles until we get this exporting the common form styles.

src/Select/styles.js
```diff
@@ -1,47 +1,60 @@
...
+export const SELECT_SIZES = { SMALL: 'small' };
+
+const getHeightProps = (size) => {
+  if (size === SELECT_SIZES.SMALL) {
+    return {
+      height: '2.25rem',
+      minHeight: '2.25rem',
+    };
+  }
+  return null;
```
rsaris, line 12: Does this need to return `{}` or will splatting null work?

rroppolo: It does work with null because object spread operator follows the behavior of Object.assign https://tc39.es/proposal-object-rest-spread/#AbstractOperations-CopyDataProperties. I can change it if it looks too weird though.

src/FormGroup/FormGroup.scss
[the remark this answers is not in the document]
rroppolo: We are already overriding ```.FormGroup .form-control``` this way in rails-server css, so this was more to make this repo consistent with what we are already doing there. The select was one height inconsistency, but I imagine there could still be others. But my understanding from Skye and JH was that the correct height is the 2.25rem and not the bootstrap default styles. I can remove this, but this change only really effects the design system repo b/c this css already exists in rails-server.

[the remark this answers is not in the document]
rsaris: Oh sorry -- I think this is the right solution, but I'd only do it in `.FormGroup .form-control` and not on the global scope.

src/Select/styles.js
rsaris: Nope -- that's fine! Was just making sure.

PR #52, opened 2020-06-17
RS-31 Move as many dependencies to peerDependencies as possible

### LINKED ISSUE #31: Move font-awesome, react-tracking, react-popper dependencies into peerDependencies
See: https://github.com/user-interviews/rails-server/pull/2573/files#r399433207

davidmferris:
Closes #31 

The design system package is actually one of the largest packages we have installed on rails-server because it needlessly installs a bunch of dependencies that already exist there.  By moving them to peerDependencies the design system installation size is reduced from 10mb to 2.8mb.


src/Card/Card.jsx
```diff
@@ -1,27 +0,0 @@
-import React from 'react';
-import PropTypes from 'prop-types';
-
-import BootstrapCard from 'react-bootstrap/Card';
```
davidmferris, line 4: I deleted this component for now because it's not in use & actually doesn't mirror the real `Card` component. This was the first one I added and was kind of a placeholder and is the only place `react-bootstrap` was used. Next reset week I can implement `Card` the correct way.

package.json
```diff
@@ -44,14 +31,29 @@
...
   },
   "devDependencies": {
     "@babel/cli": "^7.8.4",
     "@babel/core": "^7.8.4",
     "@babel/plugin-proposal-object-rest-spread": "^7.8.3",
     "@babel/preset-env": "^7.8.4",
+    "@fortawesome/fontawesome-svg-core": "^1.2.28",
+    "@fortawesome/free-regular-svg-icons": "^5.13.0",
+    "@fortawesome/free-solid-svg-icons": "^5.13.0",
+    "@fortawesome/react-fontawesome": "^0.1.9",
```
rsaris, line 56: I don't understand any of this, so pardon my ignorance, but why do we need this in dev and peer dependencies?

rsaris: Is it because we use this to make fullstory work?

package.json
```diff
@@ -44,14 +31,29 @@
     ]
   },
   "peerDependencies": {
+    "bootstrap": "^4.3.1",
+    "classnames": "^2.2.5",
```
rsaris, line 35: This is annoying question, especially if this is auto generated, but why is the sort different here from devDeps?

package.json
```diff
@@ -44,14 +31,29 @@
...
+    "@fortawesome/free-regular-svg-icons": "^5.13.0",
+    "@fortawesome/free-solid-svg-icons": "^5.13.0",
+    "@fortawesome/react-fontawesome": "^0.1.9",
+    "node-sass": "^4.9.2",
+    "prop-types": "^15.6.1",
     "react": "^16.12.0",
-    "react-dom": "^16.12.0"
+    "react-copy-to-clipboard": "^5.0.2",
+    "react-dom": "^16.12.0",
+    "react-popper": "^1.3.7",
```
rroppolo, line 45: Are some of these missing from devDependencies?  Like react-popper, classnames? Apologies if they are there and I just can't read :p

package.json
davidmferris: Yeah exactly! `peerDependencies` won't actually install the packages locally so if we want to see the components in full story we need to add them here.

package.json
davidmferris: Ugh no they were missing. Updated to include them there.

package.json
davidmferris: This was a poor copy and paste job on my part. My eyes completely missed the "@" somehow.

PR #53, opened 2020-07-08
[CHORE] Add green pill component

SMaxOwok:
Also adds a 'squared' prop to make the pill have sharper corners.

### DIFF
--- a/src/Pill/Pill.jsx
+++ b/src/Pill/Pill.jsx
@@ -4,11 +4,12 @@ import classNames from 'classnames';
 
 import './Pill.scss';
 
-const Pill = ({ color, size, text }) => (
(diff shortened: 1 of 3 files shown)


SMaxOwok:
Per https://www.figma.com/file/EWu17C9ZGqJ5vBlGXIOcEl/Location-Requirements?node-id=61%3A3341

I know squaring it technically makes it not a "pill", but functionally these seemed like the same components, so I thought it made more sense to add it here.

SMaxOwok:
What's the status on this?  We're running into another instance where we need this green one.

davidmferris (approved):
Sorry for the delay, looks good!

PR #55, opened 2020-09-04
UIDS-54 Initial commit of typography story

rroppolo:
Just a first start on the typography story and tokens. I expect we will iterate on this!

At first I had a token for every combination of size and weight, but I noticed other design systems being more opinionated about this. For example not giving the option to use a larger font with the highest font-weight, the smallest size for helper text does not have a bolded option, etc. I started that here but still trying to get a feel for what makes the most sense in our current usage in the app, we have a lot of combinations that we may be able to consolidate into fewer options once we do the designer review.

Story has been deployed here: https://docs.userinterviews.com/ui-design-system/?path=/story/design-system-styles--typography

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -12,7 +12,8 @@
     {
       "files": [
         "src/**/*.{js,jsx}",
-        "spec/**/*.{js,jsx}"
+        "spec/**/*.{js,jsx}",
+        "stories/**/*.{js,jsx}"
       ],
       "env": {
         "browser": true
--- a/.github/workflows/gh-pages-deploy.yml
+++ b/.github/workflows/gh-pages-deploy.yml
@@ -2,7 +2,7 @@
   on:
     push:
       branches:
-        - master
+        - feature/UIDS-54-typography-story
   jobs:
     gh-pages-deploy:
       name: Deploying to gh-pages
--- a/scss/global.scss
+++ b/scss/global.scss
@@ -1,7 +1,2 @@
 @import url('https://fonts.googleapis.com/css?family=Lato:300,300italic,400,700:latin');
 @import './theme';
-
-body {
-  font-family: $font-family;
-  font-size: $font-size-base;
-}
--- a/scss/global.scss
+++ b/scss/global.scss
@@ -1,2 +1,3 @@
 @import url('https://fonts.googleapis.com/css?family=Lato:300,300italic,400,700:latin');
 @import './theme';
+@import './baseline';
--- a/scss/typography.scss
+++ b/scss/typography.scss
@@ -1,3 +1,5 @@
+@import './colors/palette';
+
 $font-family: Lato, Arial, sans-serif;
 $font-weight-light: 200;
 $font-weight-normal: 300;
@@ -7,21 +9,18 @@ $headings-font-weight: 400;
 $headings-font-weight-bold: 700;
 
 @mixin font-type-100() {
-  font-family: $font-family;
   font-size: .75rem;
   font-weight: $font-weight-normal;
   line-height: 1rem;
 }
 
 @mixin font-type-200() {
-  font-family: $font-family;
   font-size: $font-size-base;
   font-weight: $font-weight-normal;
   line-height: 1.25rem;
 }
 
 @mixin font-type-300() {
-  font-family: $font-family;
   // font-size:  $font-size-base; -- should base be changed to 1 rem?
   font-size: 1rem;
   font-weight: $font-weight-normal;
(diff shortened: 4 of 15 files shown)


.eslintrc
```diff
@@ -12,7 +12,8 @@
     {
       "files": [
         "src/**/*.{js,jsx}",
-        "spec/**/*.{js,jsx}"
+        "spec/**/*.{js,jsx}",
+        "stories/**/*.{js,jsx}"
```
rroppolo, line 16: My story was giving me the 'no spread operator' rule which annoyed me so I enabled the same elsint config for stories.

scss/typography.scss
```diff
@@ -11,13 +13,15 @@ $headings-font-weight-bold: 700;
   font-size: .75rem;
   font-weight: $font-weight-normal;
   line-height: 1rem;
+  color: $ux-gray-900;
```
rroppolo, line 16: I added color here as a base - can you think of any reason why we would not want to add this? I feel like some places were adding color in a one-off declaration which felt weird, and some used slight variations of greys which would eventually be cool to standardize in the DS.

scss/typography.scss
```diff
@@ -26,11 +30,54 @@ $headings-font-weight-bold: 700;
...
+}
+
+@mixin font-type-500--bold() {
+  @include font-type-500;
+  font-weight: $headings-font-weight;
+}
+
+@mixin font-type-600--bold() {
+  @include font-type-600;
+  font-weight: $headings-font-weight;
```
rroppolo, line 82: I started out with every combination of weight + size and then as I said in the description sort of paired it back to use the heading weight in larger sizes and bold in smaller. I think in our app we have a lot of combinations that could be paired back, so I imagine this will evolve...

stories/Alert.stories.jsx
```diff
@@ -15,8 +15,8 @@ export default {
 
 export const WithoutDismiss = () => (
   <AlertMessage
-    message={text('Message', 'Default message')}
```
rroppolo, line 18: Sorry a bunch of eslint fixes in this PR...

davidmferris: Yeah it was probably an oversight to not be linting these files in the first place.

scss/typography.scss
davidmferris: I can't think of any reason why not.

scss/typography.scss
```diff
@@ -7,30 +9,69 @@ $headings-font-weight: 400;
...
+}
+
+body {
+  font-family: $font-family;
+  font-size: $font-size-base;
+  color: $ux-gray-900;
```
rroppolo, line 76: Ripped these out per our convo earlier @davidmferris

rroppolo: Ah ok @rsaris that makes sense. Do you think it is ok to include this in theme.scss then? At first I had these three declarations included in each of the types, but as Dave and I investigated a bit this would mean they would be stamped in every place we use the mixin in the app, which could increase our css size. It seemed better to include these properties once on the body and then override them on a case-by-case basis. Global.scss where they were before is not included in rails-server, I also considered moving it into theme.scss

rsaris: I am worried about that since we include that in `ui_theme.scss` which is included in a ton of other CSS files. I wonder if we should just include global in rails server somewhere or add a different file which we include in the rails server somewhere.

rroppolo: Ahh I see now.. I moved it to a new file called baseline.scss and included that in global.scss. Then I imagine we would include it in rails-server somewhere (maybe layout?) where we can be ensured it would be included at most per page. Sound good?

Since the old mixins are also defined in the app in app/javascript/common/mixins/_ui_fonts.scss this shouldn't affect current rails-server code as we renamed the mixins anyway so any existing code will just use the old types from there.

.github/workflows/gh-pages-deploy.yml
```diff
@@ -2,7 +2,7 @@
   on:
     push:
       branches:
-        - master
+        - feature/UIDS-54-typography-story
```
rsaris, line 5: :bulb: Any way we can have this be "current branch"?!

stories/Styles/Typography/Typography.scss
```diff
@@ -0,0 +1,118 @@
...
+  max-width: 33.3%;
+  padding: 1rem 1.5rem;
+  width: 33.3%;
+
+  li {
+    min-height: 1.5rem;
+  }
+}
+
+.font-type-10 {
```
rsaris, line 36: @davidmferris did we ever figure out a good way to add CSS classes which can be used in non-React pages? If so, do we want these to be in there?

.github/workflows/gh-pages-deploy.yml
davidmferris: I'm guessing this is a temporary change to see the results? We wouldn't want a permanent current branch I don't think because that would overwrite the storybook anytime someone pushed a commit to any branch.

stories/Styles/Typography/Typography.scss
rroppolo: Right now the tokens are mixins, so right now they are only used in scss files. I created these css classes only for the purposes of the story, to show a styled example of each mixin. Not sure I totally follow why we couldn't use these mixins on non-React pages as well? I didn't expose the fonts in Javascript or anything b/c it didn't seem necessary..

rsaris: I think that makes sense. Was just thinking through if there is value in just being able to use these directly but I have a feeling that's not really going to be useful.

scss/global.scss
```diff
@@ -1,7 +1,3 @@
 @import url('https://fonts.googleapis.com/css?family=Lato:300,300italic,400,700:latin');
 @import './theme';
-
-body {
-  font-family: $font-family;
-  font-size: $font-size-base;
-}
+@import './baseline';
```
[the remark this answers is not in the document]
rsaris: OK great -- that all makes sense! Do we have to export this anywhere?

rsaris (approved):

rroppolo:
If we follow what we did for theme.scss then we can just do `@import '~ui-design-system/scss/baseline.scss';` if that looks ok. Tested it and it works

stories/Styles/Typography/Typography.stories.jsx
```diff
@@ -0,0 +1,160 @@
...
+    size: '1.5rem',
+    weight: '400 / Regular',
+    lineHeight: '1.75rem',
+    modifiers: ['font-type-70--bold', 'font-type-70--light'],
+  },
+  {
+    preset: 'font-type-60',
+    font: 'Lato, Arial, sans-serif',
+    size: '1.25rem',
+    weight: '300 / Light',
```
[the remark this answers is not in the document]
rroppolo: Ah thank you - I had this locally and never pushed it! 🤦  Fixed this and added the text-transform for the 10px size!

PR #62, opened 2020-10-14
UIDS-61 Add Tooltip component

### LINKED ISSUE #61: Add Tooltip component

davidmferris:
Closes #61

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "@fortawesome/free-regular-svg-icons": "^5.13.0",
     "@fortawesome/free-solid-svg-icons": "^5.13.0",
     "@fortawesome/react-fontawesome": "^0.1.9",
+    "@popperjs/core": "^2.5.3",
     "bootstrap": "^4.3.1",
     "classnames": "^2.2.5",
     "node-sass": "^4.9.2",
@@ -42,7 +43,7 @@
     "react": "^16.12.0",
     "react-copy-to-clipboard": "^5.0.2",
     "react-dom": "^16.12.0",
-    "react-popper": "^1.3.7",
+    "react-popper": "^2.2.3",
     "react-tracking": "^7.3.0"
   },
   "devDependencies": {
@@ -54,6 +55,7 @@
     "@fortawesome/free-regular-svg-icons": "^5.13.0",
     "@fortawesome/free-solid-svg-icons": "^5.13.0",
     "@fortawesome/react-fontawesome": "^0.1.9",
+    "@popperjs/core": "^2.5.3",
     "@storybook/addon-a11y": "^5.3.13",
     "@storybook/addon-actions": "^5.3.13",
     "@storybook/addon-docs": "^5.3.13",
@@ -90,7 +92,7 @@
     "react": "^16.12.0",
     "react-copy-to-clipboard": "^5.0.2",
     "react-dom": "^16.12.0",
-    "react-popper": "^1.3.7",
+    "react-popper": "^2.2.3",
     "react-test-renderer": "^16.12.0",
     "react-tracking": "^7.3.0",
     "sass-loader": "^8.0.2",
--- a/src/Popper/Popper.jsx
+++ b/src/Popper/Popper.jsx
@@ -9,11 +9,11 @@ function Popper(props) {
   return (
     <Manager>
       <Reference>
-        {({ ref }) => <div ref={ref}>{props.children}</div>}
+        {({ ref }) => React.cloneElement(props.children, { ref })}
       </Reference>
       {
         props.visible && (
-          <ReactPopper placement={props.placement}>
+          <ReactPopper placement={props.placement} strategy={props.strategy}>
             {
               ({ placement, ref, style }) => (
                 <div
@@ -22,6 +22,14 @@ function Popper(props) {
                   ref={ref}
                   style={style}
                 >
+                  {
+                    props.header && (
+                      <div className="Popper__header">
+                        {props.header}
+                      </div>
+                    )
+                  }
+
                   {props.text}
                 </div>
               )
@@ -34,15 +42,20 @@ function Popper(props) {
 }
 
 Popper.propTypes = {
+  children: PropTypes.node.isRequired,
   dark: PropTypes.bool,
+  header: PropTypes.string,
   placement: PropTypes.string,
+  strategy: PropTypes.string,
   text: PropTypes.string.isRequired,
   visible: PropTypes.bool,
 };
 
 Popper.defaultProps = {
   dark: false,
+  header: undefined,
   placement: 'top',
+  strategy: undefined,
   visible: false,
 };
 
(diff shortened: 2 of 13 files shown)


src/Popper/Popper.jsx
```diff
@@ -9,11 +9,11 @@ function Popper(props) {
   return (
     <Manager>
       <Reference>
-        {({ ref }) => <div ref={ref}>{props.children}</div>}
+        {({ ref }) => React.cloneElement(props.children, { ref })}
```
davidmferris, line 12: Turns out we're not even using Popper yet in the app so this isn't a breaking change, but the div's display block was causing the tooltip icon to wrap to the next line so I think we don't want to add that extra element.

davidmferris:
@jojolee1996 - lemme know if you want a rundown on how to run storybook locally to see this in action

package.json
```diff
@@ -90,7 +92,7 @@
     "react": "^16.12.0",
     "react-copy-to-clipboard": "^5.0.2",
     "react-dom": "^16.12.0",
-    "react-popper": "^1.3.7",
+    "react-popper": "^2.2.3",
```
rsaris, line 95: Is this what we're using on the app? May be worth cutting that over to this version if we're not (and use the new component EVERYWHERE!)

src/Tooltip/Tooltip.scss
```diff
@@ -0,0 +1,24 @@
...
+    &--gray {
+      color: $ux-gray-800;
+    }
+
+    &--green {
+      color: $ux-green;
+    }
+
+    // TODO this is a fallback to match the in-app naming. Once all uses
+    // have been cut over we should consider using --green instead
```
rsaris, line 19: Do we use this currently in the app? If so couldn't we just do this rename when we move the components that use this to use the right color?

rsaris: One question though -- we may want to keep some of these semantic concepts in the DS. In theory -- we want our DS to match our themes right? So it probably should have the concept of "researcher" and "participant" colors? Or maybe we can just use "primary" and "secondary" respectively?

rsaris (approved):

package.json
davidmferris: Yep @rroppolo just updated this in app so I did it here to match.

src/Tooltip/Tooltip.scss
davidmferris: Yeah I'm down to keep the semantic names. I wasn't sure if we wanted to do that because we haven't as of yet. We should probably be mindful of that at least a little though, i.e. we wouldn't want any `--company` modifiers 😝

package.json
rsaris: :fire:

src/Tooltip/Tooltip.scss
rsaris: True -- I'm kinda tempted to use "primary" and "secondary" though the researcher / participant split is a little weird there... I'd probably defer to the design team here though as hopefully they'd have grander ideas about how we could use these semantic terms to build a real theming engine.

src/Popper/Popper.scss
```diff
@@ -1,15 +1,19 @@
...
+@import '../../scss/theme';
 
 .Popper {
   background-color: $ux-white;
   border-radius: $ux-border-radius;
   border: 1px solid $ux-gray-300;
   box-shadow: $ux-box-shadow-card;
   max-width: 15rem;
+  letter-spacing: normal;
   padding: .25rem .5rem;
```
rroppolo, line 10: In the app we made the padding: 0.75rem; Also should we add typography?

src/Tooltip/TrackedTooltip.jsx
```diff
@@ -0,0 +1,5 @@
+import withTrackedClick from 'src/hoc/withTrackedClick';
```
rroppolo, line 1: I think we would only want to track on show, right?  Getting two events for each interaction feels like it would clutter up the purpose fo the event.

src/Popper/Popper.scss
```diff
@@ -1,15 +1,19 @@
...
+  z-index: $z-index-flash;
+
+  &__header {
+    border-bottom: 1px solid $ux-gray-300;
+    margin-bottom: 0.5rem;
+  }
 
   &--dark {
```
rroppolo, line 18: Also I'll defer to @jojolee1996 but we may want to just make the dark styles the default if we plan to use them app-wide

jojolee1996: Sounds good to me, I'll make note of it in our design system as well.

src/Popper/Popper.scss
davidmferris: Whoops missed those. Will add!

src/Popper/Popper.scss
davidmferris: Do you mean for Tooltip in particular or Popper as a whole? I want to say we have light Poppers around the app currently.

jojolee1996: This should just be for the tooltip message that appears. What would be considered a popper?

davidmferris: Popper is the container that contains the tooltip (i.e. just any box that shows up). Just glancing through the code, it looks like the container that shows up when you select a label is also a popper so we probably just want to enforce this for tooltips only.

jojolee1996: Ah yeah if we could limit it to just tooltips for now that would be great!

AlexAlt: Sorry if this is annoying but can we add `white-space: pre-line;` to this? I am adding it to a Tooltip popper for the email themes work so I can put a line break in, and I don't want it to be busted when we switch to use the design system!

davidmferris: Done! Thanks for the heads up @AlexAlt

AlexAlt: 🙏 bless

src/Tooltip/Tooltip.jsx
```diff
@@ -0,0 +1,99 @@
...
+      this.handleClickOutside,
+    );
+
+    this.setState((state) => ({ visible: !state.visible }), this.handleShow);
+  };
+
+  render() {
+    return (
+      <Popper
+        dark={this.props.theme !== 'light'}
```
rroppolo, line 80: 💡 why not this.props.theme === 'dark', in case we ever have more themes? (We don't, so feel free to merge as is)

rroppolo (approved):
LGTM! Thanks for creating this!

src/Tooltip/Tooltip.jsx
davidmferris: I was just thinking since `theme` is a string we should serve the default dark theme in case there's a typo or something but I'm not tied to that.

rroppolo: Ahh got it - yeah this is fine with me. If we get more we can just export a constant but honestly probably concerning if we have more that two tooltip themes anyway haha :D

Issue #63, opened 2020-10-16
Tooltip text prop should be node or string


AlexAlt:
allows us to use linebreaks 🎉

PR #64, opened 2020-10-16
UIDS-63 update tooltip text prop

AlexAlt:
closes https://github.com/user-interviews/ui-design-system/issues/63

result of https://github.com/user-interviews/rails-server/pull/4678/

### DIFF
--- a/src/Popper/Popper.scss
+++ b/src/Popper/Popper.scss
@@ -9,7 +9,6 @@
   max-width: 15rem;
   letter-spacing: normal;
   padding: 0.75rem;
-  white-space: pre-line;
   z-index: $z-index-flash;
 
   &__header {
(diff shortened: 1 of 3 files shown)


src/Popper/Popper.scss
```diff
@@ -9,7 +9,6 @@
   max-width: 15rem;
   letter-spacing: normal;
   padding: 0.75rem;
-  white-space: pre-line;
```
AlexAlt, line 12: we don't need this if we can do JSX

src/Tooltip/Tooltip.jsx
```diff
@@ -32,7 +32,7 @@ export default class Tooltip extends Component {
     iconClasses: PropTypes.string,
     placement: PropTypes.string.isRequired,
     strategy: PropTypes.string,
-    text: PropTypes.string.isRequired,
+    text: PropTypes.oneOfType([propTypes.string, propTypes.node]).isRequired,
```
davidmferris, line 35: Might be worth adding a "story" in Tooltip.stories.jsx so we can see how it behaves if you pass in HTML? Lemme know if you want a hand with that.

rroppolo (approved):
I like Dave's story idea just to help people discover this, but yeah LGTM! 👍

davidmferris (approved):
Looks good!

PR #66, opened 2020-10-22
UIDS-65 Initial commit of form component with client-side validation

rroppolo:
Still figuring the ideal interface out but here's an idea

### DIFF
--- a/src/Form/ValidatedForm.jsx
+++ b/src/Form/ValidatedForm.jsx
@@ -1,0 +1,51 @@
+import React, { useEffect } from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import { useForm } from 'react-hook-form';
+
+export default function ValidatedForm(props) {
+  const { errors: serverErrors } = props;
+  const {
+    register,
+    control,
+    errors,
+    handleSubmit,
+    setError,
+  } = useForm();
+
+  useEffect(() => {
+    if (serverErrors) {
+      Object.keys(serverErrors)
+        .forEach((errorKey) => {
+          setError(errorKey, {
+            type: 'custom',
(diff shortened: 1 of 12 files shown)


src/Form/ValidatedForm.jsx
```diff
@@ -0,0 +1,51 @@
...
+  const {
+    register,
+    control,
+    errors,
+    handleSubmit,
+    setError,
+  } = useForm();
+
+  useEffect(() => {
+    if (serverErrors) {
```
davidmferris, line 17: Should this check for keys? an empty object is truthy, no?

stories/Form.stories.jsx
```diff
@@ -59,3 +62,81 @@ export const Inline = () => (
...
 );
+
+const options = [
+  { label: 'Bob', value: 1 },
+  { label: 'Dennis', value: 2 },
+  { label: 'Basel', value: 3 },
+];
+
+export const Validated = () => {
+  const [serverErrors, setServerErrors] = useState({});
```
davidmferris, line 73: If we were to use this implementation in the app, would the `useEffect` in `ValidatedForm` fire on every render or does the default `serverErrors` point to the same `{}` in memory until it's updated? I'm not actually sure of this.

src/Form/ValidatedForm.jsx
```diff
@@ -0,0 +1,51 @@
...
+
+export default function ValidatedForm(props) {
+  const { errors: serverErrors } = props;
+  const {
+    register,
+    control,
+    errors,
+    handleSubmit,
+    setError,
+  } = useForm();
```
davidmferris, line 14: Per our slack conversation, we might want to pass in a `mode` prop to handle when we show errors.

PR #71, opened 2020-10-29
UIDS-70 Add Input component

### LINKED ISSUE #70: create Input component

davidmferris:
Closes #70 

See in action [here](https://docs.userinterviews.com/ui-design-system/?path=/story/design-system-form-elements-form-group--default)

I tried to mimic all the varieties found in [this](https://www.figma.com/file/471iS3QrpPtXZpS1xbsJ20/Components?node-id=1932%3A9) Figma

### DIFF
--- a/src/FormGroup/FormGroup.scss
+++ b/src/FormGroup/FormGroup.scss
@@ -4,19 +4,30 @@
   margin: 1rem 0 0.5rem;
 
   &--is-invalid {
-    input, select, textarea {
+    .Input {
+      background-color: $ux-red-100;
+      border-color: $ux-red-600;
+    }
+
+    .InputLabel {
+      color: $ux-red-800;
+    }
+
+    select, textarea {
       border-color: $ux-red;
     }
   }
 
-  &__description {
-    padding: .25rem 0;
+  &__helper-text {
+    @include font-type-20;
+    color: $ux-gray-800;
+    padding: .375rem 0;
   }
 
   &__invalid-feedback {
-    color: $ux-red;
-    display: block;
-    font-size: 100%;
+    @include font-type-20;
+    color: $ux-red-800;
+    padding: .375rem 0;
 
     &__list {
       @include ui-remove-list-styles();
--- a/src/Input/Input.jsx
+++ b/src/Input/Input.jsx
@@ -1,0 +1,80 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+
+import './Input.scss';
+
+export default function Input(props) {
+  const {
+    disabled,
+    id,
+    leadingIcon,
+    name,
+    placeholder,
+    trailingIcon,
+    type,
+    value,
+    onChange,
+    ...rest
+  } = props;
+
+  const inputClasses = classNames('Input form-control', {
+    'Input--with-leading-icon': leadingIcon,
+    'Input--with-trailing-icon': trailingIcon,
+  });
+
+  return (
+    <div className="Input input-group">
+      {leadingIcon && (
+        <div className="input-group-prepend">
+          <div className="input-group-text">
+            <FontAwesomeIcon icon={leadingIcon} />
+          </div>
+        </div>
+      )}
+
+      <input
+        className={inputClasses}
+        disabled={disabled}
+        id={id}
+        name={name}
+        placeholder={placeholder}
+        type={type}
+        value={value}
+        onChange={onChange}
+        {...rest}
+      />
+
+      {trailingIcon && (
+        <div className="input-group-append">
+          <div className="input-group-text">
+            <FontAwesomeIcon icon={trailingIcon} />
+          </div>
+        </div>
+      )}
+    </div>
+  );
+}
+
+Input.propTypes = {
+  disabled: PropTypes.bool,
+  id: PropTypes.string.isRequired,
+  leadingIcon: PropTypes.object,
+  name: PropTypes.string.isRequired,
+  placeholder: PropTypes.string,
+  trailingIcon: PropTypes.object,
+  type: PropTypes.string,
+  value: PropTypes.string,
+  onChange: PropTypes.func,
+};
+
+Input.defaultProps = {
+  disabled: false,
+  leadingIcon: undefined,
+  placeholder: '',
+  trailingIcon: undefined,
+  type: 'text',
+  value: '',
+  onChange: undefined,
+};
--- a/src/Input/Input.scss
+++ b/src/Input/Input.scss
@@ -1,0 +1,25 @@
+@import '../../scss/theme';
+
+.Input {
+  @include font-type-30--light;
+  box-shadow: $input-box-shadow;
+
+  ::placeholder {
+    color: $ux-gray-800;
+  }
+
+  &:focus {
+    border: 1px solid #92CBFC;
+    box-shadow: inset 0px 1px 1px rgba(0, 0, 0, 0.075);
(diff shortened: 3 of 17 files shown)


src/InputLabel/InputLabel.jsx
```diff
@@ -2,6 +2,8 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 
+import Tooltip from 'src/Tooltip';
+
 import './InputLabel.scss';
```
rroppolo, line 7: Is InputLabel's font too small? I see it as type-20 but it looks like Faustino is using type-30?

rroppolo: And I think the error text on the figma is type-20?

src/Input/Input.scss
```diff
@@ -0,0 +1,15 @@
...
+    background-color: $ux-white;
+  }
+
+  &--with-leading-icon {
+    border-left: none;
+  }
+
+  &--with-trailing-icon {
+    border-right: none;
+  }
```
rroppolo, line 14: I think the placeholder text is type-30 instead of type-40?

src/Input/Input.scss
```diff
@@ -0,0 +1,15 @@
+@import '../../scss/theme';
+
+.Input {
+  .input-group-text {
+    background-color: $ux-white;
```
fausgaitan, line 5: For the normal/inactive and filled states we should have a very subtle inner shadow.
image
This was added very recently!
`box-shadow: inset 0px 2px 2px rgba(225, 228, 232, 0.65);`

src/Input/Input.scss
fausgaitan: This is also a little subtle but we need to update the font size for the placeholder and the value to type-30 LIGHT. and also differentiate the colors:

image

fausgaitan: Can we also make sure the padding / distances between the text and the input borders are 12px to the left and 8px above and below?
image

src/FormGroup/FormGroup.scss
```diff
@@ -4,17 +4,27 @@
...
 
   &--is-invalid {
-    input, select, textarea {
+    .Input {
+      background-color: $ux-red-100;
+      border-color: $ux-red;
+    }
+
+    .InputLabel {
+      color: $ux-red-800;
```
fausgaitan, line 13: We've decided to update the label to be bold now so it should be `.font-type-30--bold` also the distance between the label and input field should be 6px.

image

src/FormGroup/FormGroup.scss
```diff
@@ -4,17 +4,27 @@
   margin: 1rem 0 0.5rem;
 
   &--is-invalid {
-    input, select, textarea {
+    .Input {
+      background-color: $ux-red-100;
+      border-color: $ux-red;
```
fausgaitan, line 9: we should update the error state with these colors:

image

I've also noticed in storybook the error state the error text is different than the helper text. Can we remove the helper text when the error message shows up? and have the same font-size as the helper text? 

image

src/Input/Input.scss
fausgaitan: we need to add a variant with both the trailing icon and the leading icon exposed:
image

fausgaitan (changes requested):
I added some comments.

Thanks so much for updating this!!!

src/FormGroup/FormGroup.scss
```diff
@@ -4,17 +4,27 @@
...
+    .InputLabel {
+      color: $ux-red-800;
+    }
+
+    select, textarea {
       border-color: $ux-red;
     }
   }
 
   &__description {
```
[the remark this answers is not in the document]
davidmferris: Yeah I can change the name to helper-text. There are currently a few usages of the `description` prop in the app, so I'm going to keep both for now to preserve backwards compatibility, but I'll eventually remove `description` altogether.

src/Input/Input.jsx
```diff
@@ -0,0 +1,76 @@
...
+
+Input.propTypes = {
+  disabled: PropTypes.bool,
+  leadingIcon: PropTypes.object,
+  name: PropTypes.string.isRequired,
+  placeholder: PropTypes.string,
+  trailingIcon: PropTypes.object,
+  type: PropTypes.string,
+  value: PropTypes.string,
+  onChange: PropTypes.func.isRequired,
```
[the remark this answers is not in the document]
davidmferris: Ah yeah, I wasn't sure if there was a case where we wouldn't want that but it sounds like there is!

src/Input/Input.scss
davidmferris: Very nice!

src/InputLabel/InputLabel.jsx
```diff
@@ -21,10 +24,12 @@ InputLabel.propTypes = {
...
   required: PropTypes.bool,
   text: PropTypes.string.isRequired,
+  tooltipText: PropTypes.string,
 };
 
 InputLabel.defaultProps = {
   className: '',
   labelHtmlFor: '',
   required: false,
+  tooltipText: '',
```
rroppolo, line 34: maybe undefined instead?

src/InputLabel/InputLabel.scss
```diff
@@ -1,11 +1,10 @@
...
 .InputLabel {
-  @include font-type-20;
-  font-weight: $font-weight-normal;
-  margin-bottom: 0.5rem;
+  @include font-type-30--bold;
+  margin-bottom: 0.375rem;
 
   &__helper-text {
-    color: $ux-gray-400;
+    color: $ux-gray-600;
```
rroppolo, line 8: I think this needs a font-type-30 or something so it doesn't pick up the bold from the one above?

src/Input/Input.jsx
```diff
@@ -0,0 +1,77 @@
...
+
+  return (
+    <div className="Input input-group">
+      {leadingIcon && (
+        <div className="input-group-prepend">
+          <div className="input-group-text">
+            <FontAwesomeIcon icon={leadingIcon} />
+          </div>
+        </div>
+      )}
```
rroppolo, line 34: Screen Shot 2020-10-30 at 5 29 39 PM
I noticed this but not sure how much of a pain it is to go around the icon part...

rroppolo (commented):
Looking great! I commented like two super small things but code-wise I think it looks good. I defer to Faustino to give the final 👍

src/InputLabel/InputLabel.scss
davidmferris: Ah yeah good catch.

jojolee1996:
image 3
Minor tweak, but would you be able to change (Required) to the regular weight?

src/Input/Input.jsx
davidmferris: Yeah, this is tricky. `input-group-prepend` makes this a flex box so I think in order to add the outline to the icon divs we might have to hack into the input's onfocus event and add some extra styling with javascript, which i'm a little reluctant to do. FWIW [Bootstrap](https://getbootstrap.com/docs/4.0/components/input-group/) (which provides the `input-group-prepend` class) styles their inputs so that the icon divs don't receive the focus highlighting. What do you think @fausgaitan and @jojolee1996 ?

fausgaitan: I like how bootstrap does it! Is it possible to add a variation for the icons to be in a container like the screenshot below? (if it's too much work, don't worry!)
image

I see there's a subtle vertical shadow here, is possible to remove that for the "no container" versions?
image

fausgaitan (approved):

jojolee1996 (approved):

PR #72, opened 2020-10-30
Intro page using MDX

fausgaitan:


src/components/Intro.stories.mdx
```diff
@@ -0,0 +1,42 @@
+import { Meta } from '@storybook/addon-docs/blocks';
```
davidmferris, line 1: I wonder if it'd make sense to just have this file be stories/Intro.stories.mdx or something like that. Anything that gets placed under src/components gets bundled up into the package that we install in the app so I don't think we'd want this to be there.

fausgaitan: I just moved it to the stories folder. 👍

stories/Intro.stories.mdx
```diff
@@ -0,0 +1,43 @@
...
+<Meta title="Design System/Intro" />
+
+<div style={{ width: '50%', margin: '0 auto', textAlign: 'center'}}>
+<img src={ImageFile} />
+
+We aim to create a system that let us build a user experience that is consistent, flexible and scalable. 
+This style guide is living documentation that will be updated as we continue to evolve and improve our design system.
+
+<a type="button" style={{ 
+    background:'#f1f1f1' ,
```
fausgaitan, line 16: @davidmferris / @rroppolo 

I tried to replace this `#f1f1f1 `color using the variable `$ux-gray-200` that is defined in `scss/colors/palette.scss` but it doesn't work... everything breaks!! What's the correct way to do it? 

Thanks for your help!!

rroppolo: Hey @fausgaitan! You could try adding:
```
import { colors } from 'src/Styles';
``` 
and replace this with 
```
background: colors.uxGray200
```

I didn't actually test this so let us know if that doesn't do it!

fausgaitan: that worked! thanks so much!

One more thing, [these buttons](https://www.figma.com/file/471iS3QrpPtXZpS1xbsJ20/Components?node-id=1525%3A5) are not available in storybook yet, right? 

What if I wanted to update the text style with this variable: `font-type-30`? how do I make them available in this file?

sorry for all these basic questions!!

.storybook/main.js
```diff
@@ -7,5 +7,9 @@ module.exports = {
     '@storybook/addon-links/register',
     '@storybook/addon-storysource/register'
   ],
-  stories: ['../stories/**/*.stories.js[x]'],
+  stories: [
+    '../stories/Intro.stories.mdx',
```
davidmferris, line 11: Is this line necessary? I'd expect the line just below to pick up the story because of the wildcard.

stories/Intro.stories.mdx
```diff
@@ -0,0 +1,44 @@
+import { Meta, Story, Canvas } from '@storybook/addon-docs/blocks';
+import ImageFile from '../public/ui-design-system.svg';
+import FigmaLogo from '../public/figma-logo.svg';
+import GithubLogo from '../public/github-logo.svg';
+import { colors } from 'src/Styles';
+
+<Meta title="Design System/Intro" />
+
+<div style={{ width: '50%', margin: '0 auto', textAlign: 'center'}}>
+<img src={ImageFile} />
```
davidmferris, line 10: I'm not sure if this would break anything because of markdown, but if it doesn't could we indent everything in between the `<div>` tags so it's a little easier to read?

davidmferris (approved):
Looks good to me pending two small questions!

.storybook/main.js
fausgaitan: I removed the line and everything seems to be working ok.

stories/Intro.stories.mdx
fausgaitan: I added indentation but it breaks the page for some reason.. 

Let me know if you find any other issue, otherwise, if you think this is ok it should be ready to merge.

thanks so much!!

davidmferris: Cool yeah I had a feeling that might happen. This is fine then!

davidmferris (approved):
Looks good to me!

PR #76, opened 2020-11-12
UIDS-75 remove description prop from FormGroup

### LINKED ISSUE #75: Remove description prop from FormGroup
Once all uses of FormGroup in the main app have been updated to use the helperText prop we should deprecate `description`.

davidmferris:
Closes #75

### DIFF
--- a/src/FormGroup/FormGroup.jsx
+++ b/src/FormGroup/FormGroup.jsx
@@ -22,12 +22,8 @@ function renderErrors(errors) {
 }
 
 export default function FormGroup(props) {
-  const { errors, inputKey } = props;
+  const { errors, helperText, inputKey } = props;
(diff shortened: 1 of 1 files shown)


rsaris (approved):
Does this come with the version bump or was that version bump somewhere else?

davidmferris:
This is such a minor change I figured I'd either bump a minor version later or include with the next major update (like adding Card). It's backwards compatible so it doesn't harm anything to wait on this.

rsaris:
Will this not break the release like it did last time we changed this repo w/o updating the version number?

davidmferris:
No, that happened because i manually updated the package tarball without updating the refs in the app. You can merge as much stuff as you want safely here without needing to bump the version every time.

rroppolo (approved):

PR #78, opened 2020-11-18
UIDS-76 - Add docs to inputs section

fausgaitan:
**Docs page:**

For reference, we need to add [these guidelines](https://www.notion.so/userinterviews1/JeJo-PMM-huddle-f8d47e0de3604237af9d08202d6fd4d9)

image

image


stories/Form Elements/FormGroup.stories.mdx
```diff
@@ -0,0 +1,29 @@
...
+import Input from 'src/Input';
+import InputLabel from 'src/InputLabel';
+
+<Meta title="Design System/Form Elements" component={FormGroup} />
+
+# Input fields
+
+Input fields let users enter and edit text.
+
+<Story name="Input Fields">
```
fausgaitan, line 15: Hey @rroppolo and @davidmferris ,

I followed [these steps](https://www.learnstorybook.com/design-systems-for-developers/react/en/document/) to use MDX like I did with the Intro file, but in this case, the stories need to have this format ( <Story... )

In the original file <InputComponent> is defined like this:

```
const InputComponent = (props) => {
  const [value, setValue] = useState('');
  const handleChangeValue = (event) => setValue(event.target.value);

  return (
    <Input value={value} onChange={handleChangeValue} {...props} />
  );
};
```

Do you know how this could be translated into MDX format? I'm stuck here.  The components are not rendering for this reason (I guess).

Any help is appreciated! thanks!!

davidmferris: I think you'll need to add an `export` to the `InputComponent` function in FormGroup.stories.jsx and then import it in this file like:
`import { InputComponent } from './FormGroup.stories';`

fausgaitan: that worked! thanks so much! @davidmferris

stories/Form Elements/FormGroup.stories.mdx
```diff
@@ -0,0 +1,141 @@
...
+<Preview>
+<Story name="Error">
+  <FormGroup
+    displayErrorText
+    errors={{ test: ['invalid input!'] }}
+    id="with-errors"
+    inputKey="test"
+    label="Label"
+    labelHtmlFor="input"
+  >
```
fausgaitan, line 137: @davidmferris / @rroppolo 

Is it possible to add a link in the error message like this? Not sure if it's possible right now. 

image

rroppolo: The errors prop on FormGroup is just validated as an object, so it looks like it does work if you pass jsx in instead of a string. For example, I just tried this test in the errors story:
```
export const WithErrors = () => (
  <FormGroup
    displayErrorText
    errors={{ test: [<div>invalid input! <a href="http://www.google.com">And a link!</a></div>] }}
```
Screen Shot 2020-12-08 at 9 50 45 AM

fausgaitan: that worked! thanks!

stories/Alert.stories.mdx
```diff
@@ -0,0 +1,94 @@
...
+
+[ info alert story should go here ]
+
+<Preview>
+<Story name="Info">
+<AlertMessage
+    id="2"
+    message={text('Message', 'Default message')}
+    title={text('Title', 'Default title')}
+    type={radios('Message Type', MessageTypes, MessageTypes.INFO)}
```
fausgaitan, line 43: @rroppolo / @davidmferris 

Final question for this PR (hopefully) 

I added the guidelines for alerts too. And the idea is to display the different variants (success, info, warning, and error here) I updated this little part to `MessageTypes.INFO` but it's still displaying the success alert. 
I'm not sure what I'm doing wrong. 👎

stories/Alert.stories.jsx
```diff
@@ -8,30 +8,11 @@ import 'bootstrap/dist/css/bootstrap.min.css';
...
-  <AlertMessage
-    id="1"
-    message={text('Message', 'Default message')}
-    title={text('Title', 'Default title')}
-    type={radios('Message Type', MessageTypes, MessageTypes.SUCCESS)}
-  />
-);
-
-const onDismiss = (id) => {
+export const onDismiss = (id) => {
```
davidmferris, line 16: Hmm this is now showing up in the Storybook sidebar. I think by default it adds an entry any time it sees `export const` in a story file. Could this be defined in the MDX file? If not we might want to do something like [this](https://storybook.js.org/docs/react/api/mdx#embedding-stories) so we can keep all the JSX in this file.

stories/Alert.stories.mdx
```diff
@@ -0,0 +1,91 @@
+import { withKnobs, radios, text } from '@storybook/addon-knobs';
```
davidmferris, line 1: I noticed the knobs aren't working on this story now. Any idea why not? You should be able to add text in the "knobs" tab and see it update live on the component.

stories/Alert.stories.mdx
```diff
@@ -0,0 +1,91 @@
...
+Alerts give users updates or changes to system status. Providing immediate feedback is important for building trust. While alerts are a great way of communicating with users, they are disruptive and should be used cautiously.
+
+## Success
+
+- Used when something went as planned. Ex. When you've saved your progress in the builder. 
+
+<Preview>
+<Story name="Success">
+<AlertMessage
+    id="1"
```
davidmferris, line 25: I think the indentation is a little off on most of the components in these two MDX files. One benefit of keeping these defined in the JSX file is that our linter will pick up those issues.

stories/Form Elements/FormGroup.stories.jsx
```diff
@@ -14,7 +14,8 @@ export default {
   decorators: [withKnobs],
 };
 
-const InputComponent = (props) => {
+
+export const InputComponent = (props) => {
```
davidmferris, line 18: Same issue with this export. We don't want this to show up in the Storybook sidebar I think because this should just be a wrapper component.

davidmferris (changes requested):
Sorry for the delay @fausgaitan. I noticed a couple of issues while testing this out in Storybook that I think we might want to address.

stories/Alert.stories.jsx
davidmferris: Oh I just found [this](https://storybook.js.org/docs/react/api/csf#non-story-exports) as well which I think should fix this issue.

stories/Alert.stories.mdx
jhforster: @fausgaitan do you have the linter set up? If not, might be worth having @davidmferris or @rroppolo help with that!

stories/Alert.stories.jsx
```diff
@@ -8,30 +8,12 @@ import 'bootstrap/dist/css/bootstrap.min.css';
 import '../scss/global.scss';
 
 export default {
-  title: 'Design System/Alert',
+  title: 'Design System/Alerts',
   component: AlertMessage,
   decorators: [withKnobs],
+  excludeStories: ['onDismiss']
```
fausgaitan, line 14: @davidmferris I added this line here and now the item doesn't appear in the sidebar.

image

stories/Alert.stories.mdx
fausgaitan: Apparently, they are not working in the input fields section either 

I found this issue. One user says "Would be super cool to have knobs in the MDX stories!" 
Maybe I'm wrong but it looks like they've released "addon-controls"  (beta) that seems to allow the usage of knobs in mdx files?

stories/Alert.stories.mdx
fausgaitan: I don't have it. :/ So, if any of them can help I'd appreciate it! thanks!

stories/Form Elements/FormGroup.stories.jsx
fausgaitan: I'm not sure which item this line is adding to the sidebar?

rroppolo:
Hey @fausgaitan ! Would you mind checking out my changes to this when you get a sec? I think I kept the stories the way you had them set up but let me know or feel free to make changes if I missed anything!

.eslintrc
```diff
@@ -1,12 +1,23 @@
...
+        "semi": 0
+      }
+    },
+    {
+      "files": [
+        "stories/**/*.mdx",
+        "stories/**/*.stories.mdx"
+      ],
+      "rules": {
+        "react/jsx-indent": 0
```
rroppolo, line 20: I turned off indentation linting on the mdx files for now because it wasn't working well. Hopefully can reenable soon but didn't spend a lot of time yet, am probably missing something.

stories/Alert/Alert.stories.jsx
```diff
@@ -0,0 +1,70 @@
...
+import { action } from '@storybook/addon-actions';
+
+import { MessageTypes, AlertMessage } from 'src/Flash';
+import mdx from './Alert.mdx';
+
+import 'bootstrap/dist/css/bootstrap.min.css';
+import '../../scss/global.scss';
+
+export default {
+  title: 'Design System/Alert',
```
rroppolo, line 12: For now I kept the name Alert vs. Alerts but probably we need to take a pass over out left-hand navigation and make sure we have a good convention in place. For now I figured we could do that in a separate PR though if that's cool. Technically the React component is called 'AlertMessage' and it lives under 'Flash/' so might want to revisit that too..

stories/Alert/Alert.stories.jsx
```diff
@@ -0,0 +1,70 @@
+import React from 'react';
```
rroppolo, line 1: This all existed btw but I moved the file into a directory so it looks like it was added

stories/Form Elements/FormGroup.stories.jsx
```diff
@@ -33,73 +40,55 @@ const InputComponent = (props) => {
...
+    label="Label"
+    labelHtmlFor="input"
   >
-    <InputComponent id="with-required-input" name="required" placeholder="Text is required" />
+    <InputComponent id="input" name="with-helper-text" />
   </FormGroup>
 );
 
-export const WithLabel = () => (
+export const Required = () => (
```
rroppolo, line 58: Moved around the order to better match the docs section here

rsaris (commented):
This looks awesome, is there any way we can reconcile the differences in the babel imports in `yarn.lock`? Adding 3k lines to the file w/ a very small change to `package.json` feels like we did something odd. I imagine if we manually remove the `babel/core` blocks in yarn.lock it'll make it so they all use the same version of `babel/core` and we won't need to duplicate so much content there.

rsaris (approved):

fausgaitan:
LGTM @rroppolo !

PR #82, opened 2020-11-24
UIDS-81 New CheckboxButtonGroup component

rroppolo:
Fixes #82 

Screen Shot 2020-11-23 at 9 50 26 AM

Also adds a few tweaks to other components that came up in testing of the PRoject Overview page


.babelrc
```diff
@@ -4,6 +4,7 @@
     "@babel/preset-react"
   ],
   "plugins": [
+    "@babel/plugin-transform-runtime",
```
rroppolo, line 7: I was seeing an error related to runtime after adding a test that had an async act call

jest.config.js
```diff
@@ -77,6 +77,7 @@ module.exports = {
 
   // A map from regular expressions to module names that allow to stub out resources with a single module
   moduleNameMapper: {
+    "\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga|mdx)$": "<rootDir>/spec/__mocks__/fileMock.js",
```
rroppolo, line 80: Was seeing the test suite fail due Storyshot test trying to understand the .mdx file. Added this based on https://jestjs.io/docs/en/webpack

src/FormGroup/FormGroup.scss
```diff
@@ -3,9 +3,21 @@
...
+      @include font-type-40;
+      margin-bottom: 0.75rem;
+    }
+  }
+
   &--is-invalid {
     .Input {
-      border-color: $ux-red-600;
+      // important because Input.scss focus style also sets this border
+      border-color: $ux-red-600 !important;
```
rroppolo, line 20: I don't like this but it works.. another option I can think of is to rework how these components interact and pass an error prop down to the input itself.

src/CheckboxButtonGroup/CheckboxButton.scss
```diff
@@ -0,0 +1,24 @@
+@import '../../scss/theme';
+
+.CheckboxButton {
+  border: 1px solid $ux-gray-400;
```
davidmferris, line 4: Is the bordered version our default checkbox? Do you think we would want a variant (or default) that is just the checkbox and label as well?

davidmferris (commented):
This is beautiful! I have a few comments but looks great otherwise.

src/CheckboxButtonGroup/CheckboxButton.scss
rroppolo: Makes sense! Added about as a `bordered` bool prop like I had for FormGroup that is false by default

src/CheckboxButtonGroup/CheckboxButton.jsx
```diff
@@ -0,0 +1,62 @@
...
+  className,
+  disabled,
+  label,
+  id,
+  name,
+  value,
+  onChange,
+  ...rest
+}) => (
+  <label
```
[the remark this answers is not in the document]
rroppolo: Ooh thanks for the heads up, this actually aligned really nicely with FormControlLabel and let me leverage some of that component's css I had also added here originally.  Yay!

src/CheckboxButtonGroup/CheckboxButton.jsx
```diff
@@ -0,0 +1,69 @@
...
+    htmlFor={id}
+    labelHtmlFor={id}
+    text={label}
+  >
+    <input
+      checked={checked}
+      className={className}
+      disabled={disabled}
+      id={id}
+      name={name}
```
davidmferris, line 35: Worth adding an `indeterminate` prop? I know we use them in a couple places.

src/Form/Form.jsx
```diff
@@ -18,12 +18,15 @@ export default function Form(props) {
...
   inline: PropTypes.bool,
-  onSubmit: PropTypes.func.isRequired,
+  onSubmit: PropTypes.func,
 };
 
 Form.defaultProps = {
   children: undefined,
+  className: undefined,
   inline: false,
+  onSubmit: undefined,
```
davidmferris, line 31: Is this because of react-hook-form?

src/FormGroup/FormGroup.scss
davidmferris: Hmm yeah I think passing the error down to the input is probably ideal. Fine by me if we want to handle that in a follow up.

src/CheckboxButtonGroup/CheckboxButton.jsx
```diff
@@ -0,0 +1,69 @@
...
+  className,
+  disabled,
+  label,
+  id,
+  name,
+  value,
+  onChange,
+  ...rest
+}) => (
+  <FormControlLabel
```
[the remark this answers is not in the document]
rroppolo: Like this! As chatted over slack changed to `<FormControlLabel Control={CheckboxButton/RadioButton/etc}`. I am leaving of touching RadioButton for now just to limit this branch but I think it will be easy to share and I suspect we may encounter a bordered radio somewhere else in the buidler but haven't checked for it yet. Lemme know what you think!

src/CheckboxButtonGroup/CheckboxButton.jsx
rroppolo: Yup added though it also spreads any additional props so will still work if there's a one not enumerated here.

src/FormGroup/FormGroup.scss
rroppolo: Hmm it is actually trickier than I thought b/c the `Input` is a child of `FormGroup` but there is no guarantee around number or type of children.. to pass an additional prop we would have to clone them or give `FormGroup` more control over their rendering.. I may leave this for now and come back to it to not hold up this branch, but I do think it is worth exploring more

stories/Form Elements/FormControlLabel.stories.jsx
```diff
@@ -1,17 +1,18 @@
 import React from 'react';
 
-import FormControlLabel from 'src/FormControlLabel';
+import { CheckboxButton } from 'src/CheckboxButtonGroup';
 
 export default {
   title: 'Design System/Form Elements/Form Control Label',
```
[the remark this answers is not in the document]
rroppolo: Does that still make sense now that the parent component is FormControlLabel? Actually most of the styles are there now & CheckboxButton is just a component wrapper around a bare input (no label), so now I could maybe see RadioButton story moving in here rather than the reverse?

rroppolo:
Thanks @davidmferris ! Latest changes here: https://github.com/user-interviews/ui-design-system/pull/82/commits/67100124c392bbb1d6b4af6109a1ed0c53cc5cf5

rroppolo:
Hey @jojolee1996 ! Would you mind checking out the CheckboxButtonGroup component here? https://docs.userinterviews.com/ui-design-system/?path=/story/design-system-form-elements-form-group--with-checkbox-button-group Intended for use in the new builder for things like "Device requirements"

src/CheckboxButton/CheckboxButton.jsx
```diff
@@ -0,0 +1,52 @@
...
+  value,
+  onChange,
+  ...rest
+}) => (
+  <input
+    checked={checked}
+    className={className}
+    disabled={disabled}
+    id={id}
+    name={name}
```
davidmferris, line 20: Think you're missing the `indeterminate` prop here, but it's also cool if you'd rather just have it be part of `rest`.

stories/Form Elements/FormControlLabel.stories.jsx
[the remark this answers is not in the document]
davidmferris: Yeah I think you're right. Probably makes sense now to keep this as is and add `RadioButton` in here.

src/CheckboxButtonGroup/CheckboxButton.jsx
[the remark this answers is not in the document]
davidmferris: Yeah this is great. Exactly what I had in mind!

src/CheckboxButton/CheckboxButton.jsx
rroppolo: Whoops you are right I meant to add it, will do so

jojolee1996:
Thanks Rachel! Dropping some design tweaks below:
checkbox

stories/Form Elements/FormControlLabel.stories.jsx
```diff
@@ -1,17 +1,23 @@
...
   component: FormControlLabel,
+  decorators: [withKnobs()],
 };
 
 export const Checkbox = () => (
   <FormControlLabel
-    labelHtmlFor="checkbox"
+    bordered={boolean('Bordered button', false)}
+    Control={CheckboxButton}
```
davidmferris, line 16: Would it be possible to add an `action` that sets `checked` here so we can see the styling in this component story too?

davidmferris (approved):
Looks awesome!! Just one suggestion for the FormControlLabel story.

stories/Form Elements/FormControlLabel.stories.jsx
rroppolo: Added 👍

PR #95, opened 2020-12-01
UIDS-39 Expose form css classes outside component

rroppolo:
Separate branch to expose form component classes in scss so they can be shared with rails-server `ui_form.scss` instead of copied over.

Also includes:
- Update bootstrap variables to not highlight input on focus with the green (bootstrap primary) color.
- Remove FormGroup margin-bottom
- Suppose wrapping of CheckboxButtonGroup on mobile and updated the story to be able to test with longer lables

Paired with this PR and its review app which show the places that I could find within the app that we use these Form classes outside of the component: https://github.com/user-interviews/rails-server/pull/5155

https://user-interviews-stagin-pr-5155.herokuapp.com/


davidmferris (approved):
LGTM!

rsaris (approved):

src/FormGroup/FormGroup.scss
```diff
@@ -1,56 +1,3 @@
 @import '../../scss/theme';
```
rsaris, line 1: :question: Is there any reason this CSS file should exist at all or should the component just import the common CSS file? Same question for the sibling files.

rroppolo: That is an excellent point and honestly I just didn't even think of it. I can't think of any reason! Probably better to remove them to avoid the temptation to put something in there.

PR #98, opened 2020-12-03
CHORE Add forwardRef to RadioButton and some styling tweaks

rroppolo:
Added forwardRef for use with React Hook Form and a few styling tweaks:
- Radio buttons that are part of a RadioButtonGroup should have bold text
- Added row orientation to RadioButtonGroup
- Additional padding b/t buttons in RadioButtonGroup

Unclear if we may need more tweaks after the first radio goes through review but I think these are the most obvious ones. I paused the refactoring of RadioButton for now due to time constraints as it wasn't blocking yet.

### DIFF
--- a/src/FormControlLabel/FormControlLabel.jsx
+++ b/src/FormControlLabel/FormControlLabel.jsx
@@ -4,7 +4,7 @@ import classnames from 'classnames';
 
 import 'scss/forms/form_control_label.scss';
 
-export default function FormControlLabel({
+const FormControlLabel = React.forwardRef(({
   bordered,
   checked,
   className,
@@ -12,10 +12,9 @@ export default function FormControlLabel({
   id,
   text,
   ...controlProps
-}) {
-  return (
-    <label
-      className={classnames(
+}, ref) => (
+  <label
+    className={classnames(
         'FormControlLabel',
         className,
         {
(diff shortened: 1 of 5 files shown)


stories/Form Elements/RadioButton.stories.jsx
```diff
@@ -28,3 +35,28 @@ export const WithChildren = () => (
...
+      Object.values(ORIENTATIONS),
+      ORIENTATIONS.COLUMN,
+    )}
+  >
+    <RadioButton
+      bordered
+      id="radio-1"
+      label={text('First Label', 'First option with child')}
+    >
+      <span style={{ color: colors.uxGray500 }}>Some descriptive text for the option above</span>
```
davidmferris, line 52: Looks like this color is failing the accessibility check. Maybe use something darker to clean up the noise?

davidmferris (approved):
Looks great! One story-related comment but that's it.

stories/Form Elements/RadioButton.stories.jsx
rroppolo: Bumped up to 800 which matches some similar text in the prototype - thanks for the heads up!

rroppolo:
Not many changes, just added a few more forwardRefs

src/FormControlLabel/FormControlLabel.jsx
```diff
@@ -45,3 +43,7 @@ FormControlLabel.defaultProps = {
   checked: undefined,
   className: undefined,
 };
+
+FormControlLabel.displayName = 'FormControlLabel';
```
davidmferris, line 47: What is the `displayName` if we don't set it explicitly here?

davidmferris (approved):
Looks great!

src/FormControlLabel/FormControlLabel.jsx
rroppolo: It comes as "ForwardRef" or something like that - was changing the app snapshots

davidmferris: Ah gotcha, I guess the React wrapper hijacks that.

PR #101, opened 2020-12-07
UIDS-100 create CopyToClipboardButton

jamisonordway:
This adds a `CopyToClipboardButton` component which allows the user to copy some text by clicking anywhere on the button.


scss/buttons.scss
```diff
@@ -7,3 +7,31 @@
     color: $ux-gray-700;
   }
 }
+
+.btn-outline-secondary {
```
davidmferris, line 11: Hmm it's unfortunate that we haven't added button styling to the design system yet, but you should be able to leverage Bootstrap's styling instead of needing to write it all yourself. I don't know exactly how we have it set up in the app, but I think if you look at ui_buttons.scss in the main app you should be able to cut over what you need here. The design system does have Bootstrap installed so you can get their styles by doing:
`@import '~bootstrap/scss/buttons';`

src/CopyToClipboardButton/CopyToClipboardButton.jsx
```diff
@@ -0,0 +1,58 @@
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faCopy } from '@fortawesome/free-regular-svg-icons';
+import { CopyToClipboard as ReactCopyToClipboard } from 'react-copy-to-clipboard';
```
davidmferris, line 5: You shouldn't need the `as` here. I had to do that in the other component because there was a naming conflict with `CopyToClipboard`.

scss/buttons.scss
```diff
@@ -7,3 +7,31 @@
...
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  padding: 8px 116px;
+
+  position: static;
+  width: 351px;
```
davidmferris, line 18: For sizes and stuff, I don't think we'll actually want to define those in the design system anyways. We should allow the button to be variable width and let the rails-server app determine that depending on how we're using it.

src/CopyToClipboardButton/CopyToClipboardButton.jsx
```diff
@@ -0,0 +1,58 @@
...
+  const handleClickCopy = () => {
+    setCopied(true);
+    setTimeout(() => setCopied(false), 1000);
+  };
+
+  return (
+    <div className="CopyToClipboardButton">
+      <ReactCopyToClipboard text={props.copyText} onCopy={handleClickCopy}>
+        <TrackedButton
+          className="btn-outline-secondary"
```
davidmferris, line 24: Rather than hard-coding a class here what do you think about allowing this component to accept a `variant` prop? Then we could map that to a `className`. That way this component could be used for multiple styles. Here's an [example](https://github.com/user-interviews/ui-design-system/blob/master/src/FormControlLabel/FormControlLabel.jsx#L17) of how we've done that elsewhere in the design system.

src/CopyToClipboardButton/CopyToClipboardButton.jsx
```diff
@@ -0,0 +1,58 @@
...
+function CopyToClipboard(props) {
+  const [copied, setCopied] = useState(false);
+
+  const handleClickCopy = () => {
+    setCopied(true);
+    setTimeout(() => setCopied(false), 1000);
+  };
+
+  return (
+    <div className="CopyToClipboardButton">
```
davidmferris, line 21: Now that this has been extracted from the existing `CopyToClipboard` component we should be able to plug this in there, so that'd simply be like:
```
<div className="CopyToClipboard">
  <span className="CopyToClipboard__copy-text">{copyText}</span>
  <CopyToClipboardButton copyText={copyText} trackingEvent={trackingEvent} />
</div>
```
We actually may want to consider renaming that component, but since that would be a breaking change we can hold off on that for now.

src/CopyToClipboardButton/CopyToClipboardButton.jsx
```diff
@@ -0,0 +1,58 @@
...
+          <Popper
+            dark
+            text="Copied!"
+            visible={copied}
+          >
+            <FontAwesomeIcon
+              className="CopyToClipboardButton__icon"
+              icon={faCopy}
+            />
+            <span className="CopyToClipboardButton__display-text">
```
davidmferris, line 37: Could we make `displayText` undefined by default and only render this if there's a value? That way we'll be able to use this in the existing `CopyToClipboard` component.

src/CopyToClipboardButton/CopyToClipboardButton.scss
```diff
@@ -0,0 +1,24 @@
...
+  border-radius: .25rem;
+  color: $ux-blue;
+  display: flex;
+  font-weight: $font-weight-bold;
+  justify-content: space-between;
+  min-height: 2rem;
+  padding: 0 0 0 .375rem;
+
+  &__display-text {
+    color: $ux-blue;
```
davidmferris, line 14: Similar to my comment above, we should be able to map this to a `variant` prop using the `classNames` package so we can allow this component to be styled differently depending on the situation.

src/CopyToClipboardButton/index.js
```diff
@@ -0,0 +1 @@
+export { default } from './CopyToClipboardButton';
```
davidmferris, line 1: In addition to this, you'll want to export the component in src/index.js too. That will allow us to access the component in the rails-server app.

davidmferris (changes requested):
Looking really good so far @jamisonordway! I have some comments for you about how we could make this more extendable and utilize some of our existing setup to simplify it a bit.

src/Checkbox/Checkbox.jsx
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+
+import './Checkbox.scss';
+
+export default function Checkbox(props) {
```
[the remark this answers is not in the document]
jamisonordway: Completely forgot I had these `Checkbox` changes. That's really slick, thank you!

src/CopyToClipboardButton/CopyToClipboardButton.jsx
```diff
@@ -0,0 +1,75 @@
...
+  };
+
+  return (
+    <div className="CopyToClipboardButton">
+      <CopyToClipboard text={props.copyText} onCopy={handleClickCopy}>
+        <TrackedButton
+          aria-label="Copy to clipboard"
+          className={classNames(
+            'CopyToClipboardButton',
+            `btn ${props.variant}`,
```
davidmferris, line 34: This is very much a nitpick...but could we maybe check the result of `ButtonVariants[props.variant]` and pass that here (if it exists) instead? That way the component is in charge of adding "approved" classes instead of allowing the app to dictate what gets passed in here.

src/CopyToClipboardButton/CopyToClipboardButton.jsx
```diff
@@ -0,0 +1,75 @@
...
+          type="button"
+        >
+          <Popper
+            dark
+            text="Copied!"
+            visible={copied}
+          >
+            <span>
+              <FontAwesomeIcon
+                className="CopyToClipboardButton__icon"
```
davidmferris, line 46: Don't think this className is being used any longer? Can probably remove if not.

stories/CopyToClipboardButton.stories.jsx
```diff
@@ -0,0 +1,18 @@
...
+
+import CopyToClipboardButton from 'src/CopyToClipboardButton';
+
+export default {
+  title: 'Design System/Copy To Clipboard Button',
+  component: CopyToClipboardButton,
+  decorators: [withKnobs({ escapeHTML: false })],
+};
+
+export const Default = () => (
```
davidmferris, line 12: Could we add a `Neutral` variant here to this file too?

davidmferris (changes requested):
Looking super good! I did notice when you click the button, it gets a gray background. I wonder if this would be fixed by my question to @rsaris about overriding Bootstrap defaults.

scss/buttons.scss
```diff
@@ -7,3 +12,13 @@
     color: $ux-gray-700;
   }
 }
+
+.btn-outline-secondary {
+  color: $ux-blue;
```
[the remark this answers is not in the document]
rroppolo: FWIW I think it works if you put the theme in an scss import above the bootstrap ones (not sure if that is what you were doing @davidmferris) Although I mentioned to Jamison earlier I would hesitate to make that change in this branch if it is going to impact a lot of other components (for example, the btn-primary buttons in the Form stories had the wrong text color for me after adding the theme, didn't really look into why)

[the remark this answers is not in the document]
davidmferris: Yeah I might have just had it out of order. I would be ok to handle that in a follow up issue since it's maybe a heavy lift with unknown repercussions.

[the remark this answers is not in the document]
rsaris: TBH I don't think I realized that we had imported Bootstrap into this repo. With that in mind, if we have bootstrap, should we be using the mixins they provide if we are using it. For example we have:
```
.btn-outline-neutral {
  @include button-variant(transparent, $ux-gray-400);
  @include button-color($ux-gray-900);

  &:focus {
    color: $ux-gray-900;
  }
}
```
With the main repo describing:
```
@mixin button-color($color, $disabled: false) {
  color: $color;

  &:active, &:hover {
    color: $color;
  }

  // Override some very sticky Bootstrap colors
  &:not(:disabled):not(.disabled):active,
  &:not(:disabled):not(.disabled).active {
    color: $color;
  }

  @if ($disabled) {
    &:disabled {
      color: $color;
    }
  }
}
```
Which we could include in this repo.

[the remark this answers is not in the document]
rsaris: That said -- I'm personally OK w/ including our theming into this repo, especially if it makes us use more semantic terms here.

stories/CopyToClipboardButton.stories.jsx
jamisonordway: Would it make more sense for the neutral variant to be the default?

davidmferris: Hmm yeah, probably makes sense that adding no additional classes/styling is the default and `Secondary` could be a variant.

scss/buttons.scss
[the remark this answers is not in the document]
davidmferris: Ok, this is maybe something we should tackle in a future effort because there's probably some things we'll want to think through and iron out. I'm hoping getting a `Button` component in here can be done sooner or later, so maybe it makes sense to handle that at that time.

src/CopyToClipboard/CopyToClipboard.jsx
```diff
@@ -1,40 +1,18 @@
...
-            visible={copied}
-          >
-            <div><FontAwesomeIcon icon={faCopy} /></div>
-          </Popper>
-        </TrackedButton>
-      </ReactCopyToClipboard>
+      <CopyToClipboardButton
+        copyText={props.copyText}
+        trackingEvent={props.trackingEvent}
+        variant={ButtonVariants.NEUTRAL}
```
davidmferris, line 14: Sorry I may have not have expressed well what I meant earlier. `ButtonVariants.NEUTRAL` or `.SECONDARY` is going to return a CSS class. I was thinking instead you'd pass `variant="neutral"` here (probably best to lowercase it so it's consistent with our other string props) and then in the component you could check like `const buttonVariant = ButtonVariants[props.variant];` and add the result of that to the `classNames` prop in that component. That way your prop check is making sure you've passed one of "neutral" or "secondary".

stories/CopyToClipboardButton.stories.jsx
davidmferris: You know what...I thought just adding a `.btn` classname would give you some bootstrap default, but when I check it out in storybook there's no border around the button. So maybe we should actually remove this `Default` for now, until we get our button/theming figured out for the other variants. Sorry about that!

Screen Shot 2020-12-08 at 2 39 35 PM

stories/CopyToClipboardButton.stories.jsx
```diff
@@ -0,0 +1,28 @@
...
+    trackingEvent="copy-text"
+  />
+);
+
+export const Secondary = () => (
+  <CopyToClipboardButton
+    copyText={text('Copy text', 'Copy me!')}
+    displayText={text('Display text', 'Click to copy')}
+    trackingEvent="copy-text"
+    variant={ButtonVariants.SECONDARY}
```
davidmferris, line 26: As noted above, ideally we could pass `variant="secondary"` here so that we're not passing a CSS classname as the prop.

davidmferris (changes requested):
Sorry, just a couple minor things left!

src/CopyToClipboard/CopyToClipboard.jsx
rroppolo: Sorry if I am missing some context here but don't the propTypes already validate this @davidmferris ?
```
  variant: PropTypes.oneOf(Object.keys(ButtonVariants))
```
I feel like we use this pattern other places too

rroppolo: Ahh nm I think I see now, I didn't realize that ButtonVariants values where the css classes

davidmferris: Yeah, I was mainly hoping that we could pass it like `variant="secondary"` as opposed to `variant="btn-outline-secondary"`

jamisonordway: Ah okay I think I get it now, but let me know if I missed something. This means the ButtonVariant keys would be lowercased too, right?
```
export const ButtonVariants = {
  neutral: '',
  secondary: 'btn-outline-secondary',
};
```

davidmferris: Yep exactly! I guess the upper/lower case distinction isn't necessary, but we typically pass lowercase strings so I think it makes sense to keep it that way for consistency.

davidmferris (approved):
Yay looks great! Nice work @jamisonordway 🎉

PR #108, opened 2020-12-09
UIDS-87 Part one: allow FormControlLabel to accept children

rroppolo:
This doesn't completely finish this issue b/c we still have the refactor of `RadioButton` component, but I am not touching that for now b/c it will require a decent amount of changes in the app. But this branch enables `FormControlLabel` to accept children, so we can use that for helper text associated with checkboxes, like the Double screening checkbox:

Bordered checkbox not in a CheckboxButtonGroup
- Grey border
- No blue active state
- More padding (1rem)
- Children are ux-gray-800 and font-type-30 and fill the width

Screen Shot 2020-12-09 at 12 27 20 PM

Bordered checkbox in a CheckboxButtonGroup
- Grey border
- Active state is highlighted blue
- Padding is tighter

Screen Shot 2020-12-09 at 12 27 11 PM

### DIFF
--- a/scss/forms/form_control_label.scss
+++ b/scss/forms/form_control_label.scss
@@ -6,9 +6,29 @@
   align-items: center;
   display: flex;
 
+  &--with-children {
+    align-items: flex-start;
+    flex-direction: column;
+    display: flex;
+  }
+
+  &__label {
+    align-items: center;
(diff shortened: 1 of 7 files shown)


davidmferris (approved):
Looks great!

PR #114, opened 2020-12-11
UIDS-87 remove label element from RadioButton

davidmferris:
Updates the RadioButton to act more like CheckboxButton. I also added an "align" prop to RadioButtonGroup so we can have the option to not grow them to full width.


src/RadioButtonGroup/RadioButtonGroup.jsx
```diff
@@ -4,18 +4,27 @@ import classnames from 'classnames';
...
+};
+
 export const ORIENTATIONS = {
   COLUMN: 'column',
   ROW: 'row',
 };
 
 export default function RadioButtonGroup(props) {
+  const row = props.orientation === ORIENTATIONS.ROW;
+  const centered = props.align === ALIGNMENTS.CENTER;
```
davidmferris, line 19: alignment might be the wrong terminology here since it doesn't preserve the widths of the radio buttons. Maybe it should be a boolean like `props.fullWidth` or something?

jojolee1996:
This looks great! I'm curious about where we align on the naming of these components since there will eventually be the radio button version of the Checkbox Button Group:

Group 296

davidmferris:
@jojolee1996 I added a variant to `RadioButtonGroup` that I think is the equivalent to that?

Screen Shot 2020-12-11 at 2 11 37 PM
You can see it if you check "Bordered Buttons" and row "orientation" in the knobs tab

jojolee1996:
Ohh gotcha, it'll be pretty close to what you have with the bordered row, but it's more similar to what Rachel has for Checkbox Button Group only without the checkbox (see "Computer with a webcam" under Characteristics in new builder design).

Is that something we can tweak with this component? We'd just need to remove the radio button icon and adjust the width of the border.

davidmferris:
Oh cool, yeah I see that now. Would "Computer with a webcam" actually be a radio button since it only allows one choice? @rroppolo have you implemented that yet? I could see `CheckboxButtonGroup` and `RadioButtonGroup` taking a prop to show/hide the icons.

jojolee1996:
Yep, that would basically work as a radio button version to what Rachel created for the Checkbox Group

rroppolo:
I haven't added it yet but that was what I was referring to earlier that Characteristics might need some radio button work. I can do that in its own branch if preferable. Yeah not sure if it is a new variant in addition to `bordered` like maybe `grouped` or something that hides the radio icon, adds the blue background on selection and does not stretch the components like CheckboxButtonGroup does. Also was planning to do the opposite to CheckboxButtonGroup which I believe currently assumes you want the blue background always but Characteristics has some checkbox groups that are vertically aligned instead

src/RadioButton/RadioButton.jsx
```diff
@@ -1,49 +1,35 @@
...
-    </label>
-  );
-});
+const RadioButton = React.forwardRef(({
+  checked,
+  className,
+  disabled,
+  id,
+  name,
+  radioButtonProps,
```
rroppolo, line 10: Do we need this passed as radioButtonProps now or can we just use ...props like CheckboxButton?

davidmferris:
Ok awesome! Yeah that might make sense in a separate branch. I did add the `fullWidth` prop to RadioButtonGroup in this PR though so you'll be able to get the non-stretched version by default (and add the prop if you want it stretched).

rroppolo:
Just a heads up you may be aware but there will be some changes required in-app... like opt-in form and I think a few spots in the old builder.

src/RadioButton/RadioButton.jsx
davidmferris: Nope not needed! It's gone.

src/RadioButton/RadioButton.jsx
```diff
@@ -1,51 +1,34 @@
...
+  <input
+    checked={checked}
+    className={className}
+    disabled={disabled}
+    id={id}
+    name={name}
+    ref={ref}
+    type="radio"
+    value={value}
+    onChange={onChange}
```
rroppolo, line 22: Ahh sorry to be a pain but could we can have  a  `{...rest}` here for any props that are not enumerated? Like `defaultValue` or any aria stuff, I think CheckboxButton does that

davidmferris: Done!

rroppolo (approved):
🎉

PR #115, opened 2020-12-14
CHORE Update to disabled radio styles, tooltip font weight

rroppolo:
Design feedback from https://github.com/user-interviews/rails-server/pull/5218
[ ] Fix checkbox shrinking on mobile
[ ] Give Tooltip messages font-weight 400
[ ] Vertically center tooltip icon
[ ] Fullwidth Radio buttons have extra right margin on mobile when they wrap to vertical column


davidmferris (approved):
LGTM!

PR #117, opened 2020-12-15
UIDS-116 Add compact row styling of RadioButtonGroup

### LINKED ISSUE #116: Create grouped variant of RadioButtonGroup
See Design Principles https://www.notion.so/Design-principles-4cb9ed6450b9404f883dcd3be7bb294b#c194944b98714708a746f96052db03b7

Group_296

rroppolo:
Closes #116

Update row variant of RadioButtonGroup to better match design principles

Group_296

### DIFF
--- a/src/ControlButtonGroup/ControlButtonGroup.jsx
+++ b/src/ControlButtonGroup/ControlButtonGroup.jsx
@@ -1,0 +1,51 @@
+import React, { Children } from 'react';
+import PropTypes from 'prop-types';
+
+export const ORIENTATIONS = {
+  COLUMN: 'column',
+  ROW: 'row',
+};
+
+const ControlButtonGroup = ({
+  children,
+  childChecked,
+  handleChangeValue,
+  orientation,
+  onChange,
+}) => {
+  const row = orientation === ORIENTATIONS.ROW;
+
+  const renderChildElement = (child) => {
+    const { value: childValue } = child.props;
+    const checked = childChecked(childValue);
+
+    // Treat children as controlled components only if group is also controlled
+    const childProps = onChange ? {
+      checked,
+      onChange: handleChangeValue,
+    } : {};
+
+    if (row) {
+      childProps.bordered = true;
+    }
+
+    return React.cloneElement(child, childProps);
+  };
+
+  return Children.toArray(children).map(renderChildElement);
+};
+
+ControlButtonGroup.propTypes = {
+  childChecked: PropTypes.func.isRequired,
+  children: PropTypes.node.isRequired,
+  handleChangeValue: PropTypes.func.isRequired,
+  orientation: PropTypes.oneOf(Object.values(ORIENTATIONS)),
+  onChange: PropTypes.func,
+};
+
+ControlButtonGroup.defaultProps = {
+  orientation: ORIENTATIONS.COLUMN,
+  onChange: undefined,
+};
+
+export default ControlButtonGroup;
--- a/src/RadioButtonGroup/RadioButtonGroup.jsx
+++ b/src/RadioButtonGroup/RadioButtonGroup.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { Children } from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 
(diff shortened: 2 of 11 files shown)


src/RadioButtonGroup/RadioButtonGroup.jsx
```diff
@@ -14,21 +14,47 @@ export default function RadioButtonGroup({
...
+  onChange,
 }) {
   const row = orientation === ORIENTATIONS.ROW;
 
+  const handleChangeValues = (event) => {
+    const eventValue = parseInput(event.target.value);
+    onChange(eventValue);
+  };
+
+  const renderChildElement = (child) => {
```
davidmferris, line 28: These look roughly the same in both components. Would it be possible/worth it to make an HOC that modifies the children and then passes them to the underlying component?

src/RadioButtonGroup/RadioButtonGroup.jsx
```diff
@@ -38,10 +64,20 @@ RadioButtonGroup.propTypes = {
   fullWidth: PropTypes.bool,
   name: PropTypes.string,
   orientation: PropTypes.oneOf(Object.values(ORIENTATIONS)),
+  parseInput: PropTypes.func,
```
davidmferris, line 67: Looking at how we use this prop for `CheckboxButtonGroup`. Wondering if now that we have `WrappedController`'s `setValueAs` if we even need this? Is there a case you were thinking of that we wouldn't be able to use `setValueAs` for?

davidmferris (commented):
This looks really great! I don't know if this is a problem or confusing, but `bordered` is essentially ineffective on row orientation. It's probably fine, but I wonder if we should handle that somehow?

rroppolo:
I had that thought too but I wasn't sure if we want to allow a non-bordered row version? The blue background version looks pretty bad without a border at least, and I didn't want people to have to remember to add the bordered flag on all the individual children to get it to look right... Maybe @jojolee1996 has some thoughts about this? Jojo, do you think we would ever want to support a horizontal version of the radio/checkbox that doesn't have a border around each option? Like this kind of thing?

Screen Shot 2020-12-15 at 2 54 10 PM

Or is it better for us to not enable this option at all and always add a border when showing radio/checkboxes horizontally?

rroppolo:
@jojolee1996 for reference these are the options in this PR (seen in knobs here: https://docs.userinterviews.com/ui-design-system/?path=/story/design-system-form-elements-form-group--with-radio-button-group)

Screen Shot 2020-12-15 at 2 58 11 PM
Screen Shot 2020-12-15 at 2 58 03 PM
Screen Shot 2020-12-15 at 2 57 56 PM
Screen Shot 2020-12-15 at 2 57 47 PM

The issue Dave is raising is that right now you can't set orientation=ROW and *not* have bordered buttons, the fact that it is in row orientation overrides that control

davidmferris:
Yeah, to clarify I was thinking that it should always be bordered when orientation is "row" but I didn't want to confuse other devs in the rare chance they try and set `bordered={false}` and it doesn't do anything. We could probably raise a warning or just have documentation somewhere, unless @jojolee1996 thinks there is a case to have unbordered buttons.

rroppolo:
Ahh I see, yeah let me see if I can add a warning maybe

jojolee1996:
Agree with you both that there should always be a border when the group is set to row 👍

src/ControlButtonGroup/ControlButtonGroup.jsx
```diff
@@ -0,0 +1,51 @@
...
+const ControlButtonGroup = ({
+  children,
+  childChecked,
+  handleChangeValue,
+  orientation,
+  onChange,
+}) => {
+  const row = orientation === ORIENTATIONS.ROW;
+
+  const renderChildElement = (child) => {
```
rroppolo, line 18: I ended up using a component instead of an HOC b/t the two uses need different onChange and different logic to indicate whether something is checked

stories/Form Elements/FormGroup.stories.jsx
```diff
@@ -146,68 +149,89 @@ const ButtonGroupComponent = ({
...
+export const WithCheckboxButtonGroup = () => {
+  const orientation = radios(
+    'Orientation',
+    Object.values(ORIENTATIONS),
+    ORIENTATIONS.COLUMN,
+  );
+  let bordered;
+  if (orientation === ORIENTATIONS.COLUMN) {
+    bordered = boolean('Bordered buttons (only applicable on column orientation)', false);
+  }
```
rroppolo, line 175: What do you think about adding this to the label and hiding this prop if row is checked? I was going to add a warning but I didn't know a way to add it only in non-production env (also eslint doesn't like console warnings). Unless you know a better way to issue that type of warning? Looks like react added a warn method but it is in a newer version..

davidmferris: Yeah I think that's fine! It's probably not worth the trouble it'd require to configure dev-only warnings.

davidmferris: Would it make sense to just set `bordered = true` when orientation is row here? That way we don't have to allow `undefined` as a prop, which is a little confusing for a boolean attribute.

davidmferris (approved):
LGTM aside from one small thing (and if you wouldn't mind setting the storybook deploy branch back to master).

stories/Form Elements/FormGroup.stories.jsx
rroppolo: works for me! updated it to default to true here, then default to false on the component

PR #120, opened 2020-12-15
CHORE allow FormControlLabel to take JSX as text prop

davidmferris:


src/FormControlLabel/FormControlLabel.jsx
```diff
@@ -54,7 +54,7 @@ FormControlLabel.propTypes = {
   Control: PropTypes.oneOfType([PropTypes.func, PropTypes.object]).isRequired,
   disabled: PropTypes.bool,
   id: PropTypes.string.isRequired,
-  text: PropTypes.string.isRequired,
+  text: PropTypes.oneOfType([PropTypes.string, PropTypes.object]).isRequired,
```
rroppolo, line 57: Could this use PropTypes.node to cover both?

davidmferris: Can't see why not!

rroppolo (approved):
🎉

PR #122, opened 2020-12-15
UIDS-121 Add support for html form post to Form component

### LINKED ISSUE #121: Support html form POST in Form component

rroppolo:
Closes #121 

Reproduces the existing behavior of the common/form component, which accepts an action prop and can support an html form POST in addition to an ajax call. The code is mostly taken from the rails-server component except that it can accept a CSRFToken and CSRFParam.

### DIFF
--- a/src/CopyToClipboardButton/CopyToClipboardButton.jsx
+++ b/src/CopyToClipboardButton/CopyToClipboardButton.jsx
@@ -64,7 +64,7 @@ CopyToClipboardButton.propTypes = {
   copyText: PropTypes.string,
   displayText: PropTypes.string,
   trackingEvent: PropTypes.string.isRequired,
-  variant: PropTypes.oneOf(Object.keys(ButtonVariants)),
+  variant: PropTypes.oneOf(Object.values(ButtonVariants)),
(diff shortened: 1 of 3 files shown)


src/CopyToClipboardButton/CopyToClipboardButton.jsx
```diff
@@ -64,7 +64,7 @@ CopyToClipboardButton.propTypes = {
   copyText: PropTypes.string,
   displayText: PropTypes.string,
   trackingEvent: PropTypes.string.isRequired,
-  variant: PropTypes.oneOf(Object.keys(ButtonVariants)),
+  variant: PropTypes.oneOf(Object.values(ButtonVariants)),
```
rroppolo, line 67: I was getting a propTypes error about this from the story. I think this is what we want?  The story is passing ButtonVariants.NEUTRAL which is the value

davidmferris: Ah yeah you're right. Good catch.

davidmferris (approved):
LGTM!

PR #124, opened 2020-12-17
[DRAFT] UIDS-122 - Add yellow color palette and update alerts colors.

fausgaitan:
- Added new yellow color palette to match [figma definition](https://www.figma.com/file/471iS3QrpPtXZpS1xbsJ20/?node-id=1252%3A812)
- Updated classes in `Flash/AlertMessage.jsx` to resolve styles override issues
- Updated `Flash/AlertMessage.scss ` with new color variables. ([figma](https://www.figma.com/file/471iS3QrpPtXZpS1xbsJ20/?node-id=1276%3A14))
- Updated `src/Popper/Popper.scss` to improve tooltip look & feel.


scss/colors/palette.scss
```diff
@@ -57,6 +57,16 @@ $ux-orange-700: #AE6608;
 $ux-orange-800: #7D4A06;
 $ux-orange-900: #653B05;
 
+$ux-yellow-900: #655505;
```
davidmferris, line 60: Nitpick, but could we move these below the red colors to preserve the alphabetization as well as make them ascending 100-900? Also looks like there is another $ux-yellow-100 in this file so there may be a conflict.

src/Popper/Popper.scss
```diff
@@ -10,15 +10,15 @@
   letter-spacing: normal;
   padding: 0.75rem;
   z-index: $z-index-flash;
+  font-weight: $font-weight-light;
 
   &__header {
-    border-bottom: 1px solid $ux-gray-300;
-    margin-bottom: 0.5rem;
+    margin-bottom: 0.3rem;
+    font-weight: $font-weight-bold;
```
davidmferris, line 17: Could this use `@include font-type-30--bold;`?

davidmferris (changes requested):
Looks great! Just a couple of small comments.

scss/colors/palette.scss
fausgaitan: done! 71bdc1bd91254377f5ebe09384f25b7efffdad2c

src/Popper/Popper.scss
fausgaitan: done! 0578ea0fa5a9cfd2d5ffcb2887a4a6644707381d

davidmferris (approved):
Looks great!

PR #131, opened 2020-12-28
[ Draft ] Fix Input fields violations

fausgaitan:
You can check all the input fields violations in this **[Notion Doc](https://www.notion.so/userinterviews1/DS-Violations-b4c3c12921ea49d7a4aa775cd609b83c)** (feel free to comment there too)

So far in this PR we are fixing:

- Error helper text contrast issue
- Ensure aria-hidden element does not contain focusable elements ([fix based on this article](https://dequeuniversity.com/rules/axe/3.5/aria-hidden-focus?application=axeAPI))
- Added aria-label to InputComponent to fix label issue in default and required variants.

### DIFF
--- a/stories/Form Elements/FormGroup.stories.jsx
+++ b/stories/Form Elements/FormGroup.stories.jsx
@@ -27,7 +27,7 @@ const InputComponent = (props) => {
   const handleChangeValue = (event) => setValue(event.target.value);
 
   return (
-    <Input value={value} onChange={handleChangeValue} {...props} />
+    <Input aria-label="input text label goes here" value={value} onChange={handleChangeValue} {...props} />
   );
 };
 
@@ -41,6 +41,7 @@ export const Required = () => (
   <FormGroup
     id="with-required"
     label="Form Group with label"
+    labelHtmlFor="input"
     required
   >
     <InputComponent id="with-required-input" name="required" placeholder="Text is required" />
--- a/stories/Form Elements/FormGroup.stories.jsx
+++ b/stories/Form Elements/FormGroup.stories.jsx
@@ -27,13 +27,13 @@ const InputComponent = (props) => {
   const handleChangeValue = (event) => setValue(event.target.value);
 
   return (
-    <Input aria-label="input text label goes here" value={value} onChange={handleChangeValue} {...props} />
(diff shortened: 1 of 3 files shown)


jojolee1996 (approved):
Looks good!

stories/Form Elements/FormGroup.stories.jsx
```diff
@@ -27,7 +27,7 @@ const InputComponent = (props) => {
   const handleChangeValue = (event) => setValue(event.target.value);
 
   return (
-    <Input value={value} onChange={handleChangeValue} {...props} />
+    <Input aria-label="input text label goes here" value={value} onChange={handleChangeValue} {...props} />
```
rroppolo, line 30: It looks like there is only one story that does not have a label specified by the story itself. Wondering if instead of putting this here (which is called by every story, so will override the `labelHtmlFor` set on `FormGroup` in this other stories, could we move this instead to inside the Default story, like:
```
export const Default = () => (
  <FormGroup id="default">
    <InputComponent aria-label="input text label goes here" id="default-input" name="default" placeholder="default" />
  </FormGroup>
);
```

Then the other failing story is the `Required` story, but in this case it already has a `label`. It is just missing the `labelHtmlFor` prop which should match the `id` of the input inside that it is labelling
```
export const Required = () => (
  <FormGroup
    id="with-required"
    label="Form Group with label"
    labelHtmlFor="with-required-input"
    required
  >
    <InputComponent id="with-required-input" name="required" placeholder="Text is required" />
  </FormGroup>
);
```

rroppolo (commented):
Just a small comment, looks great!

stories/Form Elements/FormGroup.stories.jsx
fausgaitan: Done! thanks for the clear explanation, @rroppolo ! 1feea39a43e48894fc8d3c3dbe5323ba1c54e722

rroppolo (approved):
Look great, thanks @fausgaitan!

stories/Form Elements/FormGroup.stories.jsx
fausgaitan: test @fausgaitan

fausgaitan:
test @fausgaitan

PR #132, opened 2020-12-28
Feature/UI ds 77 card component

### LINKED ISSUE #77: Move Card common component to design system

arlemagne:
Closes #77 
Figma link: https://www.figma.com/file/471iS3QrpPtXZpS1xbsJ20/Components?node-id=1252%3A792

Adds Card component with variants no padding, active, and ruled.

Examples:
Screen Shot 2020-12-28 at 1 19 27 PM

Screen Shot 2020-12-28 at 1 19 15 PM

### DIFF
--- a/.github/workflows/gh-pages-deploy.yml
+++ b/.github/workflows/gh-pages-deploy.yml
@@ -2,7 +2,7 @@
   on:
     push:
       branches:
-        - master
+        - feature/UIDS-77-card-component
   jobs:
     gh-pages-deploy:
       name: Deploying to gh-pages
--- a/.github/workflows/gh-pages-deploy.yml
+++ b/.github/workflows/gh-pages-deploy.yml
@@ -2,7 +2,7 @@
   on:
     push:
       branches:
-        - feature/UIDS-77-card-component
+        - master
   jobs:
     gh-pages-deploy:
       name: Deploying to gh-pages
--- a/src/Card/Card.jsx
+++ b/src/Card/Card.jsx
@@ -12,7 +12,6 @@ export const CardSizes = {
 };
 
 const Card = ({
-  active,
   children,
   className,
   divided,
@@ -46,7 +45,6 @@ const Card = ({
         `Card--${size}`,
         className,
         {
-          'Card--active': active,
           'Card--divided': divided,
           'Card--no-padding': noPadding,
         },
@@ -57,7 +55,6 @@ const Card = ({
 };
 
 Card.propTypes = {
-  active: PropTypes.bool,
   className: PropTypes.string,
   divided: PropTypes.bool,
   elementType: PropTypes.oneOfType([PropTypes.func, PropTypes.string]),
@@ -69,7 +66,6 @@ Card.propTypes = {
 };
 
 Card.defaultProps = {
-  active: undefined,
   className: undefined,
   divided: false,
   elementType: 'section',
--- a/src/Card/Card.jsx
+++ b/src/Card/Card.jsx
@@ -25,7 +25,7 @@ const Card = ({
   const cardChildren = (
     <Fragment>
       { title && <h2 className="Card__title">{title}</h2> }
-      { subTitle && <h5 className="Card__subtitle">{subTitle}</h5> }
+      { subTitle && <h3 className="Card__subtitle">{subTitle}</h3> }
       { children }
     </Fragment>
   );
--- a/src/Card/Card.jsx
+++ b/src/Card/Card.jsx
@@ -15,10 +15,10 @@ const Card = ({
   active,
   children,
   className,
+  divided,
   elementType,
   helperText,
   noPadding,
-  ruled,
   size,
   subTitle,
   title,
@@ -28,10 +28,10 @@ const Card = ({
     <Fragment>
       <div className="Card__header">
         { title && <h2 className="Card__title">{title}</h2> }
-        { helperText && <span className="Card__helper-text">&nbsp;{helperText}</span>}
+        { helperText && <span className="Card__helper-text">{helperText}</span>}
       </div>
 
-      { ruled && <hr className="Card__rule" /> }
+      { divided && <hr className="Card__divider" /> }
       { subTitle && <h3 className="Card__subtitle">{subTitle}</h3> }
       { children }
     </Fragment>
@@ -47,8 +47,8 @@ const Card = ({
         className,
         {
           'Card--active': active,
+          'Card--divided': divided,
           'Card--no-padding': noPadding,
-          'Card--ruled': ruled,
         },
       ),
     },
@@ -59,10 +59,10 @@ const Card = ({
 Card.propTypes = {
   active: PropTypes.bool,
   className: PropTypes.string,
+  divided: PropTypes.bool,
   elementType: PropTypes.oneOfType([PropTypes.func, PropTypes.string]),
   helperText: PropTypes.string,
   noPadding: PropTypes.bool,
-  ruled: PropTypes.bool,
   size: PropTypes.string,
   subTitle: PropTypes.string,
   title: PropTypes.string,
@@ -71,10 +71,10 @@ Card.propTypes = {
 Card.defaultProps = {
   active: undefined,
   className: undefined,
+  divided: false,
   elementType: 'section',
   helperText: undefined,
   noPadding: false,
-  ruled: false,
   size: CardSizes.LARGE,
   subTitle: undefined,
   title: undefined,
--- a/src/Card/Card.jsx
+++ b/src/Card/Card.jsx
@@ -31,8 +31,8 @@ const Card = ({
         { helperText && <span className="Card__helper-text">&nbsp;{helperText}</span>}
       </div>
 
-      { subTitle && <h3 className="Card__subtitle">{subTitle}</h3> }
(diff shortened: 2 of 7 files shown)


arlemagne:
@rroppolo leaving this as a draft since I'm not sure how holistic this initial Card addition should be - I think we have some other styles of Cards in the app that may not be considered default. I just implemented the styles in this [figma link ](https://www.figma.com/file/471iS3QrpPtXZpS1xbsJ20/Components?node-id=1252%3A792) as a starting point, but let me know what you think if you get a chance!

arlemagne:
@jojolee1996 @fausgaitan should the horizontal rule go above or below the subtitle text?

Screen Shot 2020-12-28 at 1 40 28 PM

jojolee1996:
@arlemagne The divider should go right under the card title. Do we currently have any cards that uses subtitle text? The closest thing I can think of is something like Technical requirements in the new builder ([link](https://www.figma.com/file/ArOFSzsGij40oMMjF8S7QG/Project-builder-discovery-JoJo-Jess-Dave-Rachel?node-id=283%3A2088))

arlemagne:
@jojolee1996 I think the projects browse page (/researcher) has what would be considered subtitle text:

Screen Shot 2020-12-28 at 3 45 18 PM

either way I'll make that change and move the divider under the subtitle 👍

src/Card/Card.jsx
```diff
@@ -0,0 +1,83 @@
...
+};
+
+const Card = ({
+  active,
+  children,
+  className,
+  elementType,
+  helperText,
+  noPadding,
+  ruled,
```
arlemagne, line 21: alternately i guess this could be "divided", I don't feel too strongly one way or the other

src/Card/Card.scss
```diff
@@ -0,0 +1,82 @@
...
+  &--active {
+    outline: $ux-green-500 auto 5px;
+  }
+
+  &--no-padding {
+    padding: 0;
+  }
+
+  &--ruled {
+    .Card__title {
```
arlemagne, line 60: this can't be right... basically I need to make sure the title doesn't have the default margin so that the hr is even on top and bottom

.github/workflows/gh-pages-deploy.yml
```diff
@@ -2,7 +2,7 @@
   on:
     push:
       branches:
-        - master
+        - feature/UIDS-77-card-component
```
rroppolo, line 5: Just a reminder to change this back before merging (sorry it's not an ideal system)

src/Card/Card.jsx
```diff
@@ -0,0 +1,83 @@
...
+  return createElement(
+    elementType,
+    {
+      ...props,
+      className: classNames(
+        'Card',
+        `Card--${size}`,
+        className,
+        {
+          'Card--active': active,
```
rroppolo, line 49: Ah actually you can possibly remove `active` here and above as it will probably go in the `Tile` component.

src/Card/Card.jsx
rroppolo: Yeah, same. @jojolee1996 do you have a naming convention for this in Figma by chance? Talking about a Card with an hr under the title (currently it is not required)

Screen Shot 2020-12-28 at 5 13 12 PM

src/Card/Card.scss
```diff
@@ -0,0 +1,82 @@
...
+  padding: $card-spacing;
+
+  width: calc(100% - #{$card-xs-spacing});
+
+  @include media-breakpoint-up(sm) {
+    width: calc(100% - 2rem);
+  }
+
+  &__header {
+    display: flex;
```
rroppolo, line 29: May want flex-wrap: wrap here to wrap helper text under the title on xs sizes. (Also maybe we take the margin-bottom of the title and put it here instead?

Screen Shot 2020-12-28 at 5 21 46 PM


Screen Shot 2020-12-28 at 5 23 17 PM

src/Card/Card.scss
```diff
@@ -0,0 +1,82 @@
...
+  &__subtitle {
+    @include font-type-30;
+    color: $ux-gray-700;
+  }
+
+  &__title {
+    @include font-type-50--bold;
+  }
+
+  &--active {
```
rroppolo, line 51: Can probably remove this for now

stories/Card.stories.jsx
```diff
@@ -0,0 +1,28 @@
...
+
+export default {
+  title: 'Design System/Card',
+  component: Card,
+  decorators: [withKnobs({ escapeHTML: false })],
+};
+
+export const Default = () => (
+  <Card
+    active={boolean('Is active', false)}
```
rroppolo, line 16: Can probably remove this

src/Card/Card.scss
```diff
@@ -0,0 +1,66 @@
...
+@import '~bootstrap/scss/variables';
+@import '~bootstrap/scss/mixins/breakpoints';
+@import '../../scss/theme';
+
+$card-xs-spacing: 1rem;
+$card-sm-spacing: 1.25rem;
+$card-width: 56rem;
+$card-width-md: 40rem;
+$card-width-sm: 32rem;
+$card-width-xs: 15.625rem;
```
rroppolo, line 12: @jojolee1996 did you have a different size you wanted here that was not a decimal in rems?

src/Card/Card.scss
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  &__helper-text {
+    @include font-type-50--light;
+    color: $ux-gray-900;
+  }
+
+  &__rule {
+    border: 1px solid $ux-gray-400;
+    margin: 0.75rem auto;
```
rroppolo, line 39: @jojolee1996 I think we had these margins a little tighter in the new project builder, would we want this change to be reflected there too?
https://github.com/user-interviews/rails-server/pull/5022#issuecomment-736143446
Right now it is at
    margin-top: 0.25rem;
    margin-bottom: 0.5rem;

Screen Shot 2020-12-28 at 5 38 36 PM

src/Card/Card.scss
rroppolo: Hmm I wonder if we should just remove this margin on `&__title` always?  I noticed `card.scss` in the app has:
```

  .card-title {
    ...
    margin: 0;
  }
```

Not sure why I didn't bring that over here but looking at some of our existing Cards, they actually technically don't have bottom margins.
Screen Shot 2020-12-28 at 5 42 53 PM

@jojolee1996 what do you think, should the Card title have a margin-bottom if there is no horizontal rule?

src/Card/Card.jsx
```diff
@@ -0,0 +1,83 @@
...
+  size,
+  subTitle,
+  title,
+  ...props
+}) => {
+  const cardChildren = (
+    <Fragment>
+      <div className="Card__header">
+        { title && <h2 className="Card__title">{title}</h2> }
+        { helperText && <span className="Card__helper-text">&nbsp;{helperText}</span>}
```
rroppolo, line 31: Ooh is it possilbe to do this with margin vs. &nbsp?

src/Card/Card.jsx
jojolee1996: I have it as Divider in Figma, might be more explicit than ruled? Not sure what most devs are used to calling it, but I'm okay with either one.
Screen Shot 2020-12-29 at 9 13 46 AM

src/Card/Card.scss
jojolee1996: Are we able to give card titles a margin-bottom of 0.25rem if there's no horizontal rule to give it some breathing room?

src/Card/Card.scss
jojolee1996: Yep, can we make `$card-width-xs` a clean 15rem?

src/Card/Card.scss
jojolee1996: I actually prefer it to be tighter, so would we be able to update the margin to match what's in the new builder @arlemagne?

jojolee1996:
@arlemagne this looks great!! Just have a few tweaks for you:
- Can we change the horizontal rule's `border` to `border-top: 1px solid rgba(0,0,0,.1)`? (Pulled this from new builder work since the line is currently looking a bit thick)
- Update the subtitle to be `$ux-gray-800` and with `font-weight: 300 `
- Add a `border-radius` of 4px to each card

arlemagne:
totes! I'll make those changes @jojolee1996

src/Card/Card.scss
```diff
@@ -0,0 +1,88 @@
...
+      max-width: $card-width-sm;
+    }
+  }
+
+  &--xs {
+    @include media-breakpoint-up(sm) {
+      max-width: $card-width-xs;
+    }
+
+    .Card__helper-text {
```
arlemagne, line 83: is it better practice to add a `&--xs` to `&__helper-text` above?

rroppolo: I usually do it this way but might just be my personal preference

rroppolo (approved):
This looks great! Congrats on a first design system PR 🙌

arlemagne:
@jojolee1996 made those updates btw, let me know what you think but no rush

jojolee1996 (approved):
Looks great, thanks Arleigh!

PR #134, opened 2020-12-29
UIDS-97 Remove inline form components

jojolee1996:


rroppolo (approved):
LGTM! Sounds like we can also probably remove inline from the Form components but could be a separate issue for that.

jojolee1996:
@rroppolo I'm probably approaching this the wrong way but it's not as simple as removing the `inline` classes inside Form.jsx is it?

rroppolo:
Yup you've got it right, it would be removing the inline class getting applied from `Form.jsx` here

className={classNames('Form', className, ~{ 'Form--inline': inline }~)}

And then the entire `&--inline { ...} ` block inside `Form.scss`

jojolee1996:
> Yup you've got it right, it would be removing the inline class getting applied from `Form.jsx` here
> 
> className={classNames('Form', className, ~{ 'Form--inline': inline }~)}
> 
> And then the entire `&--inline { ...} ` block inside `Form.scss`

Thanks @rroppolo!! I think that should do it?

src/Form/Form.jsx
```diff
@@ -23,7 +23,7 @@ const Form = forwardRef(({
   return (
     <form
       action={action}
-      className={classNames('Form', className, { 'Form--inline': inline })}
+      className={classNames('Form', className,)}
```
rroppolo, line 26: Thanks for doing that @jojolee1996! Looks great, just two super small things. Could you remove the last trailing comma here? The one after className
```
className,)}
```

Could you also remove the `inline,` above in the parameters? It's on line 14 - sorry I forgot about that one :)
You can also run `yarn lint` on your command line and it will point out both of those things, although please ignore there is an error coming from Tooltip.jsx that is a known issue when you run that. But you can see it pointed out the other two in Form.jsx which is quite handy:

Screen Shot 2020-12-31 at 5 36 04 PM

jojolee1996: Thanks Rachel! I'm still getting the error about `inline` being defined but never used when I run `yarn lint`. Is it okay to keep them as is?

davidmferris: Sorry to interject, but I think that's because you're still defining an `inline` propType. If you remove that I think the warning should go away.

jojolee1996: Sweet, thanks for clarifying!

PR #135, opened 2020-12-30
[ Draft ] Update pills styling and add new variant

fausgaitan:
We need to update the pills so they all have borders. 
The contrast of the current pills when used in gray backgrounds is very poor.

[Figma Variants](https://www.figma.com/file/471iS3QrpPtXZpS1xbsJ20/?node-id=1259%3A484)

### DIFF
--- a/src/Pill/Pill.scss
+++ b/src/Pill/Pill.scss
@@ -4,35 +4,44 @@
 .Pill {
   display: inline;
   font-weight: $font-weight-normal;
-
-  //////// COLORS!
-  // Default color is gray and sadddddd
   background-color: $ux-gray-200;
   color: $ux-gray-800;
+  border: 1px solid $ux-gray-400;
 
   &--blue {
     background-color: $ux-blue-100;
     color: $ux-blue-700;
+    border: 1px solid $ux-blue-300;
   }
 
   &--orange {
     background-color: $ux-orange-100;
     color: $ux-orange-900;
+    border: 1px solid $ux-orange-400;
   }
 
   &--green {
     background-color: $ux-green-200;
     color: $ux-green-800;
+    border: 1px solid $ux-green-400;
   }
 
   &--yellow {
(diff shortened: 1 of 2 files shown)


src/Pill/Pill.scss
```diff
@@ -4,42 +4,51 @@
 .Pill {
   display: inline;
   font-weight: $font-weight-normal;
-
-  //////// COLORS!
```
AlexAlt, line 8: it seems like this line of the comment is in here to keep the CSS organized, maybe you should keep it. The one below it is probably ok to remove 😅

AlexAlt (approved):
one small comment, otherwise looks good!

jojolee1996 (approved):
These are looking great!

wenzeng4 (approved):
Looks good to me!

stories/Pill.stories.jsx
```diff
@@ -14,7 +14,7 @@ export default {
   decorators: [withKnobs],
 };
 
-const colors = ['', 'blue', 'orange', 'green', 'gray', 'yellow'];
+const colors = ['', 'blue', 'orange', 'yellow', 'green', 'gray', 'silver' ];
```
rroppolo, line 17: Can you remove the space before the last ']'? You can also run `yarn lint` on the command line and it will show you any formatting things that the linter doesn't like. There are some linter warnings from `Tooltip.jsx` that were there already though so you don't have to worry about those, (hopefully will get rid of those soon).

rroppolo (approved):
One super quick formatting thing but LGTM! 👍

stories/Pill.stories.jsx
fausgaitan: thanks for that yarn lint tip, I didn't update anything there but it's good to know I can check for errors using that. 👍 

here's the commit to remove that space:
6b90366813cfc622abe42acfd9fc84c9f5eb70b5

PR #142, opened 2021-01-13
Fix input focus box shadow picking up bootstrap styles

rroppolo:
See review app here:
https://user-interviews-stagin-pr-5476.herokuapp.com/projects/builder/1/project-overview
staging@userinterviews.com / Userinterviews1

Overrides bootstrap variables that were clashing with the design system in app for the blue input outline:

Before & after:
Screen Shot 2021-01-12 at 6 52 17 PM
Screen Shot 2021-01-12 at 6 51 23 PM

Also on focus input text color was changing from black to grey, and textarea color was not matching input color:
Before & After:
Screen Shot 2021-01-12 at 6 53 32 PM
Screen Shot 2021-01-12 at 6 54 45 PM


src/Tooltip/Tooltip.jsx
```diff
@@ -25,30 +25,14 @@ function removeClickOutsideListener(listener) {
...
-  };
-
-  static defaultProps = {
-    icon: faQuestionCircle,
-    iconClasses: undefined,
-    header: undefined,
-    strategy: undefined,
-    theme: 'dark',
-    onShow: undefined,
-  };
```
rroppolo, line 47: Just fixing an existing eslint error here to be consistent in the way we define propTypes

scss/colors/inputs.scss
```diff
@@ -4,3 +4,6 @@ $input-border-color: $ux-gray-400; // Override bootstrap default
 $input-disabled-bg: $ux-gray-300; // Override bootstrap default
 $input-focus-border-color: $ux-blue-300;
 $input-btn-focus-color: $ux-blue-300;
+
+$input-color: $ux-black;
+$input-focus-color: $ux-black;
```
rroppolo, line 9: @jojolee1996 is this the right color for inputs? I noticed we were already using `ux-black` below, but then your issue from the builder follow-on was mentioning ux-gray-900

jojolee1996: Design team has talked about moving away from `ux-black` and replacing it with `ux-gray-900` as it's slightly less harsh on lighter backgrounds. Can we also double-check that it'll be using the new value of `#101010`?

rsaris (approved):

rroppolo:
Hey @jojolee1996 ! Do you think this one is ok to merge? I updated the input colors to be ux-gray-900 as you described, which was already changed to be the new #101010.

The two changes would be that text color and fixing the weird thick outline that was previously appearing in the new builder
https://user-interviews-stagin-pr-5476.herokuapp.com/projects/builder/1/project-overview
staging@userinterviews.com / Userinterviews1

src/Input/Input.scss
```diff
@@ -10,12 +10,13 @@
 
   &:focus {
     border: 1px solid #92CBFC;
-    box-shadow: inset 0px 1px 1px rgba(0, 0, 0, 0.075);
+    box-shadow: $input-focus-box-shadow;
+    color: $input-focus-color;
     filter: drop-shadow(0px 0px 8px rgba(102, 175, 233, 0.7));
```
fausgaitan, line 15: Hey @rroppolo I think we need to update this from 8px to 3px... 

Looking at the focus state in context, the shadow is overlapping with the label a little bit and it's better if we decrease this value...

so we go from this
image

to this:
image

Super tiny! the rest looks good to me!! thanks so much!

fausgaitan (approved):
LGTM! just a minor comment.

src/Input/Input.scss
rroppolo: Thanks @fausgaitan! Updated it to 3px
Screen Shot 2021-01-21 at 3 05 28 PM

jojolee1996 (approved):
Changes look good! Thanks @rroppolo

PR #160, opened 2021-02-02
CHORE Upgrade design system package version

rroppolo:
Going to make a DS release. Also I messed up & merged the action even though you warned me about it 🤦


rsaris (approved):
Looks great. I'm not sure how annoying it is but we may want to look into updating this repo to use main so we're consistent.

PR #164, opened 2021-02-03
UIDS-5532 Added ProfileCell component and story (#5532)

SirSnacksalot:
Added the ProfileCell component to Storybook!

I moved the existing component and styling from "rails-server" and modified it to be a functional component with JoJo's design updates. I renamed the "ProfileImage" component to "Avatar" as this is what people seem to be referring to it as now. To keep the "colorId" prop modifying the color of the Avatar background, I also added an existing scss file from "rails-server". I noticed Wen was modifying those colors in the Figma, so didn't change anything in that file yet.

I added three stories for the ProfileCell component: Small, Large, With Image (when the user has a profile image).


src/ProfileCell/ProfileCell.jsx
```diff
@@ -0,0 +1,73 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import Avatar from './Avatar';
+
+import './ProfileCell.scss';
+
+function ProfileCell(props) {
+  function getProfileImage() {
+    return props.user.imageUrl || props.user.profilePictureUrl;
```
rroppolo, line 10: Same thought here re: making this a const now that we are using a functional component

stories/ProfileCell.stories.jsx
```diff
@@ -0,0 +1,68 @@
...
+  decorators: [withKnobs],
+};
+
+export const Small = () => (
+  <ProfileCell
+    colorId={number('Color ID', null)}
+    showAlert={boolean('Show Alert', false)}
+    subtitle={text('Subtitle Text', `riley@userinterviews.com`)}
+    user={userNoImage}
+    visible
```
rroppolo, line 20: I think visible might not be needed?

stories/ProfileCell.stories.jsx
```diff
@@ -0,0 +1,68 @@
...
+    colorId={number('Color ID', null)}
+    large
+    showAlert={boolean('Show Alert', false)}
+    subtitle={largeSubtitle}
+    user={userWithImage}
+    visible
+  />
+);
+
+const largeSubtitle = (
```
rroppolo, line 46: It might be more consistent with other files to move these `const`s higher up in the file above where they are used in the stories.

stories/ProfileCell.stories.jsx
```diff
@@ -0,0 +1,68 @@
...
+    showAlert={boolean('Show Alert', false)}
+    subtitle={largeSubtitle}
+    user={userNoImage}
+    visible
+  />
+);
+
+export const WithImage = () => (
+  <ProfileCell
+    colorId={number('Color ID', null)}
```
rroppolo, line 37: Maybe we remove the colorId={number('Color ID', null)} knob just from this story since I think the image always overrides it anyway?

rroppolo: Also I enjoyed the image in this story very much :D

rroppolo (commented):
This looks fantastic 👍  Had a some very minimal thoughts/questions - let me know what you think! But overall looks great, you picked up on the conventions of the repo really well. Will add Jojo to check out the styles but I think we are in great shape to start making use of this in the app probably tomorrow if that works!

rroppolo:
The only other thing I thought of is that it looks like currently in the app we override some of this css in various places to add ellipsis truncation at various max-widths, like:
```
  // ProfileCell Overrides
  .profile-cell {
    &__content__name,
    &__content__subtitle {
      @include ui-truncate(15rem);
    }
  }
```

It looks like we use different max-widths right now depending on where we use the ProfileCell (Messages has 15rem, Participants table in the workspace uses 18rem). @jojolee1996 what are your thoughts about this? Do you think it makes sense to maintain this flexibility about allowing different max-widths for truncation in different places? Or could we be ok with just using some default like 15rem, or even just a couple of sizes?

src/ProfileCell/Avatar.jsx
```diff
@@ -0,0 +1,95 @@
...
+import './Avatar.scss';
+
+function Avatar(props) {
+  const [imageLoadFailed, setImageLoadFailed] = useState(false);
+
+  function onImageLoadError() {
+    setImageLoadFailed(true);
+  }
+
+  function displayImage() {
```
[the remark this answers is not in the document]
SirSnacksalot: Cool I agree! Should I go ahead and change and commit?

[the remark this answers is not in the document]
rroppolo: Yup sounds good, thank you!

src/ProfileCell/Avatar.jsx
```diff
@@ -0,0 +1,95 @@
...
+  image: undefined,
+  large: false,
+  name: '',
+  showAlert: false,
+  url: undefined,
+};
+
+const UI_MOD_OPERAND = 6;
+const uiModClassName = (id) =>
+  Number.isNaN(id) ? null : `ui-mod ui-mod--${window.parseInt(id) % UI_MOD_OPERAND}`;
```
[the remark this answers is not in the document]
SirSnacksalot: Yeah I felt like I was missing some context on where this was used in the main app and what it generally means haha. Given that, `Styles/classNames.js` makes sense to me to start off with!

stories/ProfileCell.stories.jsx
SirSnacksalot: Oops yeah I'll get rid of that 😅 good catch!

jojolee1996:
@SirSnacksalot these look amazing!! Wen and I just solidified the calendar palette so feel free to update the avatar colors when you have a chance: https://www.figma.com/file/471iS3QrpPtXZpS1xbsJ20/Components?node-id=3451%3A6594

To answer @rroppolo's question around max-widths for truncation, I think it would make sense to keep the flexibility we have now. Is there a defined list of these different widths that I can look at?

stories/ProfileCell.stories.jsx
jojolee1996: +1, always appreciate a good doge pic

rroppolo:
Sounds good thanks @jojolee1996 ! These are the only two instances I see right now but there might be more once we start using this DS component in more places:

Screen Shot 2021-02-03 at 6 49 07 PM
Screen Shot 2021-02-03 at 6 48 27 PM

@SirSnacksalot what do you think about possibly accepting a prop for maxWidth to this component? I'd love to remove these one-off css overrides that are sprinkled in the app right now.

SirSnacksalot:
@rroppolo Would it make sense to make max-width a prop? So it can be something configured in the component rather than an override in a css file?

jojolee1996 (approved):
LGTM 👍

rroppolo (approved):
Looks great!

PR #175, opened 2021-02-18
UIDS-163 updates Alert component, changes Flash into Toast component

jasonbasuil:
- made style updates to the AlertMessage styling for the Flash component so it floats right and limited it's width.
- Haven't considered the new icon or CTA button yet based on the proposed component update on Figma. Should those be added?
- Also added scss variables for spacing and elevations since I didn't find them anywhere.

Screen Shot 2021-02-18 at 2 48 17 PM


src/Flash/AlertMessage.scss
```diff
@@ -11,7 +11,9 @@
   margin-bottom: 1rem;
   padding: .75rem 1.25rem;
   position: relative;
-  border: 1px solid transparent;
+  max-width: 22.5rem;
```
rroppolo, line 14: This might be a question more for @fausgaitan, but wondering whether we want to put this max-width on all AlertMessage components, or only those that appear as Flash messages. Sorry @jasonbasuil this is some context that I probably should have given you earlier, but there are two components in play here. One is the `Flash` component, which as the name implies gets flashed on the screen and is dismissible, and is fixed positioned at the top of the page. Like:
Screen Shot 2021-02-18 at 8 19 19 PM

Then there's the `AlertMessage` component, which is used inside the Flash but also on its own. When it is used in isolation, unlike the Flash it is inline in the content of the page, something more like this one:

Screen Shot 2021-02-18 at 8 21 42 PM

(Part of the confusion I think is that `AlertMessage` should probably be moved a level up and outside of the `Flash/` directory. Just a historical thing, the Flash came first and then we started introducing the idea of a permanent alert - but that's its own thing :D )

But anyways sorry my question for @fausgaitan is, do we only want the max-width on the alerts inside a Flash and keep the inline alerts (like that second example) flowing to their container widths?

@jasonbasuil depending on the answer there, we may want to put the max-width on the `Flash` component instead. Also, since that component is position: fixed we could probably use positioning to right-align it vs. float

rroppolo (commented):
Had a few thoughts that we can touch base on tomorrow, but this is looking awesome! I like the new styles a lot, this is going to be a really nice improvement 🙌

scss/spacing.scss
```diff
@@ -0,0 +1,7 @@
+$ux-spacing-01: 0.25rem;
```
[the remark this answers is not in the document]
fausgaitan: Hey @rroppolo the spacing section was defined a while ago, maybe we didn't share this with you. Feel free to [check it out](https://www.figma.com/file/471iS3QrpPtXZpS1xbsJ20/?node-id=3015%3A4280). Your point about the naming convention is great, maybe we can update that so it's a double-digit increment.

src/Flash/AlertMessage.scss
fausgaitan: @rroppolo exactly we should keep the flash component (we call them [Toast alerts](https://www.figma.com/file/471iS3QrpPtXZpS1xbsJ20/Components?node-id=3286%3A337) now in figma) with a max-width of 300px and alertMessage (which we now call [inline alert](https://www.figma.com/file/471iS3QrpPtXZpS1xbsJ20/Components?node-id=3286%3A337)) should expand 100% within its container)

I'll be adding the anatomy and dimensions ASAP.

Can we please rename the components in Storybook so we mirror what is defined in Figma?

fausgaitan:
I just got goosebumps by looking at this PR!! So exciting!!!

src/Flash/AlertMessage.scss
fausgaitan, line 14: I wonder if we should define just `width: 300px`? Not sure what would be the correct CSS rule but the idea is that all toasts alerts should have [the same width](https://www.figma.com/file/471iS3QrpPtXZpS1xbsJ20/Components?node-id=3286%3A5).

fausgaitan: Maybe you can work on this in a different PR or here... not sure how complex it would be to include that, but can we please make the toasts alerts fade away and make them disappear after 5 seconds?

src/Flash/AlertMessage.scss
```diff
@@ -23,39 +25,47 @@
 .AlertMessage-success {
   background-color: $ux-green-200;
```
fausgaitan, line 26: You can add the icons by the way! I'm planning to update the megaphone for announcements because it's a bit complex to use in that size... but the rest feel free to include them.

scss/spacing.scss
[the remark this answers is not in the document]
jasonbasuil: @rroppolo I agree that the naming convention can probably be discussed in our gardening meeting. It may be worthwhile to follow the same double-digit increments just for consistency and the need for flexibility between sizes. I can create a new issue for this for us to look revisit later.

src/Flash/AlertMessage.scss
jasonbasuil: Thanks for the clarification @rroppolo @fausgaitan ! I should be able to make those updates on this PR.

jasonbasuil: Wasn't sure if this was a defined width, but I will make the change. thanks!

src/Flash/AlertMessage.scss
jasonbasuil: Neat! I'll look into it thanks!

scss/spacing.scss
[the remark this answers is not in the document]
rroppolo: Perfect - thank you @jasonbasuil !

src/Flash/AlertMessage.scss
rroppolo: Heads up @jasonbasuil that the Flash component does have an `autoDismiss` property that you can probably use (not usre if the time is 5sec or not) but right now it is opt-in only. So sounds like as part of this we want to make it the default behavior!

src/Flash/AlertMessage.scss
jasonbasuil: @fausgaitan general question about icon positioning that can be answered next week. Icons are currently aligned with Alert / Toast titles. Titles are currently optional for the component. Should icons be used if there isn't a title and just align it with the message? Or omit the icon if only a message? Or make the title required for better readability at a glance? Would love your thoughts.

fausgaitan: Hey @jasonbasuil, this is a great question! My initial thought is that we should always display the icon, when there's a title only, or message only too. But let me play around a bit in figma and let you know if this is the correct approach. Thanks for bringing this up!

jasonbasuil: @fausgaitan I was thinking the same thing so I went ahead a did a first pass:

Screen Shot 2021-02-23 at 9 04 50 AM
Screen Shot 2021-02-23 at 9 05 28 AM
Screen Shot 2021-02-23 at 9 05 48 AM

Also, here are the icons as they stand now for Toast / AlertMessage:

Screen Shot 2021-02-23 at 9 08 45 AM

src/Alert/AlertMessage.jsx
```diff
@@ -0,0 +1,109 @@
...
+  if (!Object.values(MessageTypes).includes(type)) {
+    throw new TypeError(`Unexpected type ${type} used for an alert message.`);
+  }
+
+  return `AlertMessage AlertMessage-${type}`;
+};
+
+const getAlertMessageIcon = (type) => {
+  switch (type) {
+    case 'success':
```
rroppolo, line 34: Super minor but could use the enums here, like ```case MessageTypes.SUCCESS:``` to make them easier to change if we ever need to?

src/Toast/index.js
```diff
@@ -0,0 +1,13 @@
+import Toast from './Toast';
+import AlertMessage, { MessageTypes } from '../Alert/AlertMessage';
```
rroppolo, line 2: Can you move this to be exported from src/Alert/index.js now that it is in its own dir 🎉 ?

stories/Toast/Toast.stories.jsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import { withKnobs, text, radios } from '@storybook/addon-knobs';
+
+import { withToast, MessageTypes, withToastPropTypes } from 'src/Toast';
```
rroppolo, line 4: MessageTypes will need to import here from `src/Alert` if we move it out of `Toast/index.js` (from previous comment)

stories/Toast/Toast.stories.jsx
```diff
@@ -0,0 +1,37 @@
...
+export const ToastMessage = () => (
+  <ToastDummyComponent
+    message={text('Message', 'Your action was a success!')}
+    type={radios('Message Type', MessageTypes, MessageTypes.SUCCESS)}
+  />
+);
+
+const AutoDismissToastComponent = withToast(DummyComponent, { autoDismiss: true });
+
+export const AutoDismissToastMessage = () => (
```
rroppolo, line 32: Now that this is the default behavior, maybe we make this a `ManualDismiss` instead and pass `autoDismiss: false`?

rroppolo (commented):
Sorry for the delay! Had a few more super small comments but this is very close! It looks awesome 🎉

stories/Alert/Alert.stories.jsx
```diff
@@ -2,15 +2,15 @@ import React from 'react';
 import { withKnobs, text } from '@storybook/addon-knobs';
 import { action } from '@storybook/addon-actions';
 
-import { MessageTypes, AlertMessage } from 'src/Flash';
+import Alert, { MessageTypes } from 'src/Alert/Alert';
```
rroppolo, line 5: NBD, non-blocking, but usually if we have an index file we'll do the import from there to have the shorter path, like:
```
import { Alert, MessageTypes } from 'src/Alert';
```

rroppolo (approved):
LGTM! 🎉 💯

PR #184, opened 2021-02-24
Chore/UIDS-177 update style double digit naming convention

jasonbasuil:


rroppolo (approved):
LGTM - love it! 

Sidenote: There is a style linter I had set up on rails-server (it is selectively run only right now) where you can configure it to not allow one-off declarations for certain properties and enforce using variables. I had looked into at first to try to ensure we are using the typography tokens vs. one-off font-size and font-weight declaration. Feels like this is another good candidate now that we have these variables, to encourage usage w/in the team.

PR #186, opened 2021-03-02
feature/UIDS-185 move Header, SidebarNav, SidebarView into DS

jasonbasuil:


src/Header/Header.jsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
```
rroppolo, line 1: Debating whether we want all the layout components grouped under 1 top-level directory.. like src/Layout/Header/* and src/Layout/SidebarNav/*, etc.. since they don't really make as much sense when used independently. Particularly because we have `SidebarNav` and `SidebarView` which look very similar on the top-level..

rroppolo (commented):
Two super small thoughts, let me know what you think!

src/Header/Header.jsx
jasonbasuil: I think that makes sense to restructure the directory like that. Thanks!

PR #194, opened 2021-03-03
fix: moves Toast to bottom right & fixes mobile issue

jasonbasuil:
Screen Shot 2021-03-02 at 4 57 52 PM
Screen Shot 2021-03-02 at 5 45 37 PM
Screen Shot 2021-03-02 at 5 43 57 PM

### DIFF
--- a/src/Toast/Toast.scss
+++ b/src/Toast/Toast.scss
@@ -9,7 +9,7 @@
   right: 0;
   bottom: 0;
   transform: translateX(-10%);
-  width: 18.75rem;
+  width: 19rem;
 
   z-index: $z-index-flash;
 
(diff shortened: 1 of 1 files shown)


rroppolo (approved):
LGTM! 👍  My suspicion is there may be other places in the app where we are using Toasts when maybe we shouldn't be.. Like the CDC warning. But also wondering about some of the error messages we surface with Toasts, and whether they should really be Toasts. Could be an interesting future gardening topic to keep in mind as you switch these over in the app. Where can we be better served showing errors inline or using other patterns vs. relying heavily on Toasts or <gasp!> browser alerts.

PR #208, opened 2021-03-10
Feature/UIDS-188 Add new Modals to DS

jasonbasuil:


package.json
```diff
@@ -2,6 +2,7 @@
   "name": "@user-interviews/ui-design-system",
   "version": "1.14.1",
   "dependencies": {
+    "react-modal": "^3.12.1",
```
rsaris, line 5: Should we be adding this to the `peerDependencies` until we can remove it from the app?

src/Modal/Modal.jsx
```diff
@@ -0,0 +1,31 @@
...
+  );
+};
+
+Modal.propTypes = {
+  className: PropTypes.string,
+  size: PropTypes.oneOf(Object.values(MODAL_SIZES)),
+};
+
+Modal.defaultProps = {
+  size: MODAL_SIZES.small,
```
rsaris, line 27: Is our default "small" or do we default to "medium" and leave space for a "small" one. I can think of a few places we've used a much smaller modal that we may want to reserve for "small".

src/Modal/Modal.jsx
```diff
@@ -0,0 +1,31 @@
...
+  className: '',
```
rsaris, line 28: Should this default to `undefined`?

src/Modal/Modal.jsx
```diff
@@ -0,0 +1,31 @@
...
+  if (size === MODAL_SIZES.LARGE) {
+    className = classNames(className, 'ReactModal--large');
+  }
+
+  return (
+    <ReactModal {...props} className={className} />
+  );
+};
+
+Modal.propTypes = {
```
rsaris, line 21: This may just be me but I kinda like when the DS components list out common props that the library uses so it's easier to find them. Otherwise, you end up having to look at the library that the DS is using which can take away some of the abstraction of the DS.

I know we kinda went overboard w/ the way we did this with ReactSelect but I do think there is some value in this in some capacity.

src/Modal/ModalBody.jsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import './ModalBody.scss';
+
+// This component only provides a class wrapper (with a single CSS property!),
+// but:
+//
+// 1. Keeps the API consistent w/ <ModalHeader> and <ModalFooter>.
+// 2. The old modal had <ModalBody>, so transitioning may be easier.
```
rsaris, line 9: I think we can likely remove these comments -- we have removed the classes it's actually referring to so the context isn't terribly useful here.

src/Modal/ModalBody.jsx
```diff
@@ -0,0 +1,26 @@
...
+  </div>
+);
+
+ModalBody.propTypes = {
+  children: PropTypes.node.isRequired,
+  className: PropTypes.string,
+};
+
+ModalBody.defaultProps = {
+  className: '',
```
rsaris, line 23: Similarly to above -- should this default to `undefined`?

src/Modal/ModalBody.scss
```diff
@@ -0,0 +1,13 @@
...
+.ModalBody {
+  font-weight: $font-weight-normal;
+  padding: $ux-spacing-50;
+
+  &__title {
+    font-family: $font-family;
+    font-size: $font-size-base;
+    font-weight: $font-weight-normal;
+    margin-bottom: $ux-spacing-10;
+  }
```
rsaris, line 12: han-medal

rsaris (approved):
Some nitpicky comments but overall looks good!

package.json
```diff
@@ -44,6 +43,7 @@
     "react": "^16.12.0",
     "react-copy-to-clipboard": "^5.0.2",
     "react-dom": "^16.12.0",
+    "react-modal": "^3.12.1",
```
rroppolo, line 46: If we have it in peerDependencies I think we need devDependencies too, otherwise it won't be available when running storybook inside ui-design-system repo itself

package.json
```diff
@@ -1,8 +1,7 @@
 {
   "name": "@user-interviews/ui-design-system",
-  "version": "1.14.1",
+  "version": "1.13.0",
```
rroppolo, line 3: Looks like this may also be a merge conflict? Seeing some deleted files in the PR as well so not sure if it was a recent rebase or something

stories/Modal/Modal.mdx
```diff
@@ -0,0 +1,40 @@
...
+- Avoid cases where we go from one modal to another modal. 
+- Use when we want to focus the user’s attention.
+- Use to collect input from user.
+- Use to show relevant info. 
+- Should be initiated by the user: don’t use when user hasn’t triggered the modal.
+- They shouldn’t be bloated with many interactions.
+
+## Base modal
+
+- To use in projects as the foundation of any modal. We might need to detach it to customize it.
```
rroppolo, line 27: Do we want to link to the stories in here? For the alerts we showed a preview of each one in the doc like:
```
<Preview>
  <Story id="design-system-alert--success" />
</Preview>
```

It doesn't have great support for knobs, but looks like it could be ok here

stories/Modal/Modal.stories.jsx
```diff
@@ -0,0 +1,140 @@
...
+    size="small"
+  >
+    <ModalHeader
+      title="Transactional modal"
+      titleId="transactional-modal"
+      onRequestClose={handleRequestClose}
+    />
+    <ModalBody>
+      <p>The user needs to take action here.</p>
+      <input placeholder="Email" />
```
rroppolo, line 85: Sorry for the one-off comments! But would you mind using the DS `Input` component for this one to get the input styles?

stories/Modal/Modal.mdx
jasonbasuil: I think I did try putting in the stories for the docs, but since the Modal isn't bound to any root element it covers up the doc. Not sure if there's a good workaround. 

Screen Shot 2021-03-11 at 6 55 34 PM

stories/Modal/Modal.stories.jsx
jasonbasuil: thanks! I forgot that `Input` was available

PR #212, opened 2021-03-16
UIDS-211 Add arrow to Tooltip component

rroppolo:
Added an arrow to the Tooltip component. Also changed the styling of the outline to be a little less obtrusive.

Screen Shot 2021-03-16 at 9 39 42 AM
Screen Shot 2021-03-16 at 9 39 46 AM
Screen Shot 2021-03-16 at 9 40 25 AM

### DIFF
--- a/src/Popper/Popper.jsx
+++ b/src/Popper/Popper.jsx
@@ -5,7 +5,7 @@ import { Manager, Popper as ReactPopper, Reference } from 'react-popper';
 
 import './Popper.scss';
 
-const ARROW_OFFSET = 5;
+const ARROW_OFFSET = 10;
 
 function Popper(props) {
   const { showArrow } = props;
--- a/src/Popper/Popper.jsx
+++ b/src/Popper/Popper.jsx
@@ -5,7 +5,12 @@ import { Manager, Popper as ReactPopper, Reference } from 'react-popper';
 
 import './Popper.scss';
 
+const ARROW_OFFSET = 5;
+
 function Popper(props) {
+  const { showArrow } = props;
+  const offset = showArrow ? ARROW_OFFSET : 0;
+
   return (
     <Manager>
       <Reference>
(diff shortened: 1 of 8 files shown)


rsaris (approved):
Very cool and a great example of how our own DS built on top of other tools is really powerful!

src/Popper/Popper.jsx
```diff
@@ -5,32 +5,56 @@ import { Manager, Popper as ReactPopper, Reference } from 'react-popper';
...
+                        <div className="Popper__header">
+                          {props.header}
+                        </div>
+                      )
+                    }
 
                   {props.text}
+                  {
+                    props.showArrow && (
+                      <span className="Popper__arrow" ref={arrowProps.ref} style={arrowProps.style} />
```
fausgaitan, line 55: I'm not sure where in the code this definition lives:

image

super minor update:
Would it be possible to move the popper 5px so instead of 43px is 48px?

fausgaitan: same for the one that is placed at the bottom of the tooltip. if we can move it down 5px so it's not too close to the tooltip icon that would be great!

fausgaitan (approved):
thanks so much for doing this update @rroppolo ! looks great!!!

jasonbasuil:
Looks great, Rachel!

jasonbasuil (approved):

src/Popper/Popper.jsx
fausgaitan: Maybe not part of this PR, but I couldn't find the knob to place the popper at the bottom of the tooltip and the docs are not showing that variant either. I ask because I see you included that screenshot...

rroppolo: Thanks Faus! That position is managed by the Popper library, but I was able to configure an additional 5px to the offset which looks like:
Screen Shot 2021-03-19 at 9 55 24 AM
Screen Shot 2021-03-19 at 9 55 30 AM

Also added a placement knob, thanks for the suggestion!
Screen Shot 2021-03-19 at 9 55 33 AM

PR #216, opened 2021-03-16
Feature/UIDS-206-update-pills-in-ds

jasonbasuil:
There are now props for icons and a close button.

Screen Shot 2021-03-16 at 1 46 32 PM
Screen Shot 2021-03-16 at 1 46 42 PM


jojolee1996:
Thanks @jasonbasuil, these look great! I noticed a few minor discrepancies with what's in the Figma and was wondering if we'd be able to change these elements:

Icons
- Update blue icon to `$ux-blue`
- Update orange icon to `$ux-orange-700`
- Update yellow icon to `$ux-yellow-700`
- Update green icon to `$ux-green-600`
- Update silver icon to `$ux-gray-700`

Close button
- Reduce the close button's `margin-left` to `0.5rem`
- Update blue close icon to `$ux-blue`
- Update orange close icon to `$ux-orange`
- Update yellow close icon to `$ux-yellow-600`
- Update green close icon to `$ux-green-500`
- Update gray close icon to `$ux-gray-500`
- Update silver close icon to `$ux-gray-500`

rroppolo (commented):
Just a super-small comment on css class naming but looking great!

src/Pill/Pill.jsx
```diff
@@ -1,40 +1,52 @@
...
       classNames(
         'Pill',
-        { 'Pill--squared': squared },
         { [`Pill--${color}`]: !!color },
         { [`Pill--${size}`]: !!size },
       )
     }
   >
+    { icon && (
+    <FontAwesomeIcon className="icon--lead" icon={icon} style={{ marginRight: '0.25rem' }} />
```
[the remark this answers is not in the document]
jasonbasuil: Yeah sometimes the linter is picky about indentation. 

I changed the class name to `Pill_icon--lead` and `Pill__button--close` but wasn't sure if I could do `&__icon--lead` and get the different icon/button colors to apply specifically under each `&--{color}`. Wasn't sure if I'm breaking some convention or BEM rules?

src/Pill/Pill.scss
```diff
@@ -12,51 +12,134 @@
...
   border-radius: 2rem;
   font-size: $font-size-base;
-  padding: $ux-spacing-10 $ux-spacing-30;
+  padding: $ux-spacing-10 $ux-spacing-20;
 
   &--sm {
-    border-radius: 1rem;
-    font-size: .625rem;
+    @include font-type-30;
+    border-radius: $ux-border-radius;
```
rroppolo, line 136: Since we removed the `squared` prop, should we just make this the default for the `Pill` class? It's a little confusing in the story it starts off rounded if I don't specify a size, and then if I choose small it becomes squared, but the size itself is the same. The only difference between no size and size small is the border-radius, so was wondering if that is intentional?

src/Pill/Pill.scss
```diff
@@ -12,51 +12,134 @@
     background-color: $ux-blue-100;
     color: $ux-blue-700;
     border: 0.06rem solid $ux-blue-300;
+
+    .Pill__icon--lead {
```
rroppolo, line 16: I had this same issue in the radio/checkbox buttons which have several top-level modifiers like this one does. I don't really know a great way around it, I guess some people say you can define a variable for the top-level class but honestly I'm not sure I prefer that to what you have here. I think this looks fine & I think that adding the block section makes it more consistent with our ohter components.

src/Pill/Pill.scss
jasonbasuil: Good catch. I ended up making small the default size as well as blue as the default color. I ended up adding a `large` prop as a boolean to make it large size as it's the only other option. If we need to add another size down the road, I could switch back to the `size` prop.

src/Pill/Pill.scss
jasonbasuil: I think this works fine as well. Couldn't think of a quick way around it, but +1 to the block section. Getting better at my naming :)

rroppolo (approved):
LGTM! We'll have to cut this over to update the props when we upgrade the DS but I don't think we have any uses of the large Pill in the app yet, so shouldn't be too bad

PR #222, opened 2021-03-23
UIDS-91 Enable github actions for linting and tests

### LINKED ISSUE #91: Get design system set up on CI

rroppolo:
Closes #91 

Super simple actions to run tests and linter and report output on pull requests. Will send an email to the branch owner if the workflow fails.

Screen Shot 2021-03-22 at 7 42 47 PM

Screen Shot 2021-03-22 at 7 40 31 PM

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,12 @@
+name: Lint javascript
+
+on: push
+
+jobs:
+  js-lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+      - run: yarn
+      - run: yarn lint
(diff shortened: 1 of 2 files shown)


.github/workflows/lint.yml
```diff
@@ -0,0 +1,12 @@
...
+
+on: push
+
+jobs:
+  js-lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+      - run: yarn
```
rsaris, line 11: :question: Why do we have to run `yarn` here? Is it not just OK to use `yarn lint`?

rsaris (approved):

.github/workflows/lint.yml
rroppolo: I tested it just to make sure, but it needs this to install the yarn packages before it can run eslint. I didn't look into caching yet but it is still super quick right now so may not be an issue for a while.

PR #237, opened 2021-03-24
Stories with the components

careful-axe:
Heyas this is a great way to organize that I love! You can keep the stories with the components so that everything is super easy to find. Screenshots show there is no change in display.

I did this super quickly while doing some other stuff so hopefully no weird stuff is in there! I probably want to look at `main.js` again but there is plenty of time for that.

And I left some files behind in the `stories` folder, this is just poc kinda.

After:

Screen Shot 2021-03-24 at 8 42 25 AM

Before:

Screen Shot 2021-03-24 at 8 45 03 AM


wenzeng4 (approved):
Thanks Jane! This looks great

jasonbasuil (approved):
This looks great! I do like having the stories live with the components so I think it's a great start to helping us get more organized. Curious what next steps would look like for more organization for display on Storybook. Exciting!

rroppolo (approved):
Yup totally agree, this is much better, thanks Jane! FYI I think an empty file made it in here `stories/dropdowns.stories.jsx` that can be deleted, it would let me comment inline there

rroppolo:
Oh just noticed the checks are failing, looks like we might need some additional eslint config around dependencies

careful-axe:
@rroppolo bummer though yeah that will have to be resolved. I talked to @jasonbasuil and he isn't making changes to the ds for a bit now so hopefully I can get this resolved tomorrow afternoon after I get some cycle tickets sorted or if anyone else wants to give it a go please do!

Ah yeah that `dropdown` file was going to be the start of something, it can definitely go until there is something actually in it.

careful-axe:
Hey all just heads up for now to avoid `devDependencies` linter error in story files which is super helpful for actual components, we are ignoring `stories` but I will circle back to find a more linter friendly way to resolve this.

PR #239, opened 2021-03-26
add Modal.scss and export MODAL_SIZES

jasonbasuil:


scss/modals.scss
```diff
@@ -0,0 +1,2 @@
+$modal-width-default: 32rem;
+$modal-width-wide: 46.5rem;
```
rroppolo, line 2: Super minor, looks like it is missing a newline at the end. I think there is a way to configure your IDE to do this automatically so you don't have to think about it.

src/Modal/Modal.scss
```diff
@@ -0,0 +1,78 @@
+@import '../../scss/theme.scss';
+@import '~bootstrap/scss/functions';
+@import '~bootstrap/scss/variables';
+
+// The react-modal Modal is distinct from the jQuery-Bootstrap modal, though the
+// same classes can be used for the interior content.
```
rroppolo, line 6: Hmm maybe remove this comment now? There might be one remaining page in the app that uses jQuery-Bootstrap modal but for the purposes of the design system I think we get to start with a clean slate.  I hope no one needs to add new code that uses a jQuery modal at least :)

rroppolo (approved):
LGTM!

PR #250, opened 2021-03-31
update documentation, upgrade storybook v6.2.1

jasonbasuil:
Updated most of the mdx files and updated with some Guidelines (General, Implementation, Copy). It's not completely filled out for a lot of Components, but are docs ever really finished? 🤔  😬

Also upgraded to storybook v6 to make use of the `ArgsTable` addon block

Will continue to update over time.


jasonbasuil:
not sure what's happening with the test. I'm able to run `yarn test` and everything passes.

getting this error 
```
yarn test
  shell: /usr/bin/bash -e {0}
yarn run v1.22.5
$ jest
/bin/sh: 1: jest: not found
error Command failed with exit code 127.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
Error: Process completed with exit code 127.
```

src/Button/Button.mdx
```diff
@@ -0,0 +1,27 @@
...
+# Buttons
+
+
+## Anatomy:
+**Leading icon:** Convey what type of action the button triggers
+**Label:** Short descriptor of button function 
+
+## Best practices
+- A layout should have one prominent button, the others should clearly be secondary 
+- Use sentence casing and `font-weight` regular for all buttons
```
rroppolo, line 14: Wondering if the part about font-weight is needed, if the font styles will be built in to the `Button` component? Mostly noticed b/c we want to use the typography variables to style fonts and not set font-weight directly.

src/Card/Card.mdx
```diff
@@ -0,0 +1,36 @@
...
+
+<Preview>
+  <Story id="design-system-card--default" />
+</Preview>
+
+<ArgsTable of={Card} />
+
+## Anatomy
+
+- **Header:** Brief descriptor of overall card content
```
rroppolo, line 18: Should subtitle be included?

src/Card/Card.mdx
```diff
@@ -0,0 +1,36 @@
...
+- **Divider line:** 
+- **Width:** Card width is defined by, informed by the surrounding layout:
+	- `card-width`: 56rem
+	- `card-width-md`: 40rem
+	- `card-width-sm`: 32rem
+	- `card-width-xs`: 15rem 
+- **Padding:** Cards have a 1.5rem padding
```
rroppolo, line 25: maybe add '1.5rem padding by default' because there is technically a no-padding variant

src/Pill/Pill.mdx
```diff
@@ -0,0 +1,46 @@
...
+  <Story id="design-system-pill--default" />
+</Preview>
+
+<ArgsTable of={Pill} />
+
+## Anatomy
+
+- **Text: ** 
+- **Leading icon: ** relevant icon that makes it easier for users to identify
+- **Close icon: ** so users have the option to clear the pill
```
rroppolo, line 20: I'm not sure if there could be any confusion b/t anatomy and props.. It just took me a sec because text and icon are also props, I was looking for a close icon prop until it occurred to me that it was added by default when you specify an onClose function. I'm not too worried about it but might just be something to keep an eye out for as we get more folks looking into the docs.

src/Pill/Pill.stories.jsx
```diff
@@ -23,12 +29,46 @@ const handleClose = (id) => {
...
+      large={boolean('Large', false)}
+      text={text('Text', 'Text')}
+    />
+    <h4 style={{marginBottom: '2rem', marginTop: '2rem'}} >Large</h4>
+    <Pill
+      color='blue'
+      large
+      text='Text'
+    />
+    <h4 style={{marginBottom: '2rem', marginTop: '2rem'}} >Colors</h4>
```
rroppolo, line 46: 💡 This is probably something for another branch, but might want the Pill component to have some margin by default. I snooped here https://react.carbondesignsystem.com/?path=/story/components-tag--default and looks like they do add margin and remove the left margin from the first pill.

src/Popper/Popper.mdx
```diff
@@ -0,0 +1,49 @@
+import { ArgsTable, Story, Preview } from '@storybook/addon-docs/blocks';
+import { Popper } from './Popper';
+
+
```
rroppolo, line 4: Extra line snuck in (hoping to get a linter enabled on the mdx files soon but it was causing some trouble)

src/ProfileCell/ProfileCell.mdx
```diff
@@ -0,0 +1,47 @@
+import { ArgsTable, Story, Preview } from '@storybook/addon-docs/blocks';
+import { ProfileCell } from './ProfileCell';
+
+import 'bootstrap/dist/css/bootstrap.min.css';
+import '../../scss/global.scss';
+
+# Profile Cell
+Displays information about a user which can include and avatar, name, email, phone number, and timezone.
```
rroppolo, line 8: include an avatar

src/ProfileCell/ProfileCell.mdx
```diff
@@ -0,0 +1,47 @@
...
+- **Phone number** 
+- **Timezone** 
+
+## Best practices
+- Content inside cards should be left-aligned. Deviations are left up to designer's discretion. 
+
+## Stories
+
+### Small
+- Contains only the avatar, first name, last name, and email.
```
rroppolo, line 30: Hmm yeah it is tricky.. technically this is not a requirement that it contain only those fields, and I think in certain places we may show other things in the subtitle, which is why we kept the name of the prop pretty generic. Not sure if we should use more general language, like "Typically contains only the ... " to convey that this is not a hard and fast rule?

src/ProfileCell/ProfileCell.mdx
```diff
@@ -0,0 +1,47 @@
...
+
+<Preview>
+  <Story id="design-system-profile-cell--small" />
+</Preview>
+
+### Large
+- Contains avatar, first name, last name, email, phone number, and timezone.
```
rroppolo, line 37: Same thought here as above. This may be true at the moment, but I'm not sure that it is a requirement of using the component. For example, we probably wouldn't show a phone number for a researcher (not sure if we collect them for researchers?), but I could see a situation arise where we might want to use this component to show a research profile in the large size.

rroppolo (commented):
This is great! Sorry made a lot of comments but also happy to defer things to separate branches, I realize this is an ongoing effort. But I think the yarn packages we may want to resolve, let me know if I can help out there!

package.json
```diff
@@ -51,6 +51,7 @@
   "devDependencies": {
     "@babel/cli": "^7.8.4",
     "@babel/core": "^7.8.4",
+    "@babel/plugin-proposal-class-properties": "^7.13.0",
```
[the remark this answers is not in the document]
jasonbasuil: Yeah once I upgraded to the new Storybook version, this dependency issue was coming up and was looking to this babel dependency specifically. It's not complaining anymore, but yeah I'm wondering at what point it would require other babel dependencies to be upgraded. Seems to be doing fine now tho.

package.json
```diff
@@ -60,17 +61,17 @@
...
+    "@storybook/addon-actions": "^6.2.1",
+    "@storybook/addon-docs": "^6.2.1",
+    "@storybook/addon-jest": "^6.2.1",
+    "@storybook/addon-knobs": "^6.2.1",
+    "@storybook/addon-links": "^6.2.1",
+    "@storybook/addon-storyshots": "^6.2.1",
+    "@storybook/addon-storysource": "^6.2.1",
+    "@storybook/addons": "^6.2.1",
+    "@storybook/react": "^6.2.1",
+    "@storybook/storybook-deployer": "^2.8.7",
```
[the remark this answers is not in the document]
jasonbasuil: Yup! This seems to have worked. Agreed dependencies are so fun 🙃

src/Pill/Pill.stories.jsx
jasonbasuil: Definitely will look into it. I attached this to another issue I'll be working on. 👍

rroppolo (approved):
🥇

PR #253, opened 2021-04-05
feature/UIDS-248 update modal spacing and content

jasonbasuil:
Adjusted padding on `ModalHeader` and `ModalBody` to follow Figma. 
Changed the subtitle and body content to use a light font weight. 
Added Medium sized Modal.

Screen Shot 2021-04-05 at 2 29 43 PM
Screen Shot 2021-03-31 at 9 49 46 AM

### DIFF
--- a/src/Modal/Modal.stories.jsx
+++ b/src/Modal/Modal.stories.jsx
@@ -21,17 +21,63 @@ export default {
 
 const handleRequestClose = () => action('Close');
 
-export const BaseModal = () => (
+export const Default = () => (
   <Modal
     ariaHideApp={false}
-    className="BaseModal"
-    contentLabel="Base Modal"
+    className="DefaultModal"
+    contentLabel="Default Modal"
     isOpen
     size="small"
   >
     <ModalHeader
-      title="Base modal"
-      titleId="base-modal"
+      title="Default modal"
+      titleId="default-modal"
+      onRequestClose={handleRequestClose}
+    />
+    <ModalBody>
+      <p>Some content goes here.</p>
+    </ModalBody>
+    <ModalFooter>
+      <button className="btn btn-light" type="button">Cancel</button>
+      <button className="btn btn-success" type="submit">Save</button>
+    </ModalFooter>
+  </Modal>
+);
+
+export const MediumModal = () => (
+  <Modal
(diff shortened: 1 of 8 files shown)


jasonbasuil:
I know there were conversations about removing `subtitle` from the ModalHeader. As I was looking through rails-server, I did notice that there were actually about 12 or so files that utilized the `subtitle`. I am considering holding off on removing the subtitle (at least for this PR) since it would require a bit of rearranging of the `subtitle` prop and then adding in that same subtitle into the `ModalBody` (on top of needing some better definition of how these subheaders should be presented in the `ModalBody` 
e.g. `<p>` or `<h6>` or whatever we decide) Just considering effort v. impact. Happy to explore this further if there are other opinions!

**`subtitle` in `ModalHeader`**
Screen Shot 2021-04-06 at 1 49 42 PM

**same `subtitle` text moved into `ModalBody`**
Screen Shot 2021-04-06 at 1 48 11 PM

src/Modal/Modal.stories.jsx
```diff
@@ -21,17 +21,63 @@ export default {
...
+  </Modal>
+);
+
+export const MediumModal = () => (
+  <Modal
+    ariaHideApp={false}
+    className="MediumModal"
+    contentLabel="Medium Modal"
+    isOpen
+    size="medium"
```
rroppolo, line 53: Can you update this to use the constant here like MODAL_SIZES.MEDIUM (and he large below), so others will see it as an example and know that they can import the constants?

src/Modal/ModalBody.scss
```diff
@@ -1,8 +1,8 @@
 @import '../../scss/theme';
 
 .ModalBody {
-  font-weight: $font-weight-normal;
-  padding: $ux-spacing-50;
+  font-weight: $font-weight-light;
+  padding: $ux-spacing-40 $ux-spacing-50 $ux-spacing-40 $ux-spacing-50;
```
rroppolo, line 5: 💅 Is this the right nitpick emoji? :) I think you can collapse this down to the shorthand `$ux-spacing-40 $ux-spacing-50`;

rroppolo (approved):
Had two super small suggestions but otherwise looks good!

PR #263, opened 2021-04-14
Feature/UI ds 244 release GitHub action

### LINKED ISSUE #244: DS releases

rroppolo:
Closes #244 

There might be a better way to do this, I sort of put this together by following various tutorials I could find for the different pieces. The goal here is to automate the release process for the design system to make it easier/more streamlined and enable us to have more frequent releases. The workflow is as follows:

- This is a manually triggered github action, so it only runs when you manually request it. However, I haven't tested that part yet as the workflow_dispatch trigger is only supported on the main branch. So in testing I was triggering it on push, but may need a followup branch if there is any issue with the workflow_dispatch trigger once it is merged to main.
- Checkout the code, install the yarn dependencies, etc
- Auto increments the package version by either major, minor or patch (whichever is specified by an input parameter when you run the action). Pushes that update to main to increase the version in package.json.
- Builds the release .tgz
- Creates the github release and the new tag by extracting the updated version number
- Uploads the .tgz file to the release.

What is missing:
- Some kind of commit summary with the release. You can still go in and edit the release and add it manually, but I hope it could be possible to automate that part as well.
- Publishing to the npm repo (This should be possible but I need to figure out how to get the token set up properly. This part is not necessary for our uses for rails-server the way we have things setup currently and is more for if outside folks consume the design system)

I tested it on push as that is all that is supported on a feature branch and it successfully created the release (I deleted it since then) Successful run: https://github.com/user-interviews/ui-design-system/actions/runs/745760439

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,44 @@
+name: Create Release Branch
+on:
+  workflow_dispatch:
+    inputs:
+      releaseType:
+        description: 'Is this release: major, minor or patch?'
+        required: true
+jobs:
+  createrelease:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+      - run: yarn install
+      - run: yarn version ${{ github.event.inputs.releaseType }}
+      - run: yarn build
+      - run: npm pack
+      - run: npm publish --access public
+        env:
+          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+      - name: Extract Package Version
+        id: extract_version
+        uses: Saionaro/extract-package-version@v1.1.1
+      - name: Create Release
+        id: create_release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: v${{ steps.extract_version.outputs.version }}
+          release_name: Release v${{ steps.extract_version.outputs.version }}
+          body_path:
+          draft: false
+          prerelease: false
+      - name: Upload Release Asset
+        id: upload-release-asset
+        uses: actions/upload-release-asset@v1
+        env:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,59 @@
+name: Create Release
+on:
+  workflow_dispatch:
+    inputs:
+      releaseType:
+        description: 'Is this release: major, minor or patch?'
+        required: true
+        default: 'patch'
```
rsaris, line 8: :question: Does this automatically describe minor, major, or patch as the options or does the release manager have to input that?

.github/workflows/release.yml
```diff
@@ -0,0 +1,59 @@
...
+  create-release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+      - run: yarn install
+
+      # Update package version
+      - run: git config --local user.email "41898282+github-actions[bot]@users.noreply.github.com"
+      - run: git config --local user.name "github-actions[bot]"
```
rsaris, line 19: :question: I assume there is no concept of the current user or even ENV vars right? It'd be pretty tyte if we could pull this from the current user but would be even nicer to have the ENV set this up. I always get queezy about having emails / names in code, even if they are relatively innocuous.

rsaris (approved):
Not sure I have a lot of knowledge here on the actions, but everything seems reasonable! I assume this is similar to what we're already doing?

.github/workflows/release.yml
rroppolo: It does not.. you have to enter the string 'patch' or the like. I couldn't find a way to enumerate options or add validation :(  I haven't actually test this part yet because workflow_dispatch will only run on the main branch 🤞

.github/workflows/release.yml
rroppolo: I was able to remove this, it looks like the env variable is still set to the github action bot but at least it isn't hardcoded!

PR #264, opened 2021-04-14
Feature/UIDS-259 add Table to DS 

jasonbasuil:
Might be a good idea to get some initial feedback on the new Table components. I approached it thinking about where it would be first implemented in the app. Product seemed to point towards the Hub Participants table as a starting point. There are a lot of other considerations for Tables in terms of functionality so I'm hoping to incorporate additional functionality + polish it up as we iterate.

Figma mockup below:
https://www.figma.com/file/471iS3QrpPtXZpS1xbsJ20/Components?node-id=4311%3A9429

Screen Shot 2021-04-16 at 3 35 04 PM


src/Table/TableCell.jsx
```diff
@@ -0,0 +1,54 @@
...
+  alignRight,
+  children,
+  className,
+  compact,
+  header,
+  stickyColumn,
+  stickyRow,
+  ...props
+}) => (
+  <td
```
rsaris, line 17: Should this render a `th` if passed `header`? I think we'll likely need this for `stickyRow` too given the mobile support noted here: https://caniuse.com/?search=position%3Asticky

rsaris (commented):
This is looking pretty nifty. I didn't realize `position: sticky` existed. 

My only thought is about how hard it'd be to make it so `TableRow` took the `stickyRow` prop and then inject that into the child elements, but that may be getting a little too cute. What would actually happen if you had a `th` which was not sticky when the other ones were? You can see a solution here: https://stackoverflow.com/questions/50786382/how-to-safely-inject-props-into-react-children and a place that we do this here `app/javascript/common/form/form_group.jsx`. In theory we could have it both ways right?

src/Table/Table.stories.jsx
```diff
@@ -0,0 +1,370 @@
...
+  <MultipleSelectTable />
+)
+
+export const TableWithSorting = () => (
+  <Table>
+    <TableHead>
+      <TableRow>
+        <TableCell header>First name</TableCell>
+        <TableCell header>Last name</TableCell>
+        <TableCell header>Incentives earned <TableSortLabel active/></TableCell>
```
rroppolo, line 235: I don't see an active prop declared in this component unless I missed it, is it being used? Feels like it is a good thing to have, is the intent to style the actively sorted column differently? Also, does `TableSortLabel` need some props to subscribe to clicks/changing of the sort order? Or is that intended for a separate PR?

src/Table/Table.stories.jsx
```diff
@@ -0,0 +1,370 @@
...
+          <TableCell header>Last name</TableCell>
+          <TableCell header>Phone number</TableCell>
+          <TableCell header>Date added</TableCell>
+          <TableCell header>Last invited</TableCell>
+          <TableCell header>Last applied</TableCell>
+        </TableRow>
+      </TableHead>
+      <TableBody>
+        {data.map((row => (
+          <TableRow key={row.id} onClick={() => handleCheckboxSelectSingle(row.id)} selected={selectedRows.includes(row.id)}>
```
rroppolo, line 209: Do we want to configure the cursor to be a pointer on the row if there is an onClick registered, so it is clear that the whole row is clickable?

src/Table/Table.stories.jsx
```diff
@@ -0,0 +1,370 @@
...
+
+  const handlePinClick = () => {
+    setIsStickyColumn((prev) => !prev)
+  }
+
+  return (
+    <Table>
+      <TableHead>
+        <TableRow>
+          <TableCell header stickyColumn={isStickyColumn} stickyRow>Email <span style={{float: "right"}}><PinButton onClick={handlePinClick}/></span></TableCell>
```
rroppolo, line 332: Does sticky column have to be the first cell or could it be located anywhere in the table? Also `stickyRow` is it something intended only for the header row? If so, I was also curious about making this a prop to `TableRow` as Bob mentioned, if it is just supported for one row per table.

rroppolo (commented):
This is really cool Jason. My only slight concern is depending on what the mobile version will look like after the user testing wraps up, will we need to support a layout that could require some of the more complex layout options in css like grid layout or something. But I may prove wrong about that.

src/Table/Table.stories.jsx
```diff
@@ -0,0 +1,370 @@
...
+      )))}
+    </TableBody>
+  </Table>
+);
+
+export const TableOnCard = () => (
+  <Card
+    divided
+    subTitle="Manage a set of templates that can be applied to a project’s automated emails."
+    title="Email Template Sets"
```
rroppolo, line 71: Ah sorry one small think I noticed that I forgot to mention, the overflow on the Card is a little odd. Though not sure if this is an issue with `Card` more in which case I can create a separate issue on it

Screen Shot 2021-04-19 at 2 56 17 PM

fausgaitan: +1

fausgaitan:
when you horizontal scroll this shows up on the left hand side of the table... (I know it's probably the card in SB, but I'll flag it just in case)

image

fausgaitan:
We already discussed this on slack, but I'll add it here just in case: 

`when the column is sticky, only show the shadows when a user scrolls.`

src/Table/Table.stories.jsx
jasonbasuil: Oh yeah I didn't get to the sorting aspect of the Table, but left the TableSortLabel and active prop. It's a WIP 😬  that I'll try to get in this PR, but my understanding is that sorting wasn't on top of the list for priorities for this first iteration

src/Table/TableCell.scss
```diff
@@ -0,0 +1,44 @@
+@import '../../scss/theme.scss';
+
+.TableCell {
+  @include font-type-30--light();
+  border-bottom: 0.06rem solid #EDEDED;
```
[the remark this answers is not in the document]
jasonbasuil: I went off the Figma for the TableCell border-bottom, but we probably can use one of our gray colors

src/Table/Table.stories.jsx
rroppolo: Makes sense! Yeah not sure if any of our tables support sorting yet in the app so totally understand that it is a lower priority

src/Table/Table.stories.jsx
jasonbasuil: Good idea! I ended up adding a `clickable` prop to `TableRow` since I assume that not every table will be "clickable". I think one potential thing to look out for are "clickable" rows that have other "clickable" elements like buttons that would need some sort of microinteraction like a hover or some other visual cue to show that it is also "clickable" since the cursor is already in a pointer state.

fausgaitan (approved):
LGTM! Hopefully, we can address the shadow behavior I mentioned. The rest looks amazing! thanks for working on this!

src/Table/TableCell.scss
```diff
@@ -0,0 +1,44 @@
...
+  padding: $ux-spacing-50;
+  text-align: left;
+
+  &__header {
+    @include font-type-30--bold();
+    background-color: $ux-gray-100 !important;
```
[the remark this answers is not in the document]
jeffbax: wondering whether this type of thing should be configured in the rails SCSS specifically, kind of like the `bootstrap_extensions` encapsulate those types of overrides?

src/Table/Table.stories.jsx
jasonbasuil: I feel like I'm leaning towards making this an issue on the `Card`? I did try giving the `Table` component some styling like :

```
.Table {
  display: block;
  max-width: fit-content;
  overflow-x: auto;
  overflow-y: auto;
  white-space: nowrap;
}
```
Screen Shot 2021-04-21 at 3 01 57 PM

A separate issue that came up with applying this style directly to `Table` was that the `stickyRow` was no longer working (mostly due to `display: block` that I tried). Had trouble figuring out a workaround, so perhaps it just depends on the container that the Table is in and adjusting the css on that?

src/Table/TableCell.scss
```diff
@@ -0,0 +1,44 @@
+@import '../../scss/theme.scss';
+
+.TableCell {
```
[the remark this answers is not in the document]
angrobertsh: Similarly in this screenshot the button width in the actions column seems to be constrained by the length of the header in some sense. I guess I'm wondering what the overall strategy is around column width and overflow and if we can think of any situations where there might be problems (e.g. empty header column)

src/Table/Table.stories.jsx
jasonbasuil: A stickyColumn can be applied to any TableCell yes. I guess the from original designs, I was thinking that only the first and left-most column would be sticky. But I guess a question I have is what if we want to keep a sticky column available (e.g. email or name for ease of scanning while scrolling) and also have a column for selecting? What do you think @fausgaitan ? See vid below:

I agree @rroppolo I'm still trying to figure out the best way to give `TableRow` that single sticky prop so to avoid repeating the prop on all TableCells for the single header row which would be nice.

https://user-images.githubusercontent.com/37383785/115628884-2e54f780-a2b6-11eb-94c2-27a725702288.mov

src/Table/TableCell.scss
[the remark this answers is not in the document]
jeffbax: this might be a matter of having to set min-width on the button, but not sure if that would break out of the cell

src/Table/Table.stories.jsx
```diff
@@ -0,0 +1,396 @@
...
+          <TableCell>{row.dateAdded}</TableCell>
+          <TableCell>{row.lastInvited ? row.lastInvited : `-`}</TableCell>
+          <TableCell>{row.lastApplied ? row.lastApplied : `-`}</TableCell>
+        </TableRow>
+      )))}
+    </TableBody>
+  </Table>
+);
+
+export const TableOnCard = () => (
```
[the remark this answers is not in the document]
jeffbax: Is a no-padding `card` still a `card` or something else entirely at that point?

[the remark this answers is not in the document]
jasonbasuil: I think it's still a `Card`, it might just be some adjustments that need to be made on the `Card` component itself which I'm hoping to address here. https://github.com/user-interviews/ui-design-system/issues/269 potentially make it so you can create something like this so the optionally the body of the card has no padding, and keep the header ... or not keep the header. Make the `Card` more flexible I guess. 

Screen Shot 2021-04-22 at 9 10 03 AM

src/Table/TableCell.scss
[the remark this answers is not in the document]
jeffbax: I don't love the ellipsis solution... I feel like it should be conditional perhaps, but building in a dependency on the hover state doesn't strike me as good for usability. Does that handicap mobile users with no hover? What about screen readers (depending on how they treat the cropped data).

But even regardless of those, for me the point of a table is often the information density itself rather than it looking pretty. If I have to scrub every cell looking for information, I would just find that extremely annoying.

It's admittedly a hard problem, but I don't know that I would prioritize symmetry over convenience here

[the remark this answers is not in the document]
fausgaitan: Yeah, hover should be only for desktop (should have clarified that) and tapping for mobile. 

Just so everyone knows. The way we are making decisions around what variant we want to use for any component in the design system is by conducting preliminary research in the most important design systems out there like Material design, Carbon, Github, etc... 

Some of them were using the popper for long values and we thought it would a good solution. Especially since Material Design probably tested this with end-users. 

Would love to hear suggestions and alternatives to this... 

Totally agree that some users might want to copy the text. Super curious to know whether they actually need to do that in the app or not. 

image
image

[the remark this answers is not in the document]
jeffbax: Just making a note, that if this is a common convention ignore me then though hopefully still food for thought :)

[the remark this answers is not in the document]
rsaris: FWIW IIRC if you try and select all and copy text in a cell which is truncated, your browser will grab the full string. The one thing to be concerned w/ here is if we're overloading the click mechanic we could end up ruining this which would be very annoying.

As for screen readers I can only imagine they ignore the `whitespace` CSS class since I don't see why it would affect them. Therefore, I wouldn't think it'd truncate it for them at all. That is unless of course we're truncating in JS which seems bananas and definitely a bad idea.

src/Table/Table.stories.jsx
```diff
@@ -0,0 +1,483 @@
...
+  { id: 7, firstName: 'Riley', lastName: 'Researcher', email: 'riley@userinterviews.com', phoneNumber: '+18888888888', dateAdded: '01/21/2021', lastInvited: null, lastApplied: null, date: '03/25/2021', boolean: true, decimal: false, pickAny: false, incentivesEarned: 5 },
+  { id: 8, firstName: 'Riley', lastName: 'Researcher', email: 'riley@userinterviews.com', phoneNumber: '+18888888888', dateAdded: '01/21/2021', lastInvited: null, lastApplied: null, date: '03/25/2021', boolean: true, decimal: false, pickAny: false, incentivesEarned: 5 },
+  { id: 9, firstName: 'Riley', lastName: 'Researcher', email: 'riley@userinterviews.com', phoneNumber: '+18888888888', dateAdded: '01/21/2021', lastInvited: null, lastApplied: null, date: '03/25/2021', boolean: true, decimal: false, pickAny: false, incentivesEarned: 5 },
+  { id: 10, firstName: 'Riley', lastName: 'Researcher', email: 'riley@userinterviews.com', phoneNumber: '+18888888888', dateAdded: '01/21/2021', lastInvited: null, lastApplied: null, date: '03/25/2021', boolean: true, decimal: false, pickAny: false, incentivesEarned: 5 },
```
rsaris, line 26: Sorry I didn't mention this before but we likely want to have some more dynamic data to ensure we're showing off what it looks like when we have columns which don't have the same size data. I know this came to bite us at one point where we didn't realize that we weren't properly sizing columns because all our test data was the same length.

src/Table/TableCell.scss
```diff
@@ -0,0 +1,47 @@
...
+    max-height: 3.25rem;
+  }
+
+  &--right {
+    text-align: right;
+  }
+
+  &--sticky-column {
+    position: sticky; 
+    left: 0;
```
rsaris, line 32: Just checking -- will this ruin the ability to have two sticky columns? I don't really know how `left` and `position: sticky` interact.

src/Table/TableCell.scss
```diff
@@ -0,0 +1,47 @@
...
+    z-index: $z-index-10;
+
+    &--corner {
+      z-index: $z-index-20 !important;
+    }
+  }
+
+  &--sticky-row {
+    position: sticky;
+    top: 0;
```
rsaris, line 44: Same question as above w.r.t. two sticky rows.

src/Table/TableRow.jsx
```diff
@@ -0,0 +1,55 @@
...
+  clickable,
+  selected,
+  stickyRow,
+  ...props
+}) => {
+  const addStickyRowProp = (child) => {
+    if (!stickyRow) {
+      return child;
+    }
+      return cloneElement(child, { stickyRow });
```
rsaris, line 19:
```suggestion
    if (!stickyRow) {
      return child;
    }
    return cloneElement(child, { stickyRow });
```

src/Table/TableRow.scss
```diff
@@ -0,0 +1,15 @@
...
+  &:hover {
+    background-color: $ux-gray-100;
+  }
+
+  &--clickable {
+    cursor: pointer;
+  }
+
+  &--selected {
+    background-color: $ux-blue-100 !important;
```
rsaris, line 13: Just checking -- are these `!important` flags here because we want to stop people from overriding or because it's hard to pin down the specificity? In general `!important` scares me as it really limits a consumer from being able to style things. I imagine we could do:
```suggestion
    background-color: $ux-blue-100; 
    
    &:hover {
      background-color: $ux-blue-100;
    }
```
To override the specificity of the above `&:hover` style and should prefer that to using `!important` purely as a way to fight specificity issue. (Note -- this applies to other uses of `!important` in this PR.

src/Table/TableSortLabel.scss
```diff
@@ -0,0 +1,10 @@
+@import '../../scss/theme.scss';
+
+.TableSortLabel {
+  margin-left: $ux-spacing-20;
+  color: $ux-blue;
+
+  &:hover {
+    cursor: pointer;
+  }
+}
```
rsaris, line 10:
```suggestion
}

```

src/Table/TableCell.jsx
```diff
@@ -0,0 +1,101 @@
...
+          {children}
+        </th>
+  ) : (
+    <td
+      className={classNames(
+    'TableCell',
+    className,
+    {
+      [`TableCell--compact`]: !!compact,
+      [`TableCell__header`]: !!header,
```
rsaris, line 60: We don't need this any more because of the ternary right?

src/Table/TableCell.jsx
```diff
@@ -0,0 +1,101 @@
...
+      [`TableCell--right`]: !!alignRight,
+      [`TableCell--sticky-column--corner`]: header && stickyColumn,
```
rsaris, line 62: Same as above -- this can't be the case any more right?

src/Table/TableCell.jsx
```diff
@@ -0,0 +1,101 @@
...
+  return (
+    <>
+      { header ? (
+        <th
+          className={classNames(
+    'TableCell',
+    className,
+    {
+      [`TableCell--compact`]: !!compact,
+      [`TableCell__header`]: !!header,
```
rsaris, line 42: This is always true because of the ternary right?

src/Table/TableCell.jsx
```diff
@@ -0,0 +1,101 @@
...
+      return maxWidthObj;
+    } if (!maxWidth && minWidth) {
+      return minWidthObj;
+    }
+    return null;
+  };
+
+  return (
+    <>
+      { header ? (
```
rsaris, line 35: :bulb: This is probably jsut me, but at this point using `<>` and a ternary feels like overkill and is hard to read. Could we just do a `if (header) { return (<th>...</th>); } return (<td>...</td>);` to make this a little easier to read. Also, contrary to my comments below about the header checks being redundant, we also may want to just build the `className` or even the whole `props` concept up here so that those statements are really small and it's clear you're really rendering the same element just with a different HTML tag.

rsaris (commented):
Sorry for not getting this all into the first review. My biggest concerns are around the use of `!important`. Most of my other comments are mostly opinions.

src/Table/TableRow.scss
jasonbasuil: We decided to give a different background color to the `TableRow--selected` when hovered, so that get rids of `!important` 

https://user-images.githubusercontent.com/37383785/116468765-1d603500-a826-11eb-8b3f-46e492b5a96b.mov

src/Table/TableCell.scss
rroppolo: Was wondering about this above - seems like right now we are targeting the header row for sticky row and the left-most column for sticky column? I actually feel like this might be enough for now and we don't necessarily need to solve for all cases, or multiple sticky row/cols, although one potential useful one I could think of is pinning the right-most column that may contain the edit/action buttons for a row. Happy to be corrected, but feels like pinning a middle-column may not be super useful, and in that case folks can rely on the custom column sorting that has been designed for a future project? For this first pass I feel like header and left-most column makes sense to me, so could even be down to making it clear in the docs that this is what we support out of the gate

src/Table/TableRow.jsx
```diff
@@ -0,0 +1,55 @@
...
+  clickable,
+  selected,
+  stickyRow,
+  ...props
+}) => {
+  const addStickyRowProp = (child) => {
+    if (!stickyRow) {
+      return child;
+    }
+    return cloneElement(child, { stickyRow });
```
rroppolo, line 19: I had a thought of whether we could style this in css on the parent, like 
```
.TableRow--sticky {
  td {
    ...
  }
}
```
though I like that this way is more explicit with each component's css in one place. Was more thinking if we have any performance concerns on a huge table, but it may be that our page sizes are small enough that we wouldn't run into anything like that.

src/Table/TableCell.jsx
```diff
@@ -0,0 +1,92 @@
...
+  const maxWidthObj = { maxWidth };
+  const minWidthObj = { minWidth };
+
+  const getWidthStyling = () => {
+    if (maxWidth && minWidth) {
+      return { ...maxWidthObj, ...minWidthObj };
+    } if (maxWidth && !minWidth) {
+      return maxWidthObj;
+    } if (!maxWidth && minWidth) {
+      return minWidthObj;
```
rroppolo, line 41: Correct me if I'm wrong but I think the `style` prop ignores null/undefined values so while not a big deal either way I think it could not be necessary to enumerate all the cases.

rroppolo (commented):
Just had a few last questions but I think this is looking great and as there are a lot of components to tables happy to extract out any follow-ons to separate branches 👍

src/Table/TableCell.scss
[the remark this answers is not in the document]
jasonbasuil: I can remove the `!important`. It was originally used to keep all the `header` cells the same `ux-gray-100` (mostly due to the `stickyColumn` that is also a `header` which can get pinned). The differences are incredibly subtle and might not even be noticeable.

See Email column with the pin
(w/ `!important`)
Screen Shot 2021-04-29 at 9 10 59 AM

(w/o `!important`) 
Screen Shot 2021-04-29 at 9 17 01 AM

src/Table/Table.mdx
```diff
@@ -0,0 +1,207 @@
...
+<Preview>
+  <Story id="design-system-table--table-with-compact-option" />
+</Preview>
+
+### Table With Sticky Column And Header
+- A Table's header row can be made sticky to ensure column labels are always visible when the users scrolls down
+- `TableRow` can be given a single `stickyRow` prop (within `TableHead`)
+- A Table column can be made sticky to help a user keep track of which row they're currently looking at. The `TableCell` on the column that is made sticky can be given the `stickyColumn` prop (within `TableBody`)
+- The sticky column should usually be the left-most column of the Table
+- A pin icon should be displayed on a sticky column to let the user toggle between a sticky column state and a default state
```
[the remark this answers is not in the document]
jasonbasuil: I was looking around at what other design systems do and found Ant Design's to be pretty neat. They have fixed columns on both ends of their Tables. Could be worth exploring in a future iteration. 

https://codesandbox.io/s/75t1y

I can update the docs to make it clear we support the sticky left-most column only.

src/Table/TableCell.scss
jasonbasuil: Made this comment elsewhere, but I like what Ant Design did with their fixed columns and headers. https://codesandbox.io/s/75t1y

I think our most basic use cases for the Table should be covered. I'd love to explore the other use cases you mentioned Rachel but most likely in a future project. I'd like to get started on getting this Table into the app and make sure all our basic use cases are covered.

src/Table/TableRow.jsx
jasonbasuil: I think we should be okay for now unless there are some expensive calculations happening on our Tables that I'm not aware of.

src/Table/Table.mdx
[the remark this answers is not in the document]
rroppolo: Oh cool! Good find. Yeah I was wondering something like that too - if I click to sticky the a column that is not the left-most it actually applies to that column and all columns to its left. But totally agree for this first pass I think keeping it left-column only is still a really nice improvement.

rroppolo (approved):
LGTM! 👍 I like this as a first release with some of the future tasks you already laid out. Excited to try it out in app, I'm sure we'll learn a lot from that too. Nice work!

src/Table/TableCell.scss
jhforster: Just wanted to say I agree with how @rroppolo and @jasonbasuil are thinking about things here!

PR #268, opened 2021-04-20
feature/UIDS-251 update pill sizing

jasonbasuil:
update default Pill size to `font-type-20`

added a Pill with link variant. If an `href` prop is provided, renders an `<a>` tag. Otherwise renders a `<span>`. Wasn't sure if it was better to render an `<a>` within the existing span to avoid code duplication and then add some scss hover pseudoclass or just do this for better semantics?

~~not sure what's been decided for the `large` variant. Keep it at `font-type-40`?~~
Decision was made to remove the `large` variant for now.

### DIFF
--- a/src/Pill/Pill.jsx
+++ b/src/Pill/Pill.jsx
@@ -14,7 +14,6 @@ const Pill = ({
   href,
   icon,
   id,
-  large,
   onClose,
   text,
 }) => (
@@ -24,7 +23,6 @@ const Pill = ({
         classNames(
           'Pill',
           { [`Pill--${color}`]: !!color },
-          { [`Pill--large`]: !!large },
         )
       }
       href={href}
@@ -43,7 +41,6 @@ const Pill = ({
         classNames(
           'Pill',
           { [`Pill--${color}`]: !!color },
-          { [`Pill--large`]: !!large },
         )
       }
     >
@@ -63,7 +60,6 @@ Pill.propTypes = {
   href: PropTypes.string,
   icon: PropTypes.any,
   id: PropTypes.string,
-  large: PropTypes.bool,
   text: PropTypes.node.isRequired,
   onClose: PropTypes.func,
 };
@@ -73,7 +69,6 @@ Pill.defaultProps = {
   href: undefined,
   icon: undefined,
   id: undefined,
-  large: undefined,
   onClose: undefined,
 };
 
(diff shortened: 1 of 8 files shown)


src/Pill/Pill.jsx
```diff
@@ -5,49 +5,70 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
...
-      )
-    }
-  >
-    { icon && (
-      <FontAwesomeIcon className="Pill__icon--lead" icon={icon} />
-    )}
-    {text}
-    { onClose && (
-    <button className="Pill__button--close" type="button" onClick={() => onClose(id)}> &times;</button>
+  href ? (
```
rroppolo, line 20: Wondering if we want the Pill to be responsible for building the link or knowing about it, or if we want Pill to be more flexible to accept different types of children than just text inside a Pill (and not really need to know/care about their internals). Like something like:
```
<Pill text={<a href="MY LINK">It's a link</a>} />
```
Looks like we already support this as `text` is type `text: PropTypes.node.isRequired` so it can accept jsx.  Which could let you be a little more flexible to combine strings and links together.  (Arguably it is a confusing prop name to call it `text` though if this is supported) Curious what you think!

jasonbasuil: I think you've got a good point going in the direction of just allowing it to accept children. Now that I think about it, making the pill render either a `<span>` or `<a>` was a bit overkill IMO 😅. Could just remove the `text` prop and just have it accept `children` overall. Could be a cleaner solution. If we went this route, would have to do a little bit of cleanup in the cutover, but I think long term the flexibility is good. 

```
<Pill
    color={PILL_COLORS.BLUE)}
    id="3"
  >
    <a href="https://www.userinterviews.com/" rel="noreferrer" target="_blank">Visit our site</a>
  </Pill>
```

rroppolo: 👍 Could support both options for a transition period and the follow up removing the `text` prop once everything is cut over?

jasonbasuil: Actually it shouldn't take that long. There are only 5 files using Pills (total 9 Pills). Could get it done in one go I think so we could just be done with it :)

rroppolo: Some feedback I've received and I think it a good idea is to keep releases backward compatible (just in case someone needs access to something in a release before we get a chance to cut everything over) so might be still worth it to break it into two steps

jasonbasuil: Gotcha! Sounds good. We can do that. 👍

rroppolo (approved):
LGTM!

PR #280, opened 2021-04-29
Moving remaining tests into components, enabling them plus tiny Avatar story fix

careful-axe:


careful-axe:
Going to merge since is just moving some files, nothing has actually changed.

PR #281, opened 2021-04-30
Add KebabDropdown and EllipsisIcon + testing-library!

careful-axe:
Initial commit, just port of common files from `rails-server`.

Also added `testing-library` so we can start to match the testing strategy we are moving towards in `rails-server`.

There is a lot added that looks like it isn't in use, but it was when I flailing around trying to test functionality that wouldn't work because the missing bootstrap dependencies. Those `testing-library` packages will be used shortly since I will circle back and also don't want anyone else to have to face the `toBeInTheDocument is not a function` bug that I did. ;)

Also even if these don't go into use right away if we get them merged in I can use them for the basis of the spike I want to do with fully encapsulated components.


src/KebabDropdown/KebabDropdown.jsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import EllipsisIcon from '../EllipsisIcon';
+
+const KebabDropdown = ({
```
jasonbasuil, line 6: This may be fine for now as a component if it's being used in the app, though I feel like it's a bit too specific to be a general component for a design system. I can see this being broken down into some smaller, reusable components like 

```
<IconButton icon={faEllipsisV} onClick={handleOpen} >
<Dropdown open={handleOpen} anchorEl={anchorEl} >
   <DropdownItem />
   <DropdownItem />
</Dropdown>
```
drawing inspiration again from [Material UI](https://material-ui.com/components/menus/#simple-menu)

Definitely think we can explore this abstraction for better reusability, but again I think the `KebabDropdown` is okay for now. This may just be part of some of the work I've been wanting to get done 😄

src/KebabDropdown/KebabDropdown.stories.jsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+
+import 'jquery';
+import 'popper.js';
+import 'bootstrap/dist/css/bootstrap.min.css';
```
jasonbasuil, line 5: hopefully we'll be at a place where we can move away from using `bootstrap`. Or at least that's my goal 😄

jasonbasuil (commented):
Thanks again for taking the time to bring this over into the DS! I think it's a good start. I know we're currently using it in the app and hopefully we'll be able to make improvements on it by abstracting this into more reusable components. Not sure if it's ready to be exported and used by `ui-design-system`, but I know you mentioned that it would be helpful for some other work you'd be doing? Curious to hear your thoughts!

src/KebabDropdown/KebabDropdown.stories.jsx
careful-axe: I wish! I found it super scary that we are depending on it being in `rails-server` for our components to work... but I saw that is how things are rolling now so I had to follow suit. ;) I really think we need to get away from that even if it means having `react-bootstrap` for a bit until the change can happen per the other pr I have out.

But otherwise the perfect becomes the enemy of the good since trying to escape from Bootstrap is going to be the only thing happening and I want my ds now! Please please please know that once they are discrete components then you can start moving away.

src/KebabDropdown/KebabDropdown.jsx
careful-axe: Heya I gots no skin in the game on KebabDropdown since I am just trying to abstract out some code that was getting rewritten multiple times in the app. Code that works exactly as design expects however, so seems like we should be able to put it into a ds component and then change the implementation behind the scenes later?

Which I think you are agreeing to as well. 

Regarding changing the inners of the component that could be a fun thing for you to do in your copious spare time. Looks legit as long as I don't have to write it.  ;)  And I get a dumb component to use that only cares that I provide an aria label and some children. Bootstrap is easy to hate but I am so seriously in the "I don't want to be in the business of creating even simple components that are cross browser accessible hooks work right etc etc when there are frameworks that do it for us" camp. :)

src/EllipsisIcon/EllipsisIcon.jsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import classNames from 'classnames';
+
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faEllipsisV } from '@fortawesome/free-solid-svg-icons';
+
+const EllipsisIcon = ({ className }) => (
```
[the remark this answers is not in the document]
careful-axe: Actually this thing started because there was a bunch of bootstrap classes on the icon which I since found out where not needed. Totally fine getting rid of this now that I tested enough to realize those classes weren't needed. Was just trying to isolate some bootstrap dependencies.

Original:

`
<FontAwesomeIcon
 className={classNames('action-icon edit', className)}
 icon={faEllipsisV}
/>
`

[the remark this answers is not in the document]
careful-axe: Removed!

src/KebabDropdown/KebabDropdown.jsx
rroppolo: I could see a place for both - a set of buildable Dropdown components and then a shortcut to create the KebabDropdown that we use many places across the app. One thing I wonder though, since you were bringing in react-bootstrap in your other PR Jane (and here we would need to bring in jquery to the design system), if we just make the switch to react-bootstrap right away? Versus the two steps of bringing in jquery first, then swapping it out for react-bootstrap after? Mostly wanting to avoid bringing in jquery if we at all can :) 

https://react-bootstrap.github.io/components/dropdowns API already looks very similar to what you are describing Jason, and possibly is already 1:1 with the bootstrap classes we are using now? Curious what y'all think (sorry for my delayed reply!) I agree we could change the inners later too but I'm definitely into the idea of react-bootstrap replacing bootstrap

careful-axe: Agreed about the both, small components are stellar as are ones you don't have to think about at all. :) Was just waiting for sign off, seems like good time to make the switch and have it available.

careful-axe:
@jasonbasuil and @rroppolo I won't have time to convert this to `react-bootstrap` until a reset week if not later since it is a non-trivial change so could we let this go as is since it does follow the current conventions and has some nice stuff like `testing-library` in it? 

@frankhock is going to take a run at `Accordion` soon and that would be a better one to bring it on since it is part of his cycle work. Bonus he might have a solution for working with with rails-server live while developing which has been on my mind too.

PR #282, opened 2021-05-03
Add Button and Theming

careful-axe:
https://www.figma.com/file/471iS3QrpPtXZpS1xbsJ20/Components?node-id=1252%3A777

### Figma Types:

Figma Button Types

### Design System Types:

Button Types

### Figma Styles:

Figma Button Styles

### Design System Styles:

Button Styles

### DIFF
--- a/src/Button/Button.jsx
+++ b/src/Button/Button.jsx
@@ -1,0 +1,29 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { baseTheme } from '../Theming/theme';
+
+import { StyledButton } from './Button.styles';
+
+const Button = (props) => (
+  <StyledButton {...props}>{props.children}</StyledButton>
+);
+
+Button.propTypes = {
+  children: PropTypes.node.isRequired,
+  icon: PropTypes.any,
+  size: PropTypes.string,
+  theme: PropTypes.object,
+  variant: PropTypes.string,
+  onClose: PropTypes.func,
+};
+
+Button.defaultProps = {
+  icon: undefined,
+  onClose: undefined,
+  size: undefined,
+  variant: 'primary',
+  theme: baseTheme,
+};
+
+export default Button;
--- a/src/Button/Button.stories.jsx
+++ b/src/Button/Button.stories.jsx
@@ -1,0 +1,132 @@
+import React from 'react';
+import { withKnobs } from '@storybook/addon-knobs';
+
+import styled, { ThemeProvider } from 'styled-components';
+
+import Button from 'src/Button';
+import { participantTheme } from '../Theming/theme';
+
+import mdx from './Button.mdx';
+
+export default {
+  title: 'Design System/Button',
+  component: Button,
+  decorators: [withKnobs({ escapeHTML: false })],
+  parameters: {
+    docs: {
+      page: mdx,
+    },
+  },
+};
+
+const StyledH3 = styled.h3`
+    margin: 2rem 0;
+    padding-bottom: .5rem;
+    border-bottom: 1px solid #000;
+`;
+
+const h3Styles = {
+  margin: '2rem 0',
+  paddingBottom: '.5rem',
+  borderBottom: '1px solid #000',
+};
+
+export const Default = () => (
+  <>
+    <StyledH3>Button Types</StyledH3>
+
+    <h4 style={{ marginBottom: '2rem' }}>Primary aka Default</h4>
+
+    <p>Very important actions that usually trigger a longer workflow.</p>
+
+    <p>This button changes color based on set theme. All other buttons stay the same.</p>
+
+    <br />
+    <p><b>Base/researcherTheme Primary Button</b></p>
+
+    <Button variant="primary">Primary</Button>
+    <Button variant="outline-primary">Primary Outline</Button>
+    <Button variant="ghost-primary">Primary Ghost</Button>
+    <Button variant="embedded-primary">Primary Embedded</Button>
+
+    <ThemeProvider theme={participantTheme}>
+      <br />
+      <br />
+      <p><b>participantTheme Primary Button</b></p>
+
+      <Button theme={participantTheme} variant="primary">Participant Primary</Button>
+      <Button theme={participantTheme} variant="outline-primary">Participant Primary Outline</Button>
+      <Button theme={participantTheme} variant="ghost-primary">Participant Primary Ghost</Button>
+      <Button theme={participantTheme} variant="embedded-primary">Participant Primary Embedded</Button>
+    </ThemeProvider>
+
+    <h4 style={{ margin: '2rem 0' }}>Secondary used to be Tertiary Button</h4>
+
+    <p>Currently only used for managing project drafts.</p>
+
+    <Button variant="secondary">Secondary</Button>
+    <Button variant="outline-secondary">Secondary Outline</Button>
+    <Button variant="ghost-secondary">Secondary Ghost</Button>
+    <Button variant="embedded-secondary">Secondary Embedded</Button>
+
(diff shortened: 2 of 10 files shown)


src/Button/Button.jsx
```diff
@@ -0,0 +1,29 @@
...
+import PropTypes from 'prop-types';
+
+import { baseTheme } from '../Theming/theme';
+
+import { StyledButton } from './Button.styles';
+
+const Button = (props) => (
+  <StyledButton {...props}>{props.children}</StyledButton>
+);
+
```
careful-axe, line 11: This button is not real!!! It is just a thing to stick some styles on for now to show how the system could work.

src/Theming/theme.js
```diff
@@ -0,0 +1,122 @@
...
+      color: '#337AB7',
+      backgroundColor: 'transparent',
+      borderColor: 'transparent',
+    },
+    'embedded-primary': {
+      color: '#337AB7',
+      backgroundColor: 'transparent',
+      borderColor: 'transparent',
+    },
+    secondary: baseTheme.button.secondary,
```
careful-axe, line 108: This is horrible, just a quick thing to be sure I wasn't losing anything between the objects. Please look away.

src/Button/Button.styles.js
```diff
@@ -0,0 +1,14 @@
...
+export const StyledButton = styled(Button)`
+  background: ${({ theme, variant }) => theme.button[variant].backgroundColor};
+  color: ${({ theme, variant }) => theme.button[variant].color};
+  border: 2px solid ${({ theme, variant }) => theme.button[variant].borderColor};
+  font-size: 1em;
+  border-radius: 4px;
+  box-sizing: border-box;
+  margin: 1em;
+  padding: 0.25em 1em;
+`;
```
careful-axe, line 14: Again this is not real code, just an example of what could happen...

rroppolo (commented):
Thank you so much for putting this together Jane! This is super helpful for me to better understand styled components and what it would look like if we adopted them. I had some more questions for you, lmk what you think!

On the react-bootstrap side I am 100% in favor. Is it correct that moving from bootstrap to react-bootstrap lets us get rid of our jquery dependency? I agree with you about not starting from scratch implementing UI controls which need to be made accessible, and where there are 3rd party options out there. Had a few questions to try to understand the implications of styled components better since we are already pretty far down the path of using scss

src/Button/Button.stories.jsx
```diff
@@ -0,0 +1,126 @@
...
+
+    <br />
+    <p><b>Base/researcherTheme Primary Button</b></p>
+
+    <Button variant="primary">Primary</Button>
+    <Button variant="outline-primary">Primary Outline</Button>
+    <Button variant="ghost-primary">Primary Ghost</Button>
+    <Button variant="embedded-primary">Primary Embedded</Button>
+
+    <ThemeProvider theme={participantTheme}>
```
rroppolo, line 46: Would this typically be at the top-level of the app? And then Button wouldn't need to receive a theme prop, it would honor the context declared in the component tree?

careful-axe:
@rroppolo thanks so much for the thoughts and questions on this! I definitely want to clarify the line between me knocking out something super simple and eh-hem a bit grotesque and the staggering beauty that styled-components can be.

I think the best thing to do is maybe look a bit at how Grommet uses it: https://github.com/grommet/grommet/blob/master/src/js/themes/base.js

They have pretty much everything in their theme that could be set as magic number or color for backgrounds, padding, margins etc which ends up being majorly useful and super readable to devs and product/design as well. The cool thing is you can find the line of what goes into the theme for yourself though it is so gratifying to use eventually most stuff ends up in there since it makes standardization very simple.

Also completely please disregard how I set up theming in some regards, I was in a massive hurry which is why I set up complete themes for both, usually you wouldn't do that and instead have a smaller subset for the branching - I need to do a bit of a refresher for sure. I also didn't have time to pull in the palette file or review mdx guidelines. Doh.

The cool thing is we can start doing this and just eventually take over everything especially with the design system, I didn't see anything that would lead me to think it would be huge lift to make the shift.

And yes, usually the theme would not be passed in as a prop, it would just live at a high level, that was possibly a limitation of the story file since I had it working within a component initially but just didn't have time to troubleshoot enabling it.

One of the other great things about this approach is that it creates components that can then get styled again within the app if we need to make changes that cannot be accomplished with theming. Though best scenario is that we don't need to.

Answering super quickly and in batch but hopefully this clarifies a bit more before I have to sign off.

src/Button/Button.stories.jsx
```diff
@@ -0,0 +1,126 @@
...
+    border-bottom: 1px solid #000;
+`;
+
+export const Default = () => (
+  <>
+    <StyledH3>Button Types</StyledH3>
+
+    <h4 style={{ marginBottom: '2rem' }}>Primary aka Default</h4>
+
+    <p>Very important actions that usually trigger a longer workflow.</p>
```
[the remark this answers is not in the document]
jasonbasuil: Thanks for pointing this out Rachel! I went ahead and added Storybook documentation how-to + the [Notion Doc](https://www.notion.so/userinterviews1/Storybook-9a3585db57514ec783e39c78518ad5e6) to the README

rsaris (commented):
I love the movement towards a React library for bootstrap. I definitely agree that we should not be rebuilding the wheel for a lot of this stuff and that the DS can be a great abstraction for us to be able to build a standard API for different libraries that we use (e.g. how we wrap React Select). Being able to move to something that's native to React vs. based in jQuery is the true dream of the DS.

As for styled components, I love the idea but I've always had one blocker which is the amount of HTML pages we have which use these styles as well. I think this is especially relevant for buttons given that every form page we use uses those styles. I think the vigor with which we protect the ability to build non-React pages has died down as we've gotten deeper and deeper into React. That said, I wouldn't want that cutover to block this effort or make it so we have to keep everything consistent with the CSS we have currently. How tied is this change over to using CSS in JS vs. using CSS like we are now? I mostly want to set ourselves up for success on whatever cutovers we're making and understanding what it will entail to adopt something new so we can fully embrace whatever we're moving to.

If the goal here is just theming, I actually did something similar purely in CSS here: https://github.com/rsaris/recipe-hub/blob/main/app/javascript/styles/base.scss

jeffbax:
FYI Bootstrap 5 which just released does not require jquery anymore

careful-axe:
@rsaris sorry for the delay in reply I have been focusing on cycle work so I don't get in trouble with @jeffbax ;) 

The big thing I feel very strongly we need to do with the design system is to have it be completely stand alone, not dependent on any css within `rails-server` otherwise we will lose the main benefit of it, which is isolated components that do not change (without theming or very thoughtful and hopefully **extremely** rare overrrides) whether they are in the `rails-server`, storybook or even some other yet to be built app. It is path used by the other design systems with a lot of success, though I will double/triple check that. Style-components creates that isolation with their generated class names automatically which is great. It is a paradigm shift to this style and seems to create redundant code, however avoiding css tug-o-wars is priceless.

But! Keeping the html pages up-to-date with the design system and hopefully staying with best practices is something I would need to research more to find the best solution. And get the scope of what is still not in react... Hopefully there is a path forward to use best practices in the ds and have a way to output usable css for those html elements. Argh though... there are some pain points now of too much css sharing/cascading that I would loooove to put in our rear view. Heh.

@jeffbax I am keeping with cycle work right now, but `react-bootstrap` which is the bootstrap component system we would use is jquery free: https://github.com/react-bootstrap/react-bootstrap/blob/984e22702e811467a06dba084f18414adced47b4/www/src/pages/getting-started/why-react-bootstrap.mdx

@rroppolo and a few others have said go for it so it will likely go in with another pr since this one is mostly a throw away spike.

jeffbax:
Is this PR still relevant? I get pinged every morning about it :)

PR #294, opened 2021-05-13
feature/UIDS-290-table-edits-part-2-multiple-sticky-columns

jasonbasuil:
This edit to the DS Table comes from the need to support multiple sticky columns. This emerged while working in `rails-server` on the Hub Participants view. 

This was the attempt to make adjustments on the `results.scss` for this view, but was running into some overriding issues especially with the multiple select functionality. 

https://user-images.githubusercontent.com/37383785/118170749-53351a00-b3df-11eb-9dcb-44a311e1cd02.mov

I thought it would be better to build support on the DS side instead as a solution, including support for a right-side sticky column should we ever need it in the future.

https://user-images.githubusercontent.com/37383785/118170418-e883de80-b3de-11eb-8b70-79291a020ad6.mov

### DIFF
--- a/src/Table/Table.stories.jsx
+++ b/src/Table/Table.stories.jsx
@@ -331,7 +331,7 @@ export const TableWithStickyColumnAndHeader = () => {
    );
 };
 
-export const TableWithStickyColumnsAndHeader = () => {
+export const TableWithMultipleStickyColumnsAndHeader = () => {
   const [isStickyColumn, setIsStickyColumn] = useState(true);
 
   const handlePinClick = () => {
@@ -475,7 +475,7 @@ export const TableWithMultipleSelect = () => {
   );
 };
 
-export const TableWithMultipleSelectAndStickyColumnsAndHeader = () => {
+export const TableWithMultipleSelectAndMultipleStickyColumnsAndHeader = () => {
   const [selectedRows, setSelectedRows] = useState([]);
   const [isSelectAllCheckboxChecked, setIsSelectAllCheckboxChecked] = useState(false);
 
--- a/src/Table/Table.stories.jsx
+++ b/src/Table/Table.stories.jsx
@@ -97,8 +97,8 @@ export const TableWithFixedColumnWidths = () => (
         <TableRow key={row.id}>
           {/* TODO: Wrap this TableCell in a Popover to show overflow text
               once that component is finished */}
-          <TableCell maxWidth="12rem">{row.email} (with a max-width)</TableCell>
-          <TableCell minWidth="20rem">{row.firstName} (with a min-width)</TableCell>
+          <TableCell maxWidth={192}>{row.email} (with a max-width)</TableCell>
+          <TableCell minWidth={320}>{row.firstName} (with a min-width)</TableCell>
           <TableCell>{row.lastName}</TableCell>
           <TableCell>{row.phoneNumber}</TableCell>
           <TableCell>{row.dateAdded}</TableCell>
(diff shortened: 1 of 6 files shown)


src/Table/Table.stories.jsx
```diff
@@ -324,6 +331,81 @@ export const TableWithStickyColumnAndHeader = () => {
...
+            stickyLeft
+          >
+            Email
+          </TableCell>
+          <TableCell
+            header
+            stickyColumn={isStickyColumn}
+            stickyColumnXOffset={10}
+            stickyLeft
+          >
```
jasonbasuil, line 358: I think I'm mostly curious about how I did the implementation of the multiple sticky columns with these new props. Using `stickyLeft` or `stickyRight` to designate which side of the Table it should stick. It's definitely a valuable use case to have this and an interesting implementation IMO. Would love your thoughts.

src/Table/Table.stories.jsx
```diff
@@ -427,16 +509,23 @@ export const TableWithMultipleSelectAndStickyColumnAndHeader = () => {
     <Table>
       <TableHead>
         <TableRow stickyRow>
-          <TableCell header>
+          <TableCell header stickyColumn={isStickyColumn} stickyColumnXOffset={0} stickyLeft>
             <CheckboxButton id="checkbox" onChange={() => handleCheckboxSelectAll(data)} />
           </TableCell>
-          <TableCell header stickyColumn={isStickyColumn}>Email <span style={{ float: 'right' }}><PinButton isPinActive={isStickyColumn} onClick={handlePinClick} /></span></TableCell>
+          <TableCell header stickyColumn={isStickyColumn} stickyColumnXOffset={3.813} stickyLeft>Email <span style={{ float: 'right' }}><PinButton isPinActive={isStickyColumn} onClick={handlePinClick} /></span></TableCell>
```
jasonbasuil, line 515: Does it seem a bit odd how sometimes you just have to be really precise with the `stickyColumnXoffset` value?

src/Table/TableCell.jsx
```diff
@@ -57,7 +73,7 @@ const TableCell = ({
   return (
     <td
       className={getTableCellClassName()}
-      style={getWidthStyling()}
+      style={{ ...getWidthStyling(), ...getStickyStyling() }}
```
jasonbasuil, line 76: Not the biggest fan of dealing with the `style` prop directly, but couldn't immediately think of a more dynamic, simple solution.

src/Table/TableCell.jsx
```diff
@@ -76,6 +92,9 @@ TableCell.propTypes = {
   maxWidth: PropTypes.string,
   minWidth: PropTypes.string,
   stickyColumn: PropTypes.bool,
+  stickyColumnXOffset: PropTypes.number,
+  stickyLeft: PropTypes.bool,
+  stickyRight: PropTypes.bool,
```
jasonbasuil, line 97: I'm wondering what data type would be better for the `stickyColumnXOffset` ? I know if you're setting `maxWidth` or `minWidth` with something like `"10rem"`, should `stickyColumnXOffset` also take in an explicit string like `"10rem"` instead of the component handling string interpolation.

src/Table/TableCell.jsx
```diff
@@ -13,6 +13,9 @@ const TableCell = ({
   maxWidth,
   minWidth,
   stickyColumn,
+  stickyColumnXOffset,
```
jasonbasuil, line 16: Wondering if there's a better name for this or if it's kind of confusing what this prop is supposed to do?

fausgaitan:
Hey @jasonbasuil , I recently requested some info in the analytics channel regarding the % of users that registered and haven't used first and last name in their profiles. (It's not required at the moment). Unfortunately, Utsav mentioned they can't get that info for us. He said that engineering should be able to run a custom query vs prod to find that out. Do you think that's something you / @rroppolo  could do?

If the % of users that lack first and last name is low, then maybe we can make this new profile sticky, otherwise we should just stick to the email address :/

https://userinterviews.slack.com/archives/CLZTFL82X/p1620758214029200

src/Table/TableCell.jsx
rroppolo: Yeah, feel similarly but I think it makes sense here. We do something similar for styling the react select components. This is another point towards styled components as Jane has proposed, but in our current set-up I can't really think of a better way

src/Table/TableCell.jsx
rroppolo: Hmm good question... I kind of like it as a number, but as mentioned above was thinking about it taking px for the future potential of setting it automatically. Though even then I suppose we could just calculate the width and add the units before setting the prop.. Hmmm I don't have a super strong opinion honestly but I think I lean slightly toward number as you have it. String could maybe be a little confusing what the expected value is, although it is a good point that the style props do work like that. I'm not sure that the added flexibility benefits us much here, though

PR #295, opened 2021-05-18
Feature/UI ds 272 update checkbox groups

### LINKED ISSUE #272: Update Form Elements in Storybook
https://www.figma.com/file/471iS3QrpPtXZpS1xbsJ20/Components?node-id=5502%3A101

rroppolo:
Closes #272 

Figma link: https://www.figma.com/file/471iS3QrpPtXZpS1xbsJ20/Components?node-id=2145%3A396

[ ] Update CheckboxButtonGroup variants to better match Figma
[ ] Add new stories to CheckboxButtonGroup to better match Figma
[ ] Update checkbox font weights to 300


jojolee1996:
Thanks Rachel, this is looking a lot betterl! Leaving my notes below:

**General feedback**
- Update all helper text to be `font-weight:100` and `$ux-gray-900`
- Is there an option within each variant to turn on the border for the entire group like we do under `CheckboxButtonGroup/Default`?

**CheckboxButton / Description**
- Include helper text next to label
- Update children copy to be `font-weight-100`
- When selected, can we change the children copy text color to `$ux-blue-700`? And update border color to `$ux-blue-200` with fill set to `$ux-blue-100`?
- Can we also carry these changes over to `CheckboxButtonGroup/Description` & `Description Row` variants?


**CheckboxButtonGroup / Default Row**
- Add in `margin-right: 0.5rem` to `FormControlLabel__label`

rroppolo:
Thanks @jojolee1996 ! Sorry for the delay - I think I got all your feedback:
[ ] Updated helper text to font-weight: 100 color: ux-gray-900
[ ] Is there an option within each variant to turn on the border for the entire group like we do under CheckboxButtonGroup/Default?
*  ----> Yes there is, though technically it is on the `FormGroup` component

[ ] Include helper text next to label (Done on the last two stories)
[ ] Update children copy to be font-weight-100
[ ] When selected, can we change the children copy text color to $ux-blue-700
[ ]And update border color to $ux-blue-200 with fill set to $ux-blue-100? 
* ---> These two were already set so not sure if I missed where you are talking about!

[ ] Add in margin-right: 0.5rem to FormControlLabel__label
* ----> It was actually easier to increase the margin on FormControlLabel to 1rem since it was already set at 0.5

jojolee1996:
Thanks for making those changes, they look good to me! Regarding these two points:
- When selected, can we change the children copy text color to $ux-blue-700
- And update border color to $ux-blue-200 with fill set to $ux-blue-100?

I was looking at `CheckboxButton/Description` and the selected state doesn't have the blue fill/text. So I was just wondering if this was intentional, since the `CheckboxButtonGroups` have the correct styling.

image

rroppolo:
Hey @jojolee1996, thanks for taking a look! That CheckboxButton story was missing a prop to get the blue highlight, I just added it here if you wouldn't mind checking again: https://docs.userinterviews.com/ui-design-system/?path=/story/design-system-form-elements-checkboxbutton--description

jojolee1996:
Awesome thanks Rachel, LGTM!

scss/forms/control_button_group.scss
```diff
@@ -32,20 +32,9 @@
       padding: 0.3125rem 0.625rem;
       margin-right: 0.5rem;
     }
-
-    .FormControlLabel--active {
```
rroppolo, line 36: This moved to apply to all bordered checkbox variants, it used to only be on the compact style

src/ControlButtonGroup/ControlButtonGroup.jsx
```diff
@@ -25,7 +25,7 @@ const ControlButtonGroup = ({
       onChange: handleChangeValue,
     } : {};
 
-    if (row) {
+    if (row && child.props.bordered === undefined) {
```
rroppolo, line 28: We currently default the buttons to be bordered, so keeping that default unless it is explicitly overridden on the child checkbox component

jasonbasuil (approved):
These look great! Thanks for creating the new stories. Definitely will be useful.

PR #306, opened 2021-06-01
feature/UIDS-300 Adds indeterminate state handling to CheckboxButton

jasonbasuil:


src/CheckboxButton/CheckboxButton.jsx
```diff
@@ -11,21 +17,31 @@ const CheckboxButton = React.forwardRef(({
...
-  />
-));
+}, ref) => {
+  useEffect(() => {
+    if (ref && ref.current) {
+      /* eslint-disable no-param-reassign */
+      ref.current.indeterminate = indeterminate;
+      /* eslint-enable no-param-reassign */
+    }
+  }, [indeterminate, ref]);
```
jasonbasuil, line 27: Not sure if I particularly like the eslint-disable line here. Any thoughts on another way around this?

src/index.js
```diff
@@ -53,6 +53,7 @@ export {
   CardSizes,
   CheckboxButton,
   CheckboxButtonGroup,
+  CHECKED_STATES,
```
jasonbasuil, line 56: Just pulled this constant from the rails-app. Is it really helpful to export from DS though?

rroppolo: Probably overkill for just true/false and I don't know that we would necessarily need to update this for every use of CheckboxButton, but it is kind of nice to have a constant for "indeterminate" so maybe worth it just for that?

src/CheckboxButton/CheckboxButton.mdx
```diff
@@ -0,0 +1,28 @@
...
+
+## Stories
+
+### Default
+
+<Canvas>
+  <Story id="design-system-checkboxbutton--default" />
+</Canvas>
+
+### Indeterminate
```
rroppolo, line 24: Is it worth a sentence about what an indeterminate checkbox means? Like when a checkbox "owns" a number of sub-checkboxes and any one or more of the sub-checkboxes have a different state than the others, the owning checkbox is in the indeterminate state (or "partially" checked).

rroppolo (approved):
Had a few small thoughts for style/clarity but LGTM!

PR #309, opened 2021-06-07
chore/UIDS-308 remove box shadow from inline alert

jasonbasuil:
Inline Alerts should not have box-shadows. Only Toasts should have box-shadows.

Also fixed this weird angled issue on the border-left of Alerts/Toasts, so it should be flush against the sides now.

Screen Shot 2021-03-19 at 10 29 03 AM


rroppolo (approved):
LGTM!

PR #318, opened 2021-06-10
removes indeterminate attribute from CheckboxButton input element

jasonbasuil:
Since we are setting the indeterminate state via JavaScript now, we can remove the indeterminate attribute from the `input` element which helps us avoid that boolean value console error.


rroppolo (approved):
👍

PR #319, opened 2021-06-11
feature/UIDS-317 update Alert to support CTA

jasonbasuil:
Updating the Alert component to support a Call To Action

In preparation for Logistics pod's effort for Connecting Google Calendar

https://www.figma.com/file/btkE6y68VZ5lKqe7vG6oFH/Connect-Google-Calendar---Participant?node-id=29%3A37

Screen Shot 2021-06-11 at 1 54 29 PM


src/Alert/Alert.scss
```diff
@@ -22,11 +22,20 @@
...
   }
 
-  &__close {
+  &__action {
     grid-column-start: 3;
     grid-column-end: 4;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+  }
```
jasonbasuil, line 34: My thinking was that we probably should have a prop that creates another Grid column and aligns the action. Devs could choose to throw the CTA in either the `message` or `action` prop depending on their needs. Not sure if I wanted to only use the `action` prop for the CTA and make the layout a bit more responsive? (e.g. if the Alert gets smaller and there is a CTA, it could move below the message at certain breakpoint). Thoughts?

jasonbasuil:
Setting PR back to draft. Currently working on making the buttons more flexible.

src/Alert/Alert.jsx
```diff
@@ -74,6 +75,22 @@ function Alert(props) {
         }
         {props.message}
       </div>
+      {
+        props.action && (
+        <div className="Alert__action">
+          { React.isValidElement(props.action) ?
+            props.action : (
+              <form action={props.action.url}>
```
rroppolo, line 83: Does this need a method="post"? Also I thought there was some issue with setting method="post" and action -- for example if you look at `app/javascript/researcher/project_workspace/calendar/calendar.jsx` it sets the action through the `ref`, same in the builder code. Have you had a chance to test this out with the connected calendar flow?  I feel like there is some weirdness there..

 I do wonder though about in describing this in the design system vs. just passing the button in via the action prop. I'm curious how many use cases we will be doing a form submit like this vs. an ajax call. Although I see we allow both here so could just see how it gets used.

src/Alert/Alert.scss
```diff
@@ -7,26 +7,48 @@
 
 .Alert {
   display: grid;
-  grid-template-columns: 2rem auto 2rem;
```
rroppolo, line 10: Maybe make the button not word wrap and have it move underneath on mobile?
Screen Shot 2021-07-02 at 12 13 12 PM

src/Alert/Alert.scss
```diff
@@ -54,6 +90,20 @@
   .close {
     color: $ux-blue-700;
   }
+
+  .primary-action {
+    @include font-type-30--bold;
+    background-color: $ux-blue-200;
+    border: none;
```
rroppolo, line 97: 💡 Could make this easier to change if we define a mix-in up top and then pass in the colors as params

rroppolo (commented):
Looks really good, I just had a few thoughts but let me know what you think! Sorry for the delay 🙏

src/Alert/Alert.mdx
```diff
@@ -7,9 +7,9 @@ Alerts give users updates or changes to system status. Providing immediate feedb
 building trust.  While alerts are a great way of communicating with users, they are disruptive and
 should be used cautiously.
 
-<Preview>
+<Canvas>
```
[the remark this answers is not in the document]
jasonbasuil: Same thing, just the renaming came with the Storybook 6 upgrade. Console warnings come up saying `Preview doc block has been renamed to Canvas.` so will probably rename them all in `mdx` files as a gardening chore.

src/Alert/Alert.jsx
jasonbasuil: I ended up just using an `<a>` tag instead. I forget my reasoning for using the `<form>` element in the first place but I feel like this should work

```
{
        props.action && (
        <div className="Alert__action">
          { React.isValidElement(props.action) ?
            props.action : (
              <a
                className={classNames(`Alert-${(props.type)}`, 'primary-action')}
                href={props.action.url}
                rel="noopener noreferrer"
              >
                {props.action.content}
              </a>
          )}
        </div>
      )
}
```

Since these are buttons that seem to be unique to the CTAs on Alert components, my hope is that primarily we use the `action` prop and for most cases you should be able to pass in an object with `{content: 'string', url: 'string'}` for the most part. Otherwise you can pass in a React.element / node for custom CTAs like the GoogleCalendarButton for example

rroppolo (commented):
LGTM! 🎉

PR #320, opened 2021-06-15
UIDS-314 Update radio button group stories

### LINKED ISSUE #314: Update radio button Storybook story

rroppolo:
Closes #314 

Add RadioButton and RadioButtonGroup stories to better match the Figma design system. Updated styles to match CheckboxButton recent updates: https://github.com/user-interviews/ui-design-system/pull/295

### DIFF
--- a/scss/forms/control_button_group.scss
+++ b/scss/forms/control_button_group.scss
@@ -24,13 +24,13 @@
     flex-direction: row;
     flex-wrap: wrap;
 
-    input[type='radio'] {
-      display: none;
-    }
-
     .FormControlLabel--bordered {
       padding: 0.3125rem 0.625rem;
       margin-right: 0.5rem;
+
+      input[type='radio'] {
+        display: none;
+      }
     }
   }
 
--- a/src/RadioButtonGroup/RadioButtonGroup.mdx
+++ b/src/RadioButtonGroup/RadioButtonGroup.mdx
@@ -1,0 +1,33 @@
+import { ArgsTable, Story, Preview } from '@storybook/addon-docs/blocks';
+
+# Radio Button Groups
+
+## Variants
+
+### Default
+
+Single-selection button.
+
+<Preview>
+  <Story id="design-system-form-elements-radiobuttongroup--default"/>
+</Preview>
+
+### Bordered
+
+Single-selection button. Only used in the context of Radio FormGroups.
+
+<Preview>
+  <Story id="design-system-form-elements-radiobuttongroup--bordered-row"/>
+</Preview>
+
+### Description
+
+Single-selection button. Includes helper text and/or a description in addition to the default label.
+
+<Preview>
+  <Story id="design-system-form-elements-radiobuttongroup--description-row"/>
+</Preview>
+
+## Good practices
+
+* When using Default radio buttons, the options should always be stacked vertically (except when used in the context of Inline FormGroups)
--- a/src/RadioButtonGroup/RadioButtonGroup.mdx
+++ b/src/RadioButtonGroup/RadioButtonGroup.mdx
@@ -1,4 +1,4 @@
-import { ArgsTable, Story, Preview } from '@storybook/addon-docs/blocks';
+import { ArgsTable, Story, Canvas } from '@storybook/addon-docs/blocks';
 
 # Radio Button Groups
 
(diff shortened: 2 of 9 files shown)


scss/forms/control_button_group.scss
```diff
@@ -24,13 +24,13 @@
...
-      display: none;
-    }
-
     .FormControlLabel--bordered {
       padding: 0.3125rem 0.625rem;
       margin-right: 0.5rem;
+
+      input[type='radio'] {
+        display: none;
+      }
```
rroppolo, line 33: Moved this out to apply to both types of horizontal variant  of radios (previously only the compact version had a blue background)

Screen Shot 2021-07-01 at 2 28 02 PM
Screen Shot 2021-07-01 at 2 28 10 PM

src/RadioButtonGroup/RadioButtonGroup.mdx
```diff
@@ -0,0 +1,33 @@
...
+
+## Variants
+
+### Default
+
+Single-selection button.
+
+<Preview>
+  <Story id="design-system-form-elements-radiobuttongroup--default"/>
+</Preview>
```
jasonbasuil, line 13: 💅 if you wouldn't mind updating all `<Preview>`s to `<Canvas>` while you're here. I'll do a clean sweep of them all soon :)

src/RadioButtonGroup/RadioButtonGroup.stories.jsx
```diff
@@ -0,0 +1,262 @@
+import React, { useState } from 'react';
+import {
+  boolean, radios, text, withKnobs,
+} from '@storybook/addon-knobs';
+import { ORIENTATIONS } from '../ControlButtonGroup';
```
jasonbasuil, line 5: I'm wondering if it would be helpful to export `ORIENTATIONS`... even though there are only two. Seems like we've been exporting constants anyways? 🤷

src/RadioButtonGroup/RadioButtonGroup.stories.jsx
```diff
@@ -0,0 +1,262 @@
...
+    'Orientation',
+    Object.values(ORIENTATIONS),
+    ORIENTATIONS.COLUMN,
+  );
+  const bordered = boolean('Bordered buttons', false);
+
+  return (
+    <RadioButtonGroupComponent
+      bordered={boolean('Bordered Form Group', false)}
+      defaultValue={[]}
```
jasonbasuil, line 72: Just a heads up, there's a console warning for invalid prop type here

```
Failed prop type: Invalid prop `value` supplied to `RadioButtonGroup`.
in RadioButtonGroup (created by RadioButtonGroupComponent)
```

src/RadioButtonGroup/RadioButtonGroup.mdx
```diff
@@ -0,0 +1,33 @@
...
+
+### Bordered
+
+Single-selection button. Only used in the context of Radio FormGroups.
+
+<Preview>
+  <Story id="design-system-form-elements-radiobuttongroup--bordered-row"/>
+</Preview>
```
jasonbasuil, line 21: Some funky-ness with the Docs page on my end. When you click on any of the buttons other than the ones on the Default story, only the values on the Default story get updated. Looks great on the Canvas view, but wondering if there's a way for the values to update properly on the Docs view?

jasonbasuil (commented):
These look great! Thanks for getting the stories updated. Just left a few small comments so let me know what you think!

src/RadioButtonGroup/RadioButtonGroup.mdx
rroppolo: Done!

src/RadioButtonGroup/RadioButtonGroup.mdx
rroppolo: Looks like it was having an issue because I didn't set unique ids in the stories. Making the ids unique across the apge seemed to fix the issue on docs for me

src/RadioButtonGroup/RadioButtonGroup.stories.jsx
rroppolo: True, we do export the modal sizes for example. I wonder if we'll want to re-evaluate exporting everything at the top-level once we get more of these type of constants/enum props, but probably ok for now

src/RadioButtonGroup/RadioButtonGroup.stories.jsx
rroppolo: Oops copy/paste mistake - thanks for catching that!

src/RadioButtonGroup/RadioButtonGroup.stories.jsx
jasonbasuil: Good call.  It also may not be as obvious to other devs what constants are available / exported from here and which ones they should use (since I know there are many that they can use from rails-server). Good thing to start thinking about!

jasonbasuil (approved):
LGTM! Thanks for making the updates!

PR #327, opened 2021-06-22
[UIDS-326] export colors

### LINKED ISSUE #326: Export colors for use in app
Make js color palette available as an export

michaeldewolf85:
Exports colors for use in the app
Fixes #326


src/index.js
```diff
@@ -28,6 +28,7 @@ import {
   SELECT_SIZES,
   SingleSelect,
 } from 'src/Select';
+import { colors } from 'src/Styles';
```
jasonbasuil, line 31: I'm wondering if it makes sense to export it as `COLORS` since some of our other constants are in that format like `PILL_SIZES.SMALL` or `MODAL_SIZES.LARGE`. Though it would be interesting mixing uppercase with camelCase like `COLORS.uxBlue100`. Or could update the palette to `COLORS.UX_BLUE_100` but that might just be overkill just to be consistent with our patterns so far. Open to thoughts.

michaeldewolf85: I agree with you and my personal preference would be `COLORS.UX_BLUE_100` although 'm a little wary of missing an update somewhere in this repo since I don't know it like the back of my hand

jasonbasuil: I went ahead and updated the naming. Luckily we only had a couple stories that actually used the color palette. 

LGTM! 👍

michaeldewolf85: Thanks Jason! you rock

PR #355, opened 2021-07-22
UIDS-354 Add inputId to react select components.

rsaris:
This is necessary for adding an ID to the input element in React Select for use with `for` on forms.

I had the idea of changing the `id` to point here but that would block the ability to set ID on the react-select element which may cause other issues.


rroppolo (approved):
👍

PR #366, opened 2021-08-02
Feature/UIDS-337 update multiselect styles to match rails-server

### LINKED ISSUE #337: Update option colors/border to blues in AsyncSelect, SingleSelect and remove css in app project builder code
This component in the builder is styled individually in a few places, a good candidate to provide some default styling from teh design system

Screen Shot 2021-07-01 at 2 50 47 PM

rroppolo:
Closes #337 

We override some styles from the design system select components in the app. This PR moves these styles to be the default from the design system. Namely:
- Supporting the multi-select
- Updating the default selected color to blue for the mult-select
- Setting height to auto vs. a fixed height to support line-wrapping on the multi-select dropdown
- Making selected border blue with drop shadow to match app styles
- Set remove button styles from multi-select to match app styles

Screen Shot 2021-08-02 at 12 30 15 PM

### DIFF
--- a/spec/__snapshots__/Storyshots.test.js.snap
+++ b/spec/__snapshots__/Storyshots.test.js.snap
@@ -4527,7 +4527,7 @@ exports[`Storyshots Design System/Selects/Async Default 1`] = `
         className=" css-g1d714-ValueContainer"
       >
         <div
-          className=" css-1wa3eu0-placeholder"
+          className=" css-1269n2i-placeholder"
         >
           Select...
         </div>
@@ -4649,7 +4649,7 @@ exports[`Storyshots Design System/Selects/Async Labeled 1`] = `
         className=" css-g1d714-ValueContainer"
       >
         <div
-          className=" css-1wa3eu0-placeholder"
+          className=" css-1269n2i-placeholder"
         >
           Select...
         </div>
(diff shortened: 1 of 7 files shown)


jojolee1996:
This looks great Rachel! Just two things from me:
* Can we update the `Select...` ghost text to be $ux-gray-800 with a light font-weight?
* We should probably update the hover background-color to the same blue that @jasonbasuil will be updating on single selects (I believe it should be $ux-blue-100)

jasonbasuil (commented):
This is great! I'll just echo JoJo's comments about updating the hover bg color to $ux-blue-100. I'm starting to consolidate the Selects on the rails-server side to use the DS Selects and hopefully it goes smoothly.

Also IIRC @jojolee1996 , we're updating/consolidating the focus states moving forward? Like here:  https://www.figma.com/file/471iS3QrpPtXZpS1xbsJ20/Components?node-id=3324%3A268

since I did add that focus state to the Tiles on pre-builder https://www.figma.com/file/471iS3QrpPtXZpS1xbsJ20/Components?node-id=5511%3A1915

rroppolo:
Thank you @jojolee1996 and @jasonbasuil !

I think I captured all your feedback. I updated the "Select..." placeholder to ux-gray-800, font-weight: light, updated the hover color on options to ux-blue-100, and updated the focus styles to box-shadow `0px 0px 8px 0px #66AFE9B2` which I think is the same we use in the prebuilder now (I was having trouble getting that value in Figma, not sure why 🤔 ) Please let me know if I missed something and thanks again!! Sorry it took a while to get back to this 😬 
Screen Shot 2021-08-17 at 10 35 35 AM

jasonbasuil (approved):
LGTM! 🎉  Thanks so much!

jojolee1996 (approved):
Thanks for making those changes! This looks great 💯

PR #378, opened 2021-08-06
upgrade DS to 'bootstrap' v5.0

jasonbasuil:
Upgraded to 'bootstrap' 5.0. 

Some issues arose in a few places at first glance at the stories, mostly some close buttons on the Alert and ModalHeader as well as some alignment and spacing issues on some of our Form elements. Couldn't pin point exactly what was causing some of these issues, but just managed to adjust some scss and it seems okay now?


src/Modal/ModalHeader.jsx
```diff
@@ -40,12 +40,12 @@ export default class ModalHeader extends Component {
         {this.props.onRequestClose && (
           <button
             aria-label="Close"
-            className="btn btn-sm btn-link close"
```
rroppolo, line 43: I think this might be because the "close" class was renamed to "btn-close"
https://getbootstrap.com/docs/5.0/migration/#close-button

It also switched to using an svg internally so I think we don't need to provide our own 'x' anymore

src/Modal/ModalHeader.scss
```diff
@@ -36,7 +36,11 @@
...
 
   &__close-button {
-    padding: $ux-spacing-40;
-    margin: -1rem -1rem -1rem 0;
+    background-color: unset;
+    border: none;
+    color: $ux-gray-600;
+    &:hover {
+      color: $ux-gray-800;
+    }
```
rroppolo, line 44: I think `btn-close` will have these styles (hopefully)

rroppolo (commented):
Couple small comments -- thank you so much for doing this!!

src/Modal/ModalHeader.jsx
jasonbasuil: Thanks for the feedback and also finding the changelog for the migration! Didn't think to look at that.

PR #382, opened 2021-08-10
feature/UIDS-375 add CreatableSelect into the DS

### LINKED ISSUE #375: Look into bringing the Creatable component from react-select into the design system

jasonbasuil:
closes #375 

It was requested that we look into bringing in the `Creatable` component from `react-select` into the Design System. Looks like we have about 4 files in `rails-server` that are making use of it. Was thinking it might be good to have these available as we're doing updates on Selects.

https://react-select.com/creatable


rroppolo (approved):
LGTM!

PR #404, opened 2021-09-16
updates DS typography to use DM Sans

jasonbasuil:
Thought it might be a good idea to check out the Typography updates in the DS first and view all of our components in isolation. 

Might also be helpful to view the Typography page to view all the font side by side. http://localhost:9009/?path=/story/design-system-styles--typography

Would be great to bring to attention any components that might have a font-weight that looks a little heavier than usual. We can always adjust.

### DIFF
--- a/scss/global.scss
+++ b/scss/global.scss
@@ -1,3 +1,3 @@
-@import url('https://fonts.googleapis.com/css?family=Lato:300,300italic,400,700:latin');
+@import url('https://fonts.googleapis.com/css2?family=Inter:wght@300;400;700&display=swap');
 @import './theme';
 @import './baseline';
--- a/scss/global.scss
+++ b/scss/global.scss
@@ -1,3 +1,3 @@
-@import url('https://fonts.googleapis.com/css2?family=Inter:wght@300;400;700&display=swap');
+@import url('https://fonts.googleapis.com/css2?family=DM+Sans:wght@400;500;700&display=swap');
 @import './theme';
 @import './baseline';
--- a/scss/typography.scss
+++ b/scss/typography.scss
@@ -1,4 +1,4 @@
-$font-family: Inter, sans-serif;
+$font-family: DM Sans, sans-serif;
 $font-weight-light: 300;
 $font-weight-normal: 400;
 $font-weight-bold: 700;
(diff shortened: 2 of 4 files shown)


rroppolo (approved):
Code-wise LGTM! I will defer to the designers on the aesthetics as it does seem the weights are a little stronger on this font.

rroppolo:
FYI @jasonbasuil when we're ready we may want to try out the new git flow release process for this one (merging to develop, then create a release on main) so that we don't block patch releases on the DS for other work. Since this will also require a font upgrade on rails-server (sadly we still define fonts there too as I'm sure you've seen :( ) Let me know maybe we can do it together the first time!

jojolee1996:
Hey @jasonbasuil, sorry for the delay. I put together an updated list of the font-styles using Inter [here](https://www.figma.com/file/471iS3QrpPtXZpS1xbsJ20/Components?node-id=1262%3A0).

Curious to get your thoughts on reducing each `font-size` by 1px, except for `font-type-10.` This will be a better optical match to what we currently have in-app, but would go against the convention of keeping sizes in multiples of 8.

I'm also keen on switching the `bold` font weight to `semi-bold` since the former is a bit too heavy, especially on buttons and labels. @natkasman also noticed that the tracking should be adjusted since Inter's a bit wider at the moment.

jasonbasuil:
> Hey @jasonbasuil, sorry for the delay. I put together an updated list of the font-styles using Inter [here](https://www.figma.com/file/471iS3QrpPtXZpS1xbsJ20/Components?node-id=1262%3A0).
> 
> Curious to get your thoughts on reducing each `font-size` by 1px, except for `font-type-10.` This will be a better optical match to what we currently have in-app, but would go against the convention of keeping sizes in multiples of 8.
> 
> I'm also keen on switching the `bold` font weight to `semi-bold` since the former is a bit too heavy, especially on buttons and labels. @natkasman also noticed that the tracking should be adjusted since Inter's a bit wider at the moment.

Thanks @jojolee1996 @natkasman  I'm a bit hesitant to reduce the font-size since the convention of having multiples of 8 does spread out throughout the app for elements that don't make use of the font-type mixins. 

Throwing out one other exploration: I looked into using `DM Sans` which was one of the original font-types offered as an option in the rebrand which is very similar to `Sofia Pro` on the marketing side. 

[Review app (`Inter`)](https://user-interviews-rs-pr-7825.herokuapp.com/)

[Review app (`DM Sans`)](https://user-interviews-rs-pr-7827.herokuapp.com/)

Ignore the colors as its from the early review apps. I feel like DM Sans is just a tad bit smaller and their bold typeface is just slightly less bold which might be a potential option vs. reducing the size on Inter?

PR #415, opened 2021-10-01
HOTFIX update tooltip primary variant to ux-emerald

rroppolo:
Seen in one place in the app with the old color


jasonbasuil:
@rroppolo this can probably be merged if it's still relevant

PR #428, opened 2021-10-13
chore/UIDS-424 update font types to regular, medium, and bold (400, 500, 700)

jbasuil-ut:
- Updates `font-type` mixins names
- Updates Alert message to `font-type-30`
- Update Pill text to `font-type-20`

### DIFF
--- a/scss/typography.scss
+++ b/scss/typography.scss
@@ -4,130 +4,132 @@ $font-weight-medium: 500;
 $font-weight-bold: 700;
 $font-size-base: 0.875rem;
 
-@mixin font-type-10--medium() {
+@mixin font-type-10() {
   font-size: .625rem;
-  font-weight: $font-weight-medium;
+  font-weight: $font-weight-regular;
   letter-spacing: 0.0625rem;
   line-height: 0.875rem;
 }
 
-@mixin font-type-20--medium() {
+@mixin font-type-20() {
   font-size: .75rem;
-  font-weight: $font-weight-medium;
+  font-weight: $font-weight-regular;
   line-height: 1rem;
 }
 
-@mixin font-type-30--medium() {
-  font-size: 0.875rem;
-  font-weight: $font-weight-medium;
+@mixin font-type-30() {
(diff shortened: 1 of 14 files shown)


scss/typography.scss
```diff
@@ -1,120 +1,133 @@
 $font-family: DM Sans, sans-serif;
-$font-weight-light: 300;
-$font-weight-normal: 400;
+$font-weight-regular: 400;
+$font-weight-medium: 500;
 $font-weight-bold: 700;
 $font-size-base: 0.875rem;
 
-@mixin font-type-10() {
+@mixin font-type-10--medium() {
```
rroppolo, line 7: Hmm I was finding myself kind of missing having one base font without the modifier for our most common case. Thinking more about it, I think what I might be missing is something more like this https://www.carbondesignsystem.com/guidelines/typography/productive where they have 'curated' sets as they call them. This is something we talked about when originally introducing this scale (that we would build these more semantic, curated size/weight combinations from the very explicit typography scale). Was also noticing they added a screen size slider that scales the fonts automatically which is another thing that we've talked about but I hadn't seen displayed like this: https://www.carbondesignsystem.com/guidelines/typography/expressive/#fluid-type-sets

Anyways I think that's my long-winded way of saying I talked myself out of that suggestion and I think this makes sense!

scss/typography.scss
```diff
@@ -1,120 +1,133 @@
...
+  @include font-type-10--medium;
+  font-weight: $font-weight-bold;
+}
+
+@mixin font-type-10--regular() {
+  @include font-type-10--medium;
+  font-weight: $font-weight-regular;
+}
+
+@mixin font-type-10--medium-uppercase() {
```
rroppolo, line 60: I think originally there was some talk about size 10 being too small to pass accessibility without uppercasing it, hence the defaulting it for that size. I'm not totally sure the guidance here so will defer to the designers. I think it was Faus who was saying that way back when but can dig around if there is any official stance out there.

PR #429, opened 2021-10-14
feature/UIDS-427 add IconCell to the DS

jasonbasuil:
Screen Shot 2021-10-14 at 9 37 47 AM

### DIFF
--- a/src/IconCell/IconCell.scss
+++ b/src/IconCell/IconCell.scss
@@ -1,0 +1,14 @@
+@import '../../scss/theme';
+
+.IconCell {
+  align-items: center;
+  background: $ux-blue-200;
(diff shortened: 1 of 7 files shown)


src/IconCell/IconCell.scss
```diff
@@ -0,0 +1,14 @@
+@import '../../scss/theme';
+
+.IconCell {
+  align-items: center;
+  background: $ux-blue-200;
```
rroppolo, line 5: Would there ever be a case for multiple colors here like our alerts/pills?

rroppolo (approved):
Nifty!

jojolee1996 (approved):
Looking good!

PR #454, opened 2021-11-05
Feature/UI ds 453 improve close button aria labels on Alert

tjbarber:
In the situation where we have more than one Alert being displayed, we have multiple buttons with the ARIA label of 'close' which makes it hard to differentiate between the two. This PR will add the type of alert to the label, so the screen reader should read "close error" or "close announcement".

This also makes it easier to target the close button in testing library!


jasonbasuil (approved):
Awesome. Thanks TJ!

rroppolo (approved):
Oops sorry lost track of this with the retreat - LGTM! Can create a patch release once this goes in so you can have it in rails-server

PR #459, opened 2021-11-10
bug/UIDS-384 issue with modal prop single select

### LINKED ISSUE #384: Look into potential issue with inModal prop on SingleSelect
When setting the `modal` prop on `SingleSelect` from the DS, the options are getting cut off in the ModalBody. 

Screen Shot 2021-08-11 at 10 59 04 AM

jasonbasuil:
close #384 

- fixes the issue of the SingleSelect dropdown menu not properly setting the z-index
- adds a story for a SingleSelect inside a Modal
- adds bare minimum documentation for SingleSelect (still a WIP and need to write out best practices, guidelines, etc.)

Reported on RS that the dropdown gets cutoff:
image

Should fix it with proper z-index:
Screen Shot 2021-11-10 at 12 11 56 PM

### DIFF
--- a/src/Select/SingleSelect.mdx
+++ b/src/Select/SingleSelect.mdx
@@ -1,0 +1,60 @@
+import { ArgsTable, Story, Preview } from '@storybook/addon-docs/blocks';
+import SingleSelect from './SingleSelect';
+
+# SingleSelect
+
+A base select input. Flexible and comes with multiselect, clearable, searchable, disabled, and loading support.
+
+<ArgsTable of={SingleSelect} />
+
+## Anatomy
+
+## Best practices
+
+## Guidelines 
+
+### General
+
+### Implementation
+
+### Default
+
+<Preview>
+  <Story id="design-system-selects-single--default" />
+</Preview>
+
+### Searchable
+
+<Preview>
+  <Story id="design-system-selects-single--searchable" />
+</Preview>
+
+### Loading
+
+<Preview>
+  <Story id="design-system-selects-single--loading" />
+</Preview>
+
+### Labeled
+
+<Preview>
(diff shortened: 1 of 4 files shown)


src/Select/SingleSelect.mdx
```diff
@@ -0,0 +1,60 @@
+import { ArgsTable, Story, Preview } from '@storybook/addon-docs/blocks';
+import SingleSelect from './SingleSelect';
+
+# SingleSelect
+
+A base select input. Flexible and comes with multiselect, clearable, searchable, disabled, and loading support.
```
rsaris, line 6: :question: I noticed that we added a `isMulti` to this. Should we just call this `Select` instead of `SingleSelect`? I think when we originally made this component the idea is that there would be a `MultiSelect` sister component which would wrap the isMulti prop.

rsaris (approved):

src/Select/SingleSelect.mdx
jasonbasuil: Interesting. Not sure it's overkill to wrap around to make a new component specifically for a `MultiSelect`. I haven't seen other design system create a separate select component based on what functionality it supports vs. just having a sole Select that handles it through its props. 

I would be for changing it to `Select` at some point. Wasn't sure why it was named `SingleSelect` in the first place so that seems simpler.

rsaris: Cool cool. I agree that having a single component makes more sense (especially since that's what ReactSelect does).

I think this was likely a bit of over engineering / YANGI in trying to separate these out in case we use different solutions for different uses but I don't think that's smart.

We could just make a `SingleSelect` component which renders the new `Select` component to cut over and then update the Rails app to use the new Select signature, but I think that's worth a different PR.

PR #463, opened 2021-11-11
chore(DS-462): Add rails-server local dev workflow

frankhock:
This PR allows local development/debugging against the `rails-server` repo.


jasonbasuil:
Do we still need this draft PR @frankhock ?

PR #464, opened 2021-11-11
chore(DS-462): Add rails-server local dev workflow

### LINKED ISSUE #462: Add script to dev locally against rails-server

frankhock:
This PR allows local development/debugging against the rails-server repo.

closes #462


bin/unlink
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
```
frankhock, line 1: this script ensures all symlinks are removed and gets things back in good standing.

jasonbasuil (approved):
Just gave this a test run. This definitely feels a lot simpler now! Thanks so much Frank! 😄

PR #471, opened 2021-11-15
hotfix/UIDS-470 update Card to conditionally render header

### LINKED ISSUE #470: update Card to conditionally render header

jasonbasuil:
closes #470 

- This should make the `Card` component less opinionated on layout. Ran into this issue before where I was trying to create a flexbox layout, but because the `Card__header` always rendered, it made it difficult to do things like justify-content. Example [here](https://github.com/user-interviews/rails-server/pull/8593#discussion_r746930193)

Note to self: will need to QA on rails-server since there may be places where we were writing custom css to get around this issue.


rroppolo (approved):
👍 LGTM!

PR #472, opened 2021-11-16
Chore/release process readme

rroppolo:
- Adds new release actions to README
  - Let me know if anything is confusing here! Happy to make revisions if so
- Fixes node env on release actions
- Adds prepare hotfix action


.github/workflows/prepare-hotifx.yml
```diff
@@ -0,0 +1,59 @@
+name: "Prepare hotfix"
```
jasonbasuil, line 1: Typo on file name `prepare-hotfix.yml`

jasonbasuil (approved):
Just getting to this! Thanks so much for adding the hotfix workflow and fixing the node version issue. ⭐

.github/workflows/prepare-hotifx.yml
rroppolo: Ooh good catch - thanks!

PR #473, opened 2021-11-22
chore/UIDS-443 Add nightly Percy run in GitHub actions

brianCollinsUI:
@rroppolo not sure the best way to test this out or if this is how we should do it.

I think you need to set the `PERCY_TOKEN` for us since I don't seem to have the settings button. 

How I came up with yaml:
https://github.com/percy/exec-action

### DIFF
--- a/.github/workflows/percy-nightly-tests.yml
+++ b/.github/workflows/percy-nightly-tests.yml
@@ -1,0 +1,20 @@
+name: "Run Percy Tests"
+on:
+  schedule:
+    - cron: "0 0 * * *"
+    branches:
+      - develop
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@master
+      - name: Install
+        run: yarn
+      - name: Percy Test
+        uses: percy/exec-action@v0.x.x
+        with:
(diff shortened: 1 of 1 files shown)


brianCollinsUI:
OK running into a font awesome 401 auth issue when trying to install things for yarn. Maybe we need to do similar `npm config set` things like we do in CI for the rails-server?

brianCollinsUI:
Percy run in Githib Actions

brianCollinsUI:
This will get us nightly builds on `develop`:

> Scheduled workflows run on the latest commit on the default or base branch. 
https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#scheduled-events

I'm thinking of just setting up a slack reminder around the time it completes each day to look for new changes- is there a public slack channel that you would recommend to send this reminder?

rroppolo (approved):
LGTM - thanks Brian!

rroppolo:
Hmm I don't know that we have a slack right now that makes sense.. all the current design system channel have all the designers included which they may not care about getting this ping. Maybe we just create a new one with me you and Jason?

brianCollinsUI:
Sounds good- like to keep the bots separate from people whenever possible

PR #482, opened 2021-12-02
Chore/ Delete old sample file for cypress tests

brianCollinsUI:
Already have coverage for these element here:
https://github.com/user-interviews/ui-design-system/blob/develop/cypress/integration/form_elements/form_group_spec.js#L27-L36

Forgot to delete this one apparently

Screen Shot 2021-12-02 at 5 43 12 AM
y


jasonbasuil (approved):
Sorry for the wait on this. Looks good!

PR #486, opened 2021-12-03
chore/UIDS-476 documentation updates

### LINKED ISSUE #476: Start documentation updates in preparation for Figma cleanup

jasonbasuil:
closes #476 

This was an effort to update the template of how we write documentation in Storybook with the hopes of moving more of the text-based documentation out of Figma and into Storybook. This will definitely be an on-going process so there are some areas that are still WIP / don't have documentation yet, but hopefully we will be a little bit more settled with how we organize these two sources.

Definitely wanting to make this process for updating documentation as smooth as possible, so open to feedback on how to make it better for everyone!

The updated template can be found in the [Storybook notion doc](https://www.notion.so/userinterviews1/Storybook-9a3585db57514ec783e39c78518ad5e6#b0a84cbf91d34d4b88a84d923c7ef087)


src/Alert/Alert.mdx
```diff
@@ -106,3 +108,33 @@ Custom CTA via message prop
...
+
+### Alignment
+
+
+## Best practices
+
+### General
+
+- Alerts should be placed near their related items and embedded in the app. 
+- They should expand within its container.
```
rroppolo, line 133:
```suggestion
- They should expand within their container. 
```

src/Modal/Modal.mdx
```diff
@@ -1,37 +1,88 @@
...
 
 <ArgsTable of={Modal} />
 
-## Anatomy
 
+## Stories
+
+### Default
+
+- To use in projects as the foundation of any modal. We might need to detach it to customize it.
```
rroppolo, line 31: "We might need to detach it to customize it."

Can you elaborate on this? What does it mean to detach the modal? Is there something we can link in the React Modal docs?

rroppolo (approved):
LGTM - had a couple small questions for my own benefit. Thanks for doing all the Preview -> Canvas cleanup 🙏 This template makes sense to me & I like having the consistent structure between stories even if some components don't require as much documentation

src/Modal/Modal.mdx
jasonbasuil: ah. This was probably a copy and paste from the Components Figma library way back when and probably has something to do with detaching components within that context and not the React Modal component itself.

PR #487, opened 2021-12-06
feature/UIDS-474-add-hover-to-tooltip

### LINKED ISSUE #474: Take a look at adding variant to Tooltip to accept different icons / open on hover

jasonbasuil:
closes #474 

- Adds open on hover functionality
- The `Tooltip` already has the ability to swap out icons with the `icon` prop and can customize the icon through `iconClasses`

### DIFF
--- a/src/Tooltip/Tooltip.jsx
+++ b/src/Tooltip/Tooltip.jsx
@@ -48,13 +48,19 @@ class Tooltip extends Component {
     }
   };
 
-  handleToggleTooltip = (event) => {
+  handleToggleTooltipClick = (event) => {
+    if (this.props.withHover) return;
+
     event.preventDefault();
     this.clickOutsideListener = addClickOutsideListener(
       event.target.parentNode,
       this.handleClickOutside,
     );
 
+    this.setState((state) => ({ visible: !state.visible }), this.handleShow);
+  };
+
+  handleToggleTooltipHover = () => {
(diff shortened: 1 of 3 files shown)


src/Tooltip/Tooltip.jsx
```diff
@@ -74,8 +80,10 @@ class Tooltip extends Component {
           className={classNames('Tooltip__icon', this.props.iconClasses)}
           /* eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex */
           tabIndex="0"
-          onClick={this.handleToggleTooltip}
-          onKeyPress={this.handleToggleTooltip}
+          onClick={this.handleToggleTooltipClick}
+          onKeyPress={this.handleToggleTooltipClick}
+          onMouseEnter={this.props.withHover && this.handleToggleTooltipHover}
+          onMouseLeave={this.props.withHover && this.handleToggleTooltipHover}
```
rroppolo, line 86: The only issue is that this tooltip will not be accessible if it is mouse-only (not available to keyboard users). Maybe we can add an onFocus and onBlur? If so, it may be worth extracting out `this.props.withHover && this.handleToggleTooltipHover` to a separate var at the top of render

rroppolo (approved):
Just one comment on the accessibility but I know we're in a time crunch so could address it separately.

PR #500, opened 2021-12-13
UIDS-438 Add AsyncCreatableSelect

### LINKED ISSUE #438: Remove dependency on react-select from rails-server to only use design system selects

rroppolo:
CLoses #438 

Right now we make use of two different versions of react-select on rails-server and uids. This PR sets up some missing functionality to remove from rails-server and only use the DS component.


jasonbasuil (approved):
This is great! Thanks! We probably will need to come back to this and add some documentation for each of these Select components, but that can be on another branch.

PR #502, opened 2021-12-16
feature/UIDS-312 add Dropdown to DS

### LINKED ISSUE #312: Find or implement dropdowns

jasonbasuil:
closes #312

Screen Shot 2021-12-28 at 9 44 59 AM

Screen Shot 2021-12-28 at 9 45 31 AM


src/Dropdown/DropdownToggle.jsx
```diff
@@ -0,0 +1,65 @@
...
+  /**
+   Change the underlying component CSS base class name and modifier class names prefix.
+   This is an escape hatch for working with heavily customized bootstrap css.
+  */
+  bsPrefix: PropTypes.string,
+  /**
+    To pass through to the underlying button or whatever from DropdownButton
+  */
+  childBsPrefix: PropTypes.string,
+  children: PropTypes.node,
```
rroppolo, line 46: 💅 Typically you don't need to include children in the proptypes as every component can accept children

src/Dropdown/Dropdown.types.js
```diff
@@ -0,0 +1,20 @@
+export const DROPDOWN_ALIGN_TYPE = [
+  'start', 'end',
+  { sm: 'start' || 'end' },
+  { md: 'start' || 'end' },
+  { lg: 'start' || 'end' },
+  { xl: 'start' || 'end' },
+  { xxl: 'start' || 'end' },
+];
```
rroppolo, line 8: I'm not sure that this is enumerating all the possible options.. I think you might need to describe this using `PropTypes.shape(` to properly validate the objects as proptypes. For example, when I try to pass `{ lg: 'start' }` I get the warning: 

```
react-jsx-runtime.development.js:124 Warning: Failed prop type: Invalid prop `align` of value `[object Object]` supplied to `Dropdown`, expected one of ["start","end",{"sm":"start"},{"md":"start"},{"lg":"start"},{"xl":"start"},{"xxl":"start"}].
```
maybe something like
```
PropTypes.shape({
    sm: PropTypes.oneOf(['start', 'end']),
}),
PropTypes.shape({
    lg: PropTypes.oneOf(['start', 'end']),
}),
```

Might make sense to change the names of these constants to reflect that they are used for proptypes, or just enumerate the full proptypes directly in the component since they probably will not be able to be re-used elsewhere

rroppolo (commented):
Just one quick thing about the proptypes definition for object proptypes - looks good otherwise!

rroppolo (approved):
Looks great! Thanks!

PR #507, opened 2022-01-04
UIDS-506 Update Form to handle GET method

rsaris:
Previously passing a GET method would have worked for Rails, but it wouldn't
put the params in the URL which is the expected behavior. This makes it so we
properly pass GET directly to the form method parameter.

One other thing this does is remove the CSRF params if GET is passed. These are
ignored in GET and thus aren't necessary and would add extra params to the URL.

One potential upgrade that was left on the cutting room floor was passing any
extra props directly to the form component.

### DIFF
--- a/src/Form/Form.jsx
+++ b/src/Form/Form.jsx
@@ -18,20 +18,21 @@ const Form = forwardRef(({
 }, ref) => {
   const hasMethod = method != null;
   const hasCSRF = CSRFParam && CSRFToken;
+  const isGetFormMethod = method === 'GET';
 
   return (
     <form
       action={action}
       className={classNames('Form', className)}
       id={id}
-      method="POST"
+      method={isGetFormMethod ? 'GET' : 'POST'}
       multipart={multipart}
       name={name}
       ref={ref}
       onSubmit={onSubmit}
     >
-      { hasCSRF && <input name={CSRFParam} type="hidden" value={CSRFToken} /> }
-      { hasMethod && <input name="_method" type="hidden" value={method} /> }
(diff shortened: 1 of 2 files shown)


rsaris:
I noticed this when moving the participant search page over to the new world. It uses a form with a GET to set the filters which is nice for bookmarking.

It looks like I missed this when I wrote `common/form` and so the same behavior was ported over to the DS years ago. This also led me down the rabbit hole of removing that component which should be easy enough, I just have to look into the styles.

rroppolo (approved):
LGTM 👍 Thanks for adding this 🙏  I think this was needed in the survey code as well and may enable us to begin using the DS component there as well

rsaris:
Thanks! Let me know what I need to do to get this released, though also this isn't terribly pressing if we have reasons to delay the release.

jasonbasuil:
> Thanks! Let me know what I need to do to get this released, though also this isn't terribly pressing if we have reasons to delay the release.

Honestly, I feel like we could just do this release. I think what initially has held up a new one was some bootstrap headaches that popped up once `Button` was introduced (aka we imported some bootstrap modules to that component that causes some overrides on RS) Some examples below (e.g. `form-control` font-size increase, `<a>` tag font-size increases). Looking closer, I can probably just deal with these individually first with the new release and shouldn't be that big of an issue until I find a better solution / not waiting on turning these elements into reusable components... if that works with you. 

Screen Shot 2022-01-05 at 10 13 58 AM

Screen Shot 2022-01-05 at 10 13 44 AM

rsaris:
OK -- I have nothing blocked by this at the moment so happy to wait until we can reconcile those changes.

PR #510, opened 2022-01-05
CHORE Fix node version in deploy action

rroppolo:
Fix deploy script to use correct node version

I think this might actually be the real issue with the ally plugin not working on the deployed storybook page


jasonbasuil (approved):
Thanks! Yeah was noticing the deploys to GH pages wasn't working. I think I added a fix for the a11y plugin in a previous PR so the `a11y` parameter might not actually be necessary on the story config? Hope this works 🤞

PR #523, opened 2022-01-19
bug/UIDS-521 wait for accessibility tests (and navigation) before Percy snapshot

### LINKED ISSUE #521: Bug/Percy snapshots need to wait for accessibility tests
Screen Shot 2022-01-18 at 7 41 56 AM
vs
Screen Shot 2022-01-18 at 7 42 01 AM

Thinking I can assert on the tests completed thing in the corner before getting into the iframe and make this part of all the tests

brianCollinsUI:
Resolves #521 

Most recent build on branch:
https://percy.io/f38ae9b9/ui-design-system/builds/15127872

This looks like it fixes the flappy visual tests we have seen due to the accessibility tests running after the component has rendered in the iframe. We also got rid of all the `wait`s since we are asserting on the `Tests completed` text instead which seems to be working better. We also needed to check on the navigation element on the left of the page being rendered as well.

### DIFF
--- a/cypress/support/commands.js
+++ b/cypress/support/commands.js
@@ -20,8 +20,7 @@ Cypress.Commands.add('iframe', { prevSubject: 'element' }, ($iframe) => {
 
 Cypress.Commands.add('takePercySnapshot', (name) => {
   // Check for accessibility tests to finish first
-  // This css selector will change for each storybook release I am guessing
-  cy.get('.css-1imo1gr').should('contain', 'Tests completed');
+  cy.get('body').should('contain', 'Tests completed');
 
   cy.percySnapshot(name);
 });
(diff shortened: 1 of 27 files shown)


cypress/support/commands.js
```diff
@@ -17,3 +17,13 @@ Cypress.Commands.add('iframe', { prevSubject: 'element' }, ($iframe) => {
     });
   });
 });
+
+Cypress.Commands.add('takePercySnapshot', (name) => {
+  // Check for accessibility tests to finish first
+  cy.get('body').should('contain', 'Tests completed');
+
+  // Also wait for the nav to load
+  cy.get('body').should('contain', 'Design System');
```
brianCollinsUI, line 26: Storybook doesn't give you good class names for most of these things, and they also seemed like they would change on version updates, so this is the best I could come up with.
Screen Shot 2022-01-19 at 7 50 30 AM

jasonbasuil (approved):
LGTM! Thanks!

PR #539, opened 2022-02-03
add Steps components to DS

### LINKED ISSUE #537: Introduce or update Steps component to DS

jasonbasuil:
closes #537 

Decided:
- currently set circle size to 2.5rem (40px)
- defaults all colors to $ux-emerald-600

Screen Shot 2022-02-04 at 1 58 22 PM

### DIFF
--- a/cypress/integration/steps_spec.js
+++ b/cypress/integration/steps_spec.js
@@ -1,0 +1,7 @@
+describe('Steps', () => {
+  it('Default', () => {
+    cy.visit('components-steps--default');
+    cy.get('#storybook-preview-iframe').iframe().find('.Steps').should('exist');
+    cy.takePercySnapshot('components-steps--default');
+  });
+});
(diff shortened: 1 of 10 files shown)


cypress/integration/steps_spec.js
```diff
@@ -0,0 +1,7 @@
+describe('Steps', () => {
```
brianCollinsUI, line 1: 🏆

brianCollinsUI:
Should we expect this height and width to all be equal for these circles?
Screen Shot 2022-02-07 at 6 18 07 AM

Not sure the best way to make more even- but altering the `bottom-padding` seems to have an effect, but still is not perfect:

Screen Shot 2022-02-07 at 6 24 09 AM

Any thoughts?

jasonbasuil:
Def would expect a perfect circle! Pixel perfection haha. Will take a look @brianCollinsUI 

Also I'll try to make some adjustments. There are some parts of this that are more static / inflexible so will look into addressing that. I would think that one day we might want a stepper with varying lengths of text to handle something like 

Screen Shot 2022-02-07 at 1 17 53 PM

brianCollinsUI (approved):
looking great! Thanks again, @jasonbasuil

PR #558, opened 2022-03-01
add Grid component to DS

### LINKED ISSUE #557: Add MUI Grid proof of concept

jasonbasuil:
closes #557 

Related notion card: https://www.notion.so/userinterviews1/Card-updates-c5aad600be9746b394757fdacf7c0344

The plan is to remove the default width / margins on the DS `Card` in order to make it a more atomic and un-opinionated. We currently do the following on `Card`:
- currently apply `margin: 1.5rem auto` which is commonly removed
- annoying `width: calc(100% - 1.25rem)` which puts bumpers on the cards which may be unexpected

One potential solution is to introduce a React `Grid` component which I'm pulling in from [Material UI](https://mui.com/components/grid/). 

This PR is pretty bare bones and just wraps the MUI Grid and exports it. No major documentation or best practices have been included. I am imagining that cutover would look something like 1) Release this 2) Wrap all DS Card in this `Grid` as a parent container 3) Do a design audit 4) Remove width / margin from `Card` in separate release. 5) Follow up with documentation / best practices based on what is found during cutover.

I am also hopeful that creating some new guidance and best practices for positioning (not just with Card) will be helpful as I've heard from a couple engineers how layouts are not their favorite thing to work with so I'm hoping to make it easier & create better guidelines.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,8 @@
   "name": "@user-interviews/ui-design-system",
   "version": "1.27.0",
   "dependencies": {
+    "@emotion/react": "^11.8.1",
+    "@emotion/styled": "^11.8.1",
     "@mui/material": "^5.4.3",
     "react-bootstrap": "^2.0.2",
     "react-router-dom": "^5.2.0",
--- a/package.json
+++ b/package.json
@@ -2,6 +2,7 @@
   "name": "@user-interviews/ui-design-system",
   "version": "1.27.0",
   "dependencies": {
+    "@mui/material": "^5.4.3",
     "react-bootstrap": "^2.0.2",
     "react-router-dom": "^5.2.0",
     "react-select": "^3.0.8",
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
     "@fortawesome/react-fontawesome": "^0.1.9",
     "@percy/cli": "^1.0.0-beta.73",
     "@percy/cypress": "^3.1.1",
-    "@popperjs/core": "^2.5.3",
+    "@popperjs/core": "^2.11.2",
     "@storybook/addon-a11y": "^6.2.1",
     "@storybook/addon-actions": "^6.2.1",
     "@storybook/addon-docs": "^6.2.1",
--- a/src/Grid/Grid.jsx
+++ b/src/Grid/Grid.jsx
@@ -9,12 +9,162 @@ const Grid = ({ children, className, ...props }) => (
   </MuiGrid>
 );
 
+// Proptypes are infomed by @mui source code
+// https://github.com/mui/material-ui/blob/master/packages/mui-material/src/Grid/Grid.js
+
 Grid.propTypes = {
+  /**
+   * The content of the component.
+   */
+  children: PropTypes.node,
   className: PropTypes.string,
+  /**
+   * The number of columns.
+   */
+  columns: PropTypes.oneOfType([
+    PropTypes.arrayOf(PropTypes.number),
+    PropTypes.number,
+    PropTypes.object,
+  ]),
+  /**
+   * Defines the horizontal space between the type `item` components.
+   * It overrides the value of the `spacing` prop.
+   */
+  columnSpacing: PropTypes.oneOfType([
+    PropTypes.arrayOf(PropTypes.oneOfType([PropTypes.number, PropTypes.string])),
+    PropTypes.number,
+    PropTypes.object,
+    PropTypes.string,
+  ]),
+  /**
+   * The component used for the root node.
+   * Either a string to use a HTML element or a component.
+   */
+  component: PropTypes.elementType,
+  /**
+   * If `true`, the component will have the flex *container* behavior.
+   * You should be wrapping *items* with a *container*.
+   */
+  container: PropTypes.bool,
+  /**
+   * Defines the `flex-direction` style property.
+   * It is applied for all screen sizes.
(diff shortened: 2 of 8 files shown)


yarn.lock
```diff
@@ -2068,6 +2241,11 @@
   resolved "https://registry.yarnpkg.com/@popperjs/core/-/core-2.10.2.tgz#0798c03351f0dea1a5a4cabddf26a55a7cbee590"
   integrity sha512-IXf3XA7+XyN7CP9gGh/XB0UxVMlvARGEgGXLubFICsUMGz6Q+DU+i4gGlpOxTjKvXjkJDJC8YdqdKkDj9qZHEQ==
 
+"@popperjs/core@^2.4.4":
+  version "2.11.2"
+  resolved "https://registry.yarnpkg.com/@popperjs/core/-/core-2.11.2.tgz#830beaec4b4091a9e9398ac50f865ddea52186b9"
+  integrity sha512-92FRmppjjqz29VMJ2dn+xdyXZBrMlE42AV6Kq6BwjWV7CNUW1hs2FtxSNLQE+gJhaZ6AAmYuO9y8dshhcBl7vA==
+
 "@popperjs/core@^2.5.3", "@popperjs/core@^2.5.4", "@popperjs/core@^2.6.0":
```
rsaris, line 2249: Can we upgrade these `popperjs` deps to use the version we're adding?

yarn.lock
```diff
@@ -5568,6 +5758,11 @@ csstype@^2.5.7:
   resolved "https://registry.yarnpkg.com/csstype/-/csstype-2.6.16.tgz#544d69f547013b85a40d15bff75db38f34fe9c39"
   integrity sha512-61FBWoDHp/gRtsoDkq/B1nWrCUG/ok1E3tUrcNbZjsE9Cxd9yzUirjS3+nAATB8U4cTtaQmAHbNndoFz5L6C9Q==
 
+csstype@^3.0.10:
```
rsaris, line 5761: Can we upgrade the existing usages of this library to use this version as well? We should be able to just remove the other entry from `yarn.lock` and it'll just find this one.

package.json
```diff
@@ -2,6 +2,9 @@
   "name": "@user-interviews/ui-design-system",
   "version": "1.27.0",
   "dependencies": {
+    "@emotion/react": "^11.8.1",
+    "@emotion/styled": "^11.8.1",
```
rsaris, line 6: Are these deps of `@mui`?

rsaris (commented):
Do we have any sense about the increase in bundle size? Always conscious of adding new packages with a lot of deps and the increase in size. Would be a good habit to check that as we add deps to the DS to ensure we're not bloating it too much w/ different libraries.

rsaris:
Overall I think this makes sense -- I agree that we should get some sort of guidance around how we want to use Grids to ensure that Figma mocks etc. are laying things out in a way that makes it obvious how these should be used.

Just checking in that we checked in with design that they're thinking in Grids vs. just flexing things or even using a more CSS `grid` mentality.

src/Grid/Grid.jsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { Grid as MuiGrid } from '@mui/material';
+
+const Grid = ({ children, className, ...props }) => (
+  <MuiGrid className={className} {...props}>
```
rroppolo, line 7: Curious ow does MuiGrid compare to the react-bootstrap grid components? https://react-bootstrap.github.io/layout/grid/

package.json
jasonbasuil: They are required dependencies. Found this about bundle size: https://bundlephobia.com/package/@mui/material@5.4.4

Not sure how it would affect performance and did look into what they do for minimizing bundle size: https://mui.com/guides/minimizing-bundle-size/#when-and-how-to-use-tree-shaking

src/Grid/Grid.jsx
jasonbasuil: In general, I felt like MuiGrid's api was a bit more robust than react-bootstrap. They roughly do the same thing, but I really liked how you're able to handle flex properties on MuiGrid containers themselves through props. Also, I do like the `spacing` system that's built in vs. `react-bootstrap` uses classNames like `className="justify-content-md-center"` which I think we're trying to move away from? https://mui.com/components/grid/#interactive 

My gut feeling is that I'd be able to do more on a `<Grid />` component + props alone and be less likely to write an additional scss file vs. with a react-bootstrap system, but not 100% sure.

Perhaps this is also just me, but I enjoy the mental model of using a single `<Grid />` component vs react-bootstrap `<Container />` `<Row />` and `<Column/>` 

I guess the immediate drawback is that we're introducing a new dependency and the bundle size is something to consider.

rsaris: I am generally in favor of props vs. SCSS when using these libraries, if nothing else, we should make sure what we're exposing outside of the DS is all prop based.

yarn.lock
jasonbasuil: Apparently they rebranded and are now Floating UI 🤔  https://github.com/floating-ui/floating-ui

yarn.lock
jasonbasuil: Not sure if we can? Removing the other version and running yarn just adds it back.

PR #567, opened 2022-03-10
add CardList to DS

### LINKED ISSUE #566: Add CardList to DS

jasonbasuil:
closes #566 

Note: Comment out `margin: $card-spacing auto;` in `src/Card/Card.scss` before viewing.

We're opting to create a layout component `CardList` as an alternative to bringing in `Grid` from MUI. This should be simpler and solve for what we're wanting the accomplish: eventually removing `margin` and centering responsibilities from `Card` . 

This should hopefully handle most cases where we use the DS `Card` as they're just stacked on top of each other and centered. 

Screen Shot 2022-03-10 at 3 39 43 PM

What it would look like on RS:

Screen Shot 2022-03-10 at 3 40 18 PM

### DIFF
--- a/src/CardList/CardList.jsx
+++ b/src/CardList/CardList.jsx
@@ -4,6 +4,8 @@ import PropTypes from 'prop-types';
 
 import './CardList.scss';
 
+export const CardListAlignItemsOptions = ['center', 'flex-start', 'flex-end'];
+
 const CardList = ({
   alignItems,
   className,
@@ -25,7 +27,7 @@ const CardList = ({
   );
 
 CardList.propTypes = {
-  alignItems: PropTypes.oneOf(['center', 'flex-start', 'flex-end']),
+  alignItems: PropTypes.oneOf(CardListAlignItemsOptions),
   className: PropTypes.string,
 };
 
--- a/src/CardList/CardList.jsx
+++ b/src/CardList/CardList.jsx
@@ -1,0 +1,28 @@
+import React from 'react';
+import classNames from 'classnames';
+import PropTypes from 'prop-types';
+
+import './CardList.scss';
+
+const CardList = ({
+  className,
+  children,
+  ...props
+}) => (
+  <div
+    className={classNames('CardList', className)}
+    {...props}
+  >
+    {children}
+  </div>
+  );
+
+CardList.propTypes = {
+  className: PropTypes.string,
+};
+
+CardList.defaultProps = {
+  className: undefined,
+};
+
+export default CardList;
--- a/src/CardList/CardList.jsx
+++ b/src/CardList/CardList.jsx
@@ -5,12 +5,19 @@ import PropTypes from 'prop-types';
 import './CardList.scss';
 
 const CardList = ({
+  alignItems,
   className,
   children,
   ...props
 }) => (
   <div
-    className={classNames('CardList', className)}
+    className={classNames(
+      'CardList',
+      className,
+      {
+        [`CardList__alignItems--${alignItems}`]: !!alignItems,
+      },
+      )}
     {...props}
   >
     {children}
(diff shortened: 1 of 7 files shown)


src/CardList/CardList.jsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import classNames from 'classnames';
+import PropTypes from 'prop-types';
+
+import './CardList.scss';
+
+const CardList = ({
+  alignItems,
```
jasonbasuil, line 8: Not sure if we really need the ability to flex-start or flex-end with this `alignItems`, but happy to remove if we want to be more strict about this.

rsaris: We definitely have things that we prefer to left align vs. center so I think this is exactly right.

src/CardList/CardList.jsx
```diff
@@ -0,0 +1,37 @@
...
+      },
+      )}
+    {...props}
+  >
+    {children}
+  </div>
+  );
+
+CardList.propTypes = {
+  alignItems: PropTypes.oneOf(['center', 'flex-start', 'flex-end']),
```
rsaris, line 28: Could we make these constants that we export?

src/CardList/CardList.jsx
```diff
@@ -0,0 +1,37 @@
...
+}) => (
+  <div
+    className={classNames(
+      'CardList',
+      className,
+      {
+        [`CardList__alignItems--${alignItems}`]: !!alignItems,
+      },
+      )}
+    {...props}
```
rsaris, line 21: This is probably a bigger discussion, but it's a little odd to me to pass classname and props here. Do we have an expected use case for this or do we think of this as a convention we're going to use. It feels like it could make for a really difficult migration in the future.

src/CardList/CardList.jsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import classNames from 'classnames';
+import PropTypes from 'prop-types';
+
+import './CardList.scss';
+
+const CardList = ({
+  alignItems,
+  className,
+  children,
```
rsaris, line 10: I notice we don't have `propTypes` for children below. I think this is right but do we document that convention anywhere in the DS?

src/CardList/CardList.stories.jsx
```diff
@@ -0,0 +1,62 @@
...
+    docs: {
+      page: mdx,
+    },
+  },
+};
+
+const CardListAlignItemsOptions = ['center', 'flex-start', 'flex-end'];
+
+export const Default = () => (
+  <>
```
rsaris, line 22: Do we need this parent element?

rsaris (approved):
Looks great -- left some documentation changes but should be good as is.

src/CardList/CardList.jsx
jasonbasuil: Yeah it's a little bit odd. I think it's the only place we do this, but I think this should be okay for now as I don't really see this component needing other flex props other than this one and only works on a single flex direction: column. Can definitely revisit if we want it to be more robust.

src/CardList/CardList.jsx
jasonbasuil: Still not sure on what our convention is yet for exporting constants. It makes sense for things like colors and other js variables, but I think the prop validations and documentation should suffice unless I'm missing something where exporting is more beneficial?

rsaris: FWIW my preference is to always export constants as it allows IDEs to show the options vs. having to go to the docs. Also that makes it easier as a consumer to know that you typed it wrong vs. having a string.

src/CardList/CardList.jsx
rsaris: Yeah -- I think the hope is that we can add props here vs. monkey patching as a consumer via props or a class name (as we talked about in a thread somewhere).

src/CardList/CardList.jsx
rsaris: Yeah -- I think this is right because React already validates this. My understanding is that you should only have propTypes on children if you want to be extra restrictive. I believe we have this rule in the rails server.

rroppolo: Having trouble finding it in the React docs but yes my understanding is it is typically React convention to not include proptypes for children since every component accepts it. Unless you are doing something like restricting to one node or a single type of node

PR #572, opened 2022-03-24
remove margin from Card

### LINKED ISSUE #549: Remove Card width and margin props
We'll have to audit the app to make sure no places are relying on this. Generally they're actually setting `width: 100%;` and `margin: 0` to counteract these default styles.

Context: https://userinterviews.slack.com/archives/CJYUKSM5M/p1644251665318009

jasonbasuil:
closes #549 

This removes centering responsibility off of `Card`. Responsibility falls under (soon to be renamed) `CardContainer`


rroppolo (approved):
Yay! 🎉

PR #576, opened 2022-03-24
WIP!!! Drag and drop proof of concept

careful-axe:


careful-axe:
Code lives in rails-server...

PR #593, opened 2022-04-20
add brand variants to Button

### LINKED ISSUE #592: Add brand variants to Button

jasonbasuil:
closes #592 

While working on transitioning the TrackedButton to use the DS Button, I found the need for adding variants for the different brands that we use primarily for authentication. One example is the `OauthButton`. 

Lmk if we have any others that we might be missing!

Screen Shot 2022-04-20 at 12 00 37 PM

### DIFF
--- a/cypress/integration/button_spec.js
+++ b/cypress/integration/button_spec.js
@@ -23,6 +23,12 @@ const buttons = [
     class: '.Button',
     match: 'Skip',
   },
+  {
+    name: 'Brands',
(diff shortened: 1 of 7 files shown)


wenzeng4:
I think we need to update the Facebook blue to be compliant with their new brand guidelines: https://www.facebook.com/brand/resources

Thanks!

rroppolo (approved):
LGTM! 👍

PR #601, opened 2022-04-29
update focus state to Button and Select

### LINKED ISSUE #598: Update focus state on Buttons to be contrast accessible

jasonbasuil:
closes #598 

Applies to all Button variants as well as Dropdown
Screen Shot 2022-04-29 at 2 10 56 PM
Screen Shot 2022-04-29 at 2 10 48 PM

Applies to all Selects
Screen Shot 2022-04-29 at 2 10 12 PM

### DIFF
--- a/scss/buttons.scss
+++ b/scss/buttons.scss
@@ -20,7 +20,13 @@
 
 $primary: $ux-emerald-600;
 $danger: $ux-red;
-$warning: $ux-yellow-400; 
+$warning: $ux-yellow-400;
+
+@mixin btn-focus-outline {
+  box-shadow: none !important;
+  outline: 0.125rem solid $ux-blue-500;
+  outline-offset: 0.125rem;
+}
 
 @mixin btn-primary {
   $btn-primary-background: $primary;
@@ -48,6 +54,10 @@ $warning: $ux-yellow-400;
     $btn-primary-active-border,
     $btn-primary-active-color,
   );
+
+  &:focus, &:active {
+    @include btn-focus-outline;
+  }
 }
 
 @mixin btn-outline-primary {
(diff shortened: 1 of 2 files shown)


scss/buttons.scss
```diff
@@ -20,7 +20,13 @@
 
 $primary: $ux-emerald-600;
 $danger: $ux-red;
-$warning: $ux-yellow-400; 
+$warning: $ux-yellow-400;
+
+@mixin btn-focus-outline {
+  box-shadow: none !important;
```
rroppolo, line 26: Curious - where is this being overridden to need the important? If this is controlled by a bootstrap css variable wondering if we could update that var vs. using the !important

jasonbasuil: I think I was looking at the `button-variant` mixin and it seemed like it was handling the `box-shadow` in a few places for the different states. It seemed like the easiest way to get rid of it both in `Button` and `Dropdown` (which also has it's own handling of `box-shadow` under the `button-variant` mixin. 

Noticed that combining the pseudo selectors should work for `Button`, but isn't handling the removal for `Dropdown`.  Wondering if it's okay to use `!important` in this case to handle all the removals since we won't be using `box-shadow` ever? 
```
&:focus, &:active, &:focus:active {
    @include btn-focus-outline;
  }
```

https://getbootstrap.com/docs/5.1/components/buttons/#mixins

rroppolo: Ooh sorry this fell off my radar! Was wondering if we could override  ```$btn-focus-box-shadow``` or ```$input-btn-focus-box-shadow;``` bootstrap global scss vars and get around the `!important`. But if this is a risky move or doesn't fix all cases I think this is ok. It's more the principle than any likely case where we'll need to override :D

rroppolo (approved):
Sorry I missed this one!

scss/buttons.scss
jasonbasuil: I agree on the principle part for sure. Can revisit if needed, but for all cases moving forward there shouldn't be a need to override if this is the standard for focus.

PR #603, opened 2022-05-02
removing TrackedButton from export

### LINKED ISSUE #599: Deprecate TrackedButton from DS

jasonbasuil:
closes #599 

Removing `TrackedButton` from DS exports. It's still used in `CopyToClipboardButton` but have some questions about its use in there (might need to rework it a bit). But for now, will just make sure we are only using `TrackedButton` from `common/analytics` on rails-server.


rroppolo (approved):
Thanks!

PR #605, opened 2022-05-03
add aria-live to Toast

### LINKED ISSUE #604: Add aria-live to Toast messages
https://w3c.github.io/aria-practices/#alert

- The Toast message should provide important and potentially time-sensitive information without interfering with the user's ability to continue working

- Does not affect keyboard focus. This is what Modals are for.

- Avoid designing alerts that disappear automatically (WCAG 2.2.3). Sufficient technique: G5: Allowing users to complete an activity without any time limit

jasonbasuil:
closes #604 

Adds `aria-live="polite"` to the container that holds the `Toast` which will be read aloud to a screen reader in a non-disruptive way. We received feedback for the P messenger that at least having a screen reader say "Message sent" instead of the entire message list would be better. 

https://user-images.githubusercontent.com/37383785/166520070-863cb1d8-5c48-432f-8478-7201caadb03d.mov

### DIFF
--- a/spec/__snapshots__/Storyshots.test.js.snap
+++ b/spec/__snapshots__/Storyshots.test.js.snap
@@ -18072,6 +18072,7 @@ exports[`Storyshots Components/Toast Manual Dismiss Toast 1`] = `
   }
 >
   <div
+    aria-live="polite"
     className="Toast"
   />
   <div>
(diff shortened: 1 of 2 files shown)


wenzeng4 (approved):

rroppolo (approved):
🎉

PR #607, opened 2022-05-03
handle setting title prop on Toast

### LINKED ISSUE #606: Handle title in Toast message setup

jasonbasuil:
closes #606 

We previously weren't able to explicitly set the `title` prop for a `Toast` and often devs were finding it difficult to get this set properly (and sometimes were diverging from our defined styles). Hopefully adding this will make it easier. 

Something as simple as creating this:

image

Also, changed the `<p>` tags into `<div>` on the `Alert` component which should fix some of the issues explained here: https://userinterviews.slack.com/archives/CJYUKSM5M/p1650942546277709

Also just noticing now this margin override was made in `app/javascript/researcher/email_themes/builder/builder.jsx` and would have to be removed since the Alert message would change from a `<p>` to a `<div>`.

```
<Alert
  message={(
    <div style={{marginTop: '-1rem'}}>
      Please note that this preview is approximate.
      Make sure you test this theme with your preferred email client.
      For more tips about styling emails, you can check out our&nbsp;
      <a href={Routes.email_theme_support_path()}>support page.</a>
    </div>
  )}
  type={MessageTypes.INFO}
/>
```

### DIFF
--- a/src/Alert/Alert.jsx
+++ b/src/Alert/Alert.jsx
@@ -80,12 +80,12 @@ function Alert(props) {
       <div className="Alert__content">
         {
           props.title && (
-            <p className="Alert__title">
+            <div className="Alert__title">
               {props.title}
-            </p>
+            </div>
           )
         }
-        <p className="Alert__message">{props.message}</p>
+        <div className="Alert__message">{props.message}</div>
       </div>
       {
         props.action && (
--- a/src/Toast/useToast.js
+++ b/src/Toast/useToast.js
@@ -1,11 +1,12 @@
 import { useCallback, useReducer } from 'react';
 import { v4 as generateUUID } from 'uuid';
 
-const createMessage = (messageType, messageText, messageAction) => ({
+const createMessage = (messageType, messageText, messageAction, messageTitle) => ({
   id: generateUUID(),
   message: messageText,
   type: messageType,
   action: messageAction,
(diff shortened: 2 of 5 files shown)


src/Toast/useToast.js
```diff
@@ -37,8 +43,13 @@ const useToast = (initialMessages = []) => {
     dispatch({ type: ACTIONS.CLEAR_MESSAGES });
   }, []);
 
-  const setMessage = useCallback((messageAction, messageType, messageText) => {
-    dispatch({ type: ACTIONS.SET_MESSAGE, payload: { messageAction, messageType, messageText } });
+  const setMessage = useCallback((messageAction, messageType, messageText, messageTitle) => {
```
jeffbax, line 46: lint?

src/Alert/Alert.jsx
```diff
@@ -80,12 +80,12 @@ function Alert(props) {
       <div className="Alert__content">
         {
           props.title && (
-            <p className="Alert__title">
+            <div className="Alert__title">
```
jeffbax, line 83: I know it is easier to style the div, although semantically it makes me a little 😢 when this isn't a formal header, though I guess we don't really have a great hierarchy to use (h6 probably)

jeffbax (approved):

PR #616, opened 2022-05-12
add fieldset and legend to FormGroup

### LINKED ISSUE #615: Provide fieldset and legend to FormGroup

jasonbasuil:
closes #615 

Addresses Atlassian 1.11 - Question/Answer input groups are missing `<fieldset>` `<legend>` groupings.
Screen reader does not speak the question text when the user tabs to an input in the group.

Wrap the question and answers in a `<fieldset>` and wrap the question text in a `<legend>` element which is the first child of the `<fieldset>`.

[Grouping Controls- Accessibility Quick Guide](http://pauljadam.com/guides/grouping.html)

Make sure the screen reader speaks the question text when tabbing to an input in the group.

Should translate for example into the Survey portion of the Hub P signup flow. Questions should be read aloud and described as (group) as you tab through:

https://user-images.githubusercontent.com/37383785/168127496-43df8571-cb56-4caa-9487-e7e5e6a37a30.mov

### DIFF
--- a/src/InputLabel/InputLabel.jsx
+++ b/src/InputLabel/InputLabel.jsx
@@ -5,10 +5,11 @@ import classNames from 'classnames';
 import Tooltip from 'src/Tooltip';
 
 import 'scss/forms/input_label.scss';
+import './InputLabel.scss';
 
 export default function InputLabel(props) {
   return (
-    <label
+    <legend
       className={classNames('InputLabel', props.className)}
       htmlFor={props.labelHtmlFor}
     >
@@ -16,7 +17,7 @@ export default function InputLabel(props) {
       {props.required && <span className="InputLabel__helper-text">&nbsp;(Required)</span>}
       {props.labelHelperText && <span className="InputLabel__helper-text">&nbsp;({props.labelHelperText})</span>}
       {props.tooltipText && <Tooltip iconClasses="Tooltip__icon--gray" placement="right" text={props.tooltipText} />}
-    </label>
+    </legend>
   );
 }
 
--- a/spec/__snapshots__/Storyshots.test.js.snap
+++ b/spec/__snapshots__/Storyshots.test.js.snap
@@ -3826,16 +3826,16 @@ exports[`Storyshots Components/Form Default 1`] = `
     method="POST"
     onSubmit={[Function]}
   >
-    <div
+    <fieldset
       className="FormGroup"
       id="first-name"
     >
-      <label
+      <legend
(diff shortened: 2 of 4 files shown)


SirSnacksalot (approved):
Looks legit to me! 🏄🏼‍♂️  Thanks Jason

src/InputLabel/InputLabel.jsx
```diff
@@ -5,18 +5,19 @@ import classNames from 'classnames';
 import Tooltip from 'src/Tooltip';
 
 import 'scss/forms/input_label.scss';
+import './InputLabel.scss';
 
 export default function InputLabel(props) {
   return (
-    <label
+    <legend
```
rroppolo, line 12: Hmm I think this might be an issue.. For example, we have some props coming in here like `labelHtmlFor` that link the label and the corresponding input field. Since this component is shared across radio/checkboxes but also regular `input`s as well. I also am worried this will break a lot of our react-testing-library tests in the app since we do a lot of `findByLabelText`, and here we are removing the label and replacing with a legend. Is there any way to achieve the same with labels when dealing with checkbox/radios? It seems odd to me to need to rely on fieldset/legend for form elements since there are form-related native html fields like label and input...

rroppolo: Looking closer I think the issue may be more with how our components are shared across regular input and checkbox/radios

rroppolo: Looking at Carbon it looks like their `FormGroup` is a little different than ours (it encloses all form fields, not a single field like we do) But I did note that they have a separate `RadioButtonGroup` component that creates the fieldset/legend https://react.carbondesignsystem.com/?path=/docs/components-formgroup--default (if you click the Show Code there).

Maybe that is something we can introduce, and then InputLabel can take an `elementType` prop that defaults to `label` but can be overridden to `legend` ? And maybe update our existing `RadioButtonGroup`/`CheckboxButtonGroup` components to include the label instead of wrapping everything in a `FormGroup` for radio/checkboxes?

This may be a pretty big cutover so we could also create new components, gradually switch them over, and then deprecate the old..

jasonbasuil: Thanks for looking into this! It did seem a little too simple to just switch out for fieldset and legend and the potential issues it might cause. I think the recommendation of introducing some new components might be the right path forward and would agree that the cutover might be big. Will explore a little bit more!

PR #617, opened 2022-05-12
fix styles on Select to allow for tabbing and proper focus

### LINKED ISSUE #614: Fix focus accessibility on Select

jasonbasuil:
closes #614 

I think some parts of the custom `styles.js` for `option` was breaking the ability to tab through options and focus on them on our Select components. You should be able to tab through and the screen reader should be able to read through them all.

https://user-images.githubusercontent.com/37383785/168151426-e4cbffef-4f42-4519-a584-31c6c828bcb4.mov

### DIFF
--- a/src/Select/styles.js
+++ b/src/Select/styles.js
@@ -86,17 +86,12 @@ const defaultStyles = ({ size }) => ({
       fontWeight: fontWeights.light,
     }),
     option: (styles, {
-      data,
       isDisabled,
-      isFocused,
       isSelected,
-    }) => {
-      const colors = data.colors || {};
-
-      return {
+    }) => ({
         ...styles,
-        backgroundColor: isSelected || isFocused ? colors.hover : styles.backgroundColor,
-        color: colors.text,
+        backgroundColor: isSelected ? systemColors.UX_BLUE_200 : styles.backgroundColor,
+        color: systemColors.UX_BLACK,
         fontWeight: fontWeights.light,
         fontSize: '0.875rem',
         cursor: 'pointer',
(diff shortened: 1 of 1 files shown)


src/Select/styles.js
```diff
@@ -86,17 +86,12 @@ const defaultStyles = ({ size }) => ({
...
       isDisabled,
-      isFocused,
       isSelected,
-    }) => {
-      const colors = data.colors || {};
-
-      return {
+    }) => ({
         ...styles,
-        backgroundColor: isSelected || isFocused ? colors.hover : styles.backgroundColor,
```
jasonbasuil, line 98: seems like the `colors` was breaking the tabbing ability and was returning `undefined`. Also not sure what we're using `data` for, so I removed it.

jeffbax: "Also not sure what we're using data for, so I removed it."

Is probably as good as any reason to pause a bit... I am pretty sure that you will at least break labels in hub with these changes, where the label is injected into the select as labels have their own color

jasonbasuil: Ah forgot about that. Are you referring to the `LabelSelect` that's on the nav in Hub which doesn't use the DS currently?

jeffbax: ultimately not 100% sure what is and isn't using this, but the code looked familiar

I'd expect we would want the hub component to use this DS component at some point though, so the ability to define a color for the item would seem useful? Maybe they'll be separate components though

jeffbax: Maybe that could be a part of this -- have that use this DS component? Seems like debt we need covered and then you have the real use case to test

jasonbasuil: I think at some point yes. We're creating some custom `Option` and `MultiValue` components in `LabelSelect` and I might want to wait for more design clarity on what we expect? But agreed it should be transitioned to DS. Also we have 3 different labels with varying border-radius and padding which isn't great. 

@katiebratt for visibility 

Screen Shot 2022-05-12 at 2 12 12 PM

Screen Shot 2022-05-12 at 2 12 00 PM

Screen Shot 2022-05-12 at 2 11 32 PM

jeffbax (approved):
seems good as long as we aren't using this for the labels on hub, but definitely a use case for providing colors

PR #621, opened 2022-05-17
allow setting id on InputLabel

### LINKED ISSUE #620: allow setting id on InputLabel

jasonbasuil:
closes #620 

I was running into an issue with setting an `id` on an InputLabel and realized that wasn't possible since we didn't spread props on the component.

Mostly doing this in rare cases where you're using `InputLabel` by itself and some input element and need to set the `id` on it if another element is using `aria-labelledby` and needs it.

For context, working on `PhoneNumber` component which attempts to set id on `InputLabel` and causes some sad a11y error. Tested adding `id` in screenshot below for CountryCode to confirm. Sorry it's a bit hard to read.

Screen Shot 2022-05-17 at 1 39 15 PM

### DIFF
--- a/src/InputLabel/InputLabel.jsx
+++ b/src/InputLabel/InputLabel.jsx
@@ -6,19 +6,26 @@ import Tooltip from 'src/Tooltip';
 
 import 'scss/forms/input_label.scss';
 
-export default function InputLabel(props) {
-  return (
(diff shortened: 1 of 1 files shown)


rroppolo (approved):
LGTM! 👍

PR #632, opened 2022-05-25
update FormGroup and InputLabel to support fieldset and legend

### LINKED ISSUE #631: update InputLabel to support legend

jbasuil-ut:
closes #631

Continuing from my initial [PR-616](https://github.com/user-interviews/ui-design-system/pull/616), I think this could possibly be one / easiest? path forward to supporting `<fieldset>` and `<legend>` with our existing FormGroup with an initial cutover in order to be screen reader accessible on some key flows that need it.

- Adds `inputType` prop on `<FormGroup>` that is set to a string of 'radio' or 'checkbox'. If `inputType` == 'radio' or 'checkbox', the `<FormGroup>` is rendered as a `<fieldset>` otherwise it renders our normal `<div>`
- That `inputType` gets passed down to the `<InputLabel>` which will determine whether the default `<label>` is changed to a `<legend>`.

I tested it out locally on RS and it seems to work okay if given the correct `inputType` on a given `<FormGroup>`. Also checked after running our tests and there doesn't seem to be a difference / any breaking tests as it relates to `getByLabelText` ... at least what I saw when `label` gets switched out for legend (or the tests were not relevant to these changes currently). My thinking is that when the switch does happen (only when we decide to set `inputType`) and it does happen to break a test, I think we could just change `getByLabelText` to something like `getByRole("group", { name: "Some legend that acts as label for radio or checkbox groups" }));` 

Example on `app/javascript/researcher/surveys/survey_form/survey_question_form_group.jsx` when applied:

https://user-images.githubusercontent.com/37383785/170340225-20792243-091e-4bc4-905b-c1d3b7ea35c7.mov

 Possible iteration? -> Updating `RadioButtonGroup` and `CheckboxButtonGroup` to handle the setting `legend` instead of the `FormGroup`?

### DIFF
--- a/src/FormGroup/FormGroup.jsx
+++ b/src/FormGroup/FormGroup.jsx
@@ -1,12 +1,16 @@
-import React from 'react';
+import React, { createElement } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 
 import InputLabel from 'src/InputLabel';
+import InputLegend from 'src/InputLegend';
 
 import 'scss/forms/form_group.scss';
 
-const FORM_GROUP_INPUT_TYPES = ['input', 'radio', 'checkbox'];
+// For accessibility purposes, FormGroups with radio or checkbox groups
+// should be rendered as a <fieldset> element.
+// All other FormGroups will be rendered as a normal <div> by default.
+const FORM_GROUP_ELEMENT_TYPES = ['div', 'fieldset'];
 
 function renderErrors(errors) {
   if (typeof errors === 'string') {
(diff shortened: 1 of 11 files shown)


src/FormGroup/FormGroup.jsx
```diff
@@ -50,9 +73,16 @@ export default function FormGroup(props) {
       )}
       id={props.id}
     >
+      {renderFormGroupChildren()}
+    </div>
+  );
+
+  const renderFormGroupChildren = () => (
```
rroppolo, line 80:
```suggestion
  const formGroupChildren = (
```

Since the functional component is executed on every render there's not really a need for this to be inside a function. It could just be assigned to a const. Or if it starts to get more complicated, broken out into a child component

src/FormGroup/FormGroup.jsx
```diff
@@ -79,8 +109,13 @@ export default function FormGroup(props) {
...
         )
       }
-    </div>
+    </>
   );
+
+  if (isCheckboxOrRadioInputType) {
+    return renderFieldset();
+  }
+    return renderDiv();
```
rroppolo, line 118: I though we might have this in the react style guide but I can't find it :(  But technically the use of `renderFoo` functions inside a component are an older style pattern in our codebase which was deprecated in favor of splitting things out into child components to break up a large render method. But wondering here if we can remove some of the duplication through with something like

```
   return createElement(
     elementType,
      {
        className: classNames(
          'FormGroup',
          props.className, {
            'FormGroup--is-invalid': hasErrors,
            'FormGroup--bordered': props.bordered,
            'FormGroup--inline': props.inline,
          },
        ),
        id: props.id
      },
      formGroupChildren,
    )
```

We do this in some other places like `Card` where we take in an `elementType` prop to make the root element type dynamic.

src/FormGroup/FormGroup.jsx
```diff
@@ -6,6 +6,8 @@ import InputLabel from 'src/InputLabel';
 
 import 'scss/forms/form_group.scss';
 
+const FORM_GROUP_INPUT_TYPES = ['input', 'radio', 'checkbox'];
```
rroppolo, line 9: I'm a little worried there may be a lot of other input types not represented here, like selects or datepickers. Wondering if we would want to make this prop a boolean and more indicative that it is controlling the `fieldset`? Like `elementType="fieldset"`? We could always make a `RadioFormGroup` component that acts as a wrapper that passes the correct `elementType` through to `FormGroup` like
```
const RadioFormGroup = (props) => (
  <FormGroup 
    elementType="fieldset"
    inputElementType="legend"
    {...props} 
  />
);
```

if we wanted something a bit more semantic that is still easy to selectively roll out. Then the consumer just needs to switch `FormGroup` to `RadioFormGroup`/`CheckboxFormGroup` and everything else stays the same?

src/InputLabel/InputLabel.scss
```diff
@@ -0,0 +1,5 @@
+legend {
```
rsaris, line 1: Do we need to scope this to the specific element we're working on or is this something that should be in a generic style override?

As this is written now -- it will apply to every `legend` in the app, which may not be correct.

src/FormGroup/FormGroup.jsx
rsaris: I generally agree w/ this proposal (and it could be used in `InputLabel` as well though I do wonder if that would do better w/ a Boolean (or maybe we make an `InputLabel` / `InputLegend` which we use well here.

rroppolo (approved):
LGTM! Thanks for making that change! 💯 🌮 🙌

PR #647, opened 2022-06-13
Fix cypress tests

brianCollinsUI:
This fixes some broken tests due to class name changes and also makes new rgb function work with current node version.


jasonbasuil (approved):
Ah! Thank you for fixing these! 🙌

PR #652, opened 2022-06-28
Add data-testid to LoadingOverlay

careful-axe:
Allows us to target the `LoadingOverlay` in our tests without class name workarounds.

Closes https://github.com/orgs/user-interviews/projects/3/views/17?filterQuery=test


john-gaskin (approved):
Ah! It seems so obvious now that I'm seeing it here 😆

careful-axe:
> Ah! It seems so obvious now that I'm seeing it here 😆

No more find classes and randomly use the first one. Hurray! 🎉

PR #656, opened 2022-06-29
Add Accordion to DS

### LINKED ISSUE #336: Add Accordion to DS

jasonbasuil:
closes #336

Figma: https://www.figma.com/file/471iS3QrpPtXZpS1xbsJ20/Components?node-id=8191%3A41114

Screen Shot 2022-07-01 at 1 13 08 PM

https://user-images.githubusercontent.com/37383785/176963589-1c5114ca-6883-43e6-9d36-a7ed868416b1.mov


src/Accordion/Accordion.stories.jsx
```diff
@@ -0,0 +1,243 @@
...
+export const Default = () => (
+  <Accordion>
+    <AccordionItem>
+      <AccordionToggle
+        eventKey="0"
+        helperText="helper text"
+        label="3 Items Selected"
+        leadingIcon={faCreditCard}
+        title="Accordion Toggle #1"
+      />
```
AlexAlt, line 36: sorry if this is a dumb question- is the idea here that you will need to import and then render all three of these when building a new Accordion in the rails-server app?

jasonbasuil: That's the idea. Are you talking about importing `Accordion` `AccordionItem` `AccordionCollapse` and `AccordionToggle` ?

jasonbasuil: Or talking about the props highlighted here in order to build the `AccordionToggle`?

AlexAlt: sorry i edited before seeing your comments- yes i mean the former!

jasonbasuil: Got it. Then yes. Hoping that it makes the Accordion a bit more composable and easier to figure out how it's built when it's subcomponents are separated a bit more.

SydneyCEmerson:
Hi @jasonbasuil, I hope I'm leaving a comment in the right place here.  I noticed a few things from the video that I thought were worth bringing up.  Please see below and let me know if you have any questions!  

1. Are you showing the focus state only in the video? I assume the blue outline will only be there if the user is interacting with the screen on a keyboard?  
2. When you expand the accordion, it looks like there is more padding on the bottom card than there should be (see image below.) 
3. It looks like the text and icons don't quite share the same centerline in the video (see image below.)
 4. I noticed that the chevrons animate and rotate 180 degrees when the accordions expand. I like how it looks, but noticed that the chevron icons in our current product just flip positions rather than animating. I'm not sure if we want to change the animation or be consistent how things currently are in the product...something to think about.
 
Comments on accordian

jasonbasuil:
> Hi @jasonbasuil, I hope I'm leaving a comment in the right place here. I noticed a few things from the video that I thought were worth bringing up. Please see below and let me know if you have any questions!
> 
> 1. Are you showing the focus state only in the video? I assume the blue outline will only be there if the user is interacting with the screen on a keyboard?
> 2. When you expand the accordion, it looks like there is more padding on the bottom card than there should be (see image below.)
> 3. It looks like the text and icons don't quite share the same centerline in the video (see image below.)
> 4. I noticed that the chevrons animate and rotate 180 degrees when the accordions expand. I like how it looks, but noticed that the chevron icons in our current product just flip positions rather than animating. I'm not sure if we want to change the animation or be consistent how things currently are in the product...something to think about.
> 
> Comments on accordian

@SydneyCEmerson 
1. Yes the blue focus outline will only be there when the user is interacting or tabbing through.
2. There is a 16px `padding` on the bottom container. Perhaps it's just that the example content (Item 1, Item 2... bullets) is taking up extra space?
3. It looks like it's aligned? Maybe the chevron could be adjusted slightly down?
Screen Shot 2022-07-05 at 12 49 39 PM
 4. I personally love little micro animations to make the interaction feel less stiff / jerky. Welcome to differing opinions tho!

SydneyCEmerson:
@jasonbasuil, see my responses below in **bold**:

1. Yes the blue focus outline will only be there when the user is interacting or tabbing through.  **Sounds good!**
2. There is a 16px `padding` on the bottom container. Perhaps it's just that the example content (Item 1, Item 2... bullets) is taking up extra space?  **Ok, weird.  Maybe there's just an extra line of empty text in the content?** 
> 3. It looks like it's aligned? Maybe the chevron could be adjusted slightly down? **I apologize if this it too nit picky (still figuring out the level of detail of feedback I should be giving here,) but the text on the left, and the card icon on the left look like they could be moved down just a tad in order to be centered. The chevron is more noticeable and yah you're right, it looks like it should be adjusted down so it's centered within the card.**
> 4. I personally love little micro animations to make the interaction feel less stiff / jerky. Welcome to differing opinions tho!  **I totally agree with you - the micro animation looks great and makes the interaction feel better.  My only concern is that I think ideally we would want the interactions for the same components to be as consistant as we can across the product.  I honestly don't know how big of a deal it is though, and I do prefer the interaction you show.  Let me know if it'd help for me to pose this to the other designers to see what they think.**

jasonbasuil:
> @jasonbasuil, see my responses below in **bold**:
> 
> 1. Yes the blue focus outline will only be there when the user is interacting or tabbing through.  **Sounds good!**
> 2. There is a 16px `padding` on the bottom container. Perhaps it's just that the example content (Item 1, Item 2... bullets) is taking up extra space?  **Ok, weird.  Maybe there's just an extra line of empty text in the content?**
> 
> > 3. It looks like it's aligned? Maybe the chevron could be adjusted slightly down? **I apologize if this it too nit picky (still figuring out the level of detail of feedback I should be giving here,) but the text on the left, and the card icon on the left look like they could be moved down just a tad in order to be centered. The chevron is more noticeable and yah you're right, it looks like it should be adjusted down so it's centered within the card.**
> > 4. I personally love little micro animations to make the interaction feel less stiff / jerky. Welcome to differing opinions tho!  **I totally agree with you - the micro animation looks great and makes the interaction feel better.  My only concern is that I think ideally we would want the interactions for the same components to be as consistant as we can across the product.  I honestly don't know how big of a deal it is though, and I do prefer the interaction you show.  Let me know if it'd help for me to pose this to the other designers to see what they think.**

@SydneyCEmerson 
2. Yeah pretty sure it's just a content size thing that can be managed separately
3. I can make a small adjustment to the chevron to align better. Not too picky at all! 💅 Details are important! 
4. Agree on the consistency part. It's mostly just a matter of transitioning old components to new ones, but as long as we have consensus on what we would want ideally it should be good.

SydneyCEmerson:
@jasonbasuil cool, this all sounds good!  Let me know if I can do anything else to help with reviewing the component.

AlexAlt (approved):
generally this looks ok to me- will have to try it out in the rails-server app to really get an idea of how it works!

PR #672, opened 2022-07-20
add ToggleInput

### LINKED ISSUE #671: Add Toggle to DS

jasonbasuil:
closes #671 

Bringing the `ToggleInput` into the Design System.

Nat is currently doing a design review:
https://www.chromatic.com/build?appId=62d040e741710e4f085e0647&number=33


brianCollinsUI (approved):
Should we add a "on" story with `isChecked={true}` so we can track our color visually w/ Chromatic? (I guess also off / disabled, and on / disabled as well, but it starts to feel like overkill). Just not sure where to draw the line with visual test coverage vs making too many stories for storybook interface, so your call on that. 

Also a 'labelLeft' story would be nice just to have that easily discoverable, but also not required.

PR #673, opened 2022-07-20
updates to Accordion v2

### LINKED ISSUE #670: Update Accordion background

jasonbasuil:
closes #670 

Some updates to the Accordion
- added story for wrapping Accordion in Card
- update borderless prop

*some weirdness with bootstrap overrides with `flush` prop when used on rails-server. Added `borderless` to handle our use case and works fine on `rails-server`. Just using `flush` here to showcase correctly on Storybook.

Build 23
https://www.chromatic.com/build?appId=62d040e741710e4f085e0647&number=23


brianCollinsUI (approved):
I was about to spin this up locally and then saw the Chromatic link- looks great!

PR #679, opened 2022-07-25
Fix tests for new paths

brianCollinsUI:
Screen Shot 2022-07-25 at 6 22 37 AM

### DIFF
--- a/cypress/integration/accordion_spec.js
+++ b/cypress/integration/accordion_spec.js
@@ -18,8 +18,14 @@ const accordions = [
     match: 'Accordion Toggle #1',
   },
   {
-    name: 'Flush',
-    path: 'components-accordion--flush',
+    name: 'Borderless',
+    path: 'components-accordion--borderless',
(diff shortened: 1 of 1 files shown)


brianCollinsUI:
Not sure we even want Percy any longer since we are using Chromatic. So let me know once we have made that decision / bought a plan etc

jasonbasuil:
> Not sure we even want Percy any longer since we are using Chromatic. So let me know once we have made that decision / bought a plan etc

@brianCollinsUI I think I would be for just using Chromatic. Would be nice to not have to write out these tests every time we create new stories. I do have a PR #681 that changes the build cadence to nightly just like Percy.

brianCollinsUI:
OK - sounds good. We should just delete this stuff then- will make another branch for that in a bit if that is ok

jasonbasuil:
> OK - sounds good. We should just delete this stuff then- will make another branch for that in a bit if that is ok

Sounds good. Appreciate it Brian! ⭐

PR #696, opened 2022-08-01
[UIDS-695] Make labelTooltip prop type node

### LINKED ISSUE #695: Proptypes warning in project builder
It looks like the `labelTooltip` prop has a situation in app where it needs to be of type node so this just removes the warning we are getting in the console associated

michaeldewolf85:
Closes #695 

It turns out there's a situation where we need type node in the `labelTooltip` prop. See warning in builder. I didn't add it :(
image


jasonbasuil (approved):
Thanks for spotting!

PR #703, opened 2022-08-03
custom DropdownCard

### LINKED ISSUE #702: DropdownMenu custom variant POC

jasonbasuil:
closes #702 

- Create a `DropdownCard` subcomponent of the `Dropdown`. This is a custom use case (potentially for [Hub dropdown filters](https://www.figma.com/file/4dZK7ZVFYWN5d26p0nm1JI/Hub---Table-UX-and-Segments?node-id=2019%3A75606)). 
- The `DropdownCard` has the same sizes as the `Card` component. One limitation is that when viewing the `DropdownCard` on mobile, we won't be able to center it properly. I assume for the first use case we would only be using this component variant on desktop for Hub. 

Chromatic: https://62d040e741710e4f085e0647-cmarbdvfhi.chromatic.com/?path=/story/components-dropdown--dropdown-menu-card

* if you need the invite link to Chromatic, reach out to me and I can send it to you.

### DIFF
--- a/src/Dropdown/DropdownCard.jsx
+++ b/src/Dropdown/DropdownCard.jsx
@@ -2,16 +2,20 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 
+import { CardSizes } from 'src/Card';
+
 import './DropdownCard.scss';
 
 const DropdownCard = ({
   className,
   children,
+  size,
   title,
  }) => (
    <div className={classNames(
       className,
       'DropdownCard',
+      `DropdownCard--${size}`,
       )}
    >
      { title && <h2 className="DropdownCard__title">{title}</h2> }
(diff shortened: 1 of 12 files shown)


src/Dropdown/DropdownCard.jsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+import { CardSizes } from 'src/Card';
+
+import './DropdownCard.scss';
+
+const DropdownCard = ({
```
rroppolo, line 9: I've been thinking about the naming here and what makes something a Card, and whether that's what this is. Since I don't think I've seen a DropdownCard elsewhere, so it may not be a familiar concept. When I look at the design as a container for filters, it made me wonder if this is actually a Popover.  https://carbondesignsystem.com/components/popover/usage/#live-demo

Technically boostrap has a separate component for popover as well https://react-bootstrap.github.io/components/overlays/#popovers

The styling and positioning is a little different, but the advantage of the Popover is it has the option to reposition itself based on where the most available space is. It's hard to tell from the designs, but do you think this would make sense as a Popover? Or is there something specific that makes it a Card? (I think of a card as more resembling its paper counterpart, with the drop shadow.)

rroppolo (commented):
This looks good! I had a question on naming and whether it would make sense to bring in react-bootstraps Popover. Definitely want to hear your thoughts though b/c I'm coming in with less context on this use case so I could be wrong

PR #708, opened 2022-08-12
Accordion updates for better content flexibility and alignment with Card

### LINKED ISSUE #687: Accordion V3 updates for padding

jasonbasuil:
closes #687 

I finally got around to finding a better solution for updating Accordion component to make it more flexible and also align better with the Card component (when multiple cards are stacked on top of each other). A few things to note:

- I changed the way the layout of the component works internally. Previously, I used a flexbox layout for left-content (left chevron, icon, title, helper text), and right-content (label, right-chevron) and using `justifyContent: space-between`.  I realized that it wasn't super flexible if a dev needed to customize a little bit. I opted to use a grid layout instead with just two containers: content (auto) and right-chevron (designated a column area of 1.5rem). It should make the component more flexible and IMO will lead to less dev headache if we need to do some customization and still have the necessary default props.
- I added an example of an Accordion in a Card so that the content aligns better.

https://62d040e741710e4f085e0647-pulbljqitf.chromatic.com/?path=/story/components-accordion--in-card


src/Accordion/AccordionCollapse.scss
```diff
@@ -5,5 +5,9 @@
 
   &__container {
     padding: 1rem;
+
+    &--card-padding {
+      padding: 1rem 1.5rem;
```
jasonbasuil, line 10: only need left and right padding to be adjusted

src/Accordion/AccordionToggle.scss
```diff
@@ -5,23 +5,25 @@
   width: 100%;
 
   &__container {
-    display: flex;
-    justify-content: space-between;
-   
+    display: grid;
+    grid-template-columns: auto 1.5rem;
+    grid-template-areas: 'content chevron-right';
```
jasonbasuil, line 10: I think just having two grid template areas will be easier to manage instead of the 3 containers used with flex. By having just `content` and `chevron-right`, it will be easier (if needed) to layout the main content area in the AccordionToggle but still have the props (leadingIcon, title, helperText, chevronLeft) to fall into place.

src/Accordion/AccordionToggle.jsx
```diff
@@ -11,14 +11,14 @@ import './AccordionToggle.scss';
 import { faChevronUp } from '@fortawesome/pro-solid-svg-icons';
 
 const AccordionToggle = ({
+  cardPadding,
```
rroppolo, line 14: This definitely works but I wonder if this might be a particularly good use case to use styled-components. We do technically have styled-components available in this repo and use it in our Tabs component. Something like:

```
export const CardAccordionToggle = styled(AccordionToggle)`
  padding: 1rem 1.5rem;
`;
```

Or something like that (I could be wrong on the specific syntax). The thing that I like about it is it doesn't require us to include Card-specific props in the base-level `AccordionToggle` component, which could potentially be confusing to see out of context. That's how I got down this path.. because I was scrutinizing the name `cardPadding` as I thought it might be confusing to see "card" there without any additional context. And the the potentially usage is more compact and maybe easier to remember:
```
<CardAccordion>
   <CardAccordionToggle>
   <CardAccordionCollapse>
</CardAccordion>
```

vs. trying to remember to set both `noPadding` and `cardPadding` in the right combination. Some food for thought. I know we haven't leaned hard into styled components yet but in this instance I do feel like it does give some advantages over introducing new props that control css changes. (It still protects the component from the end-user making overrides if we only use styled-components within the context of the design system)

Curious your thoughts!

jasonbasuil:
@rroppolo Nat and I decided that it probably is easiest to manage both the default Accordion and the Accordion in a Card by just adjusting the padding to `padding: 1rem 1.5rem` so that we don't need to use the `cardPadding` prop in both components. 

Also trying out the `UNSAFE_className` in these components.

rroppolo (approved):
Cool that simplifies things a lot!

PR #713, opened 2022-08-16
add disabled ToggleInput story

### LINKED ISSUE #712: add disabled ToggleInput story

jasonbasuil:
closes #712 

wasn't exposing this story, so putting it here.

https://www.chromatic.com/build?appId=62d040e741710e4f085e0647&number=56

### DIFF
--- a/src/ToggleInput/ToggleInput.jsx
+++ b/src/ToggleInput/ToggleInput.jsx
@@ -16,13 +16,13 @@ const ToggleInput = ({
   name,
   onToggle,
   // eslint-disable-next-line camelcase
-  UNSAFE__className,
+  UNSAFE_className,
   ...props
 }) => (
   <label
     className={classNames(
       'ToggleInput__label',
-      UNSAFE__className,
+      UNSAFE_className,
       { 'ToggleInput__label--disabled': disabled },
     )}
     htmlFor={id}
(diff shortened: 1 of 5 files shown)


rroppolo (approved):
👍

src/ToggleInput/ToggleInput.jsx
```diff
@@ -14,9 +15,18 @@ const ToggleInput = ({
   labelText,
   name,
   onToggle,
+  // eslint-disable-next-line camelcase
+  UNSAFE__className,
```
jasonbasuil, line 19: doesn't have to be a part of this PR but just wanted to test the waters with this idea of making className overrides a bit harder to make if necessary. If so, you might do something like this, but could still target it through a parent or something. 

Thoughts @rroppolo ?

Screen Shot 2022-08-16 at 2 41 43 PM

jasonbasuil: doesn't look like any usages of this component are currently using the className escape hatch on rails-server so this might be easy to try?

jasonbasuil: Bump

rroppolo: Sorry for the delay reply here Jason! I am good with testing out this paradigm prop here. I like that it is more easily greppable to eventually be converted to custom DS component wrappers once more established patterns evolve. Also a fan of increasing the barrier to overriding styles (although it's still possible to set `.ParentClass .ToggleInput` but hopefully this being available will lead to parent classes overriding less frequently too)

rroppolo (approved):

PR #717, opened 2022-08-24
add Popover (proof of concept)

### LINKED ISSUE #714: Popover POC

jasonbasuil:
closes #714

### DIFF
--- a/src/Popover/PopoverBody.jsx
+++ b/src/Popover/PopoverBody.jsx
@@ -1,0 +1,11 @@
+import React from 'react';
+
+import { Popover as RBPopover } from 'react-bootstrap';
+
+const PopoverBody = ({ children, props }) => (
+  <RBPopover.Body {...props}>
+    { children }
+  </RBPopover.Body>
+  );
+
+export default PopoverBody;
--- a/src/Button/Button.jsx
+++ b/src/Button/Button.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { forwardRef } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@user-interviews/ui-design-system",
   "version": "1.34.2",
   "dependencies": {
-    "react-bootstrap": "^2.0.2",
+    "react-bootstrap": "^2.5.0",
```
jasonbasuil, line 5: `2.5.0` came with some fixes to the positioning of `Overlay`. Seen in the beta release here: https://github.com/react-bootstrap/react-bootstrap/releases/tag/v2.5.0-beta.0

scss/cards.scss
```diff
@@ -0,0 +1,7 @@
+$card-xs-spacing: 1rem;
```
jasonbasuil, line 1: thought it might be a good idea to move these so that they're available elsewhere

scss/z_stack.scss
```diff
@@ -9,6 +9,7 @@
 // Note look to update `z_stack.js` when changing these
 
 // By Level
+$z-index-1050: 1050;
```
jasonbasuil, line 12: I think this was okay to just set a new z-index variable based on what other elements we're using seen below. This is needed other wise it'll look like this in some settings:

Screen Shot 2022-09-07 at 11 54 24 AM

src/Popover/PopoverBody.jsx
```diff
@@ -0,0 +1,40 @@
...
+import PropTypes from 'prop-types';
+
+import { Popover as RBPopover } from 'react-bootstrap';
+
+import './PopoverBody.scss';
+
+const PopoverBody = ({
+  children,
+  className,
+  variant,
```
careful-axe, line 12: I know it is so tiny only one variant but I do a love a publicly available array to choose this variant from similar to what you did for `OVERLAY_TRIGGER_PLACEMENT` just a 💅 thing though...

careful-axe (approved):
Looks great Jason... have a few thoughts but are all 💅

PR #722, opened 2022-08-26
Add label for Pill close button

brianCollinsUI:
This fixes cypress-axe tests in RS that are failing when looking at pills.

### DIFF
--- a/spec/__snapshots__/Storyshots.test.js.snap
+++ b/spec/__snapshots__/Storyshots.test.js.snap
@@ -8025,6 +8025,7 @@ exports[`Storyshots Components/Pill With Close 1`] = `
   >
     Text
     <button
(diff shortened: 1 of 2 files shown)


spec/__snapshots__/Storyshots.test.js.snap
```diff
@@ -8025,6 +8025,7 @@ exports[`Storyshots Components/Pill With Close 1`] = `
   >
     Text
     <button
+      aria-label="Remove undefined"
```
brianCollinsUI, line 8028: Not sure why this is undefined here- Should be `Remove Text`
Screen Shot 2022-08-26 at 1 36 27 PM

jasonbasuil (approved):
👍

PR #724, opened 2022-08-26
remove prop spread on Accordion and ToggleInput

### LINKED ISSUE #723: Remove prop spread on Accordion and ToggleInput

jasonbasuil:
closes #723 

Also exposes additional props instead of prop spread.

As much as I like the flexibility, it seems like prop spread should be used sparingly according to the [React docs](https://reactjs.org/docs/jsx-in-depth.html#spread-attributes). 

`"Spread attributes can be useful but they also make it easy to pass unnecessary props to components that don’t care about them or to pass invalid HTML attributes to the DOM. We recommend using this syntax sparingly." `

### DIFF
--- a/src/Accordion/Accordion.jsx
+++ b/src/Accordion/Accordion.jsx
@@ -5,25 +5,65 @@ import PropTypes from 'prop-types';
 import RBAccordion from 'react-bootstrap/Accordion';
 
 const Accordion = ({
+  activeKey,
+  alwaysOpen,
+  as,
+  defaultActiveKey,
   children,
+  flush,
+  onSelect,
   // eslint-disable-next-line camelcase
   UNSAFE_className,
-  ...props
 }) => (
   <RBAccordion
+    activeKey={activeKey}
+    alwaysOpen={alwaysOpen}
+    as={as}
     className={classNames(UNSAFE_className, 'Accordion')}
-    {...props}
+    defaultActiveKey={defaultActiveKey}
+    flush={flush}
+    onSelect={onSelect}
   >
     { children }
   </RBAccordion>
   );
 
 Accordion.propTypes = {
+  /**
+    The current active key that corresponds to the currently expanded accordion
+  */
+  activeKey: PropTypes.oneOfType([PropTypes.string, PropTypes.array]),
+  /**
+    Allow accordion items to stay open when another item is opened
+  */
+  alwaysOpen: PropTypes.bool,
+  /**
+    Sets a custom element for this component
+  */
+  as: PropTypes.elementType,
+  /**
+    The default active key that is expanded on start
+  */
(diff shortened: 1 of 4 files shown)


src/Accordion/Accordion.jsx
```diff
@@ -5,25 +5,65 @@ import PropTypes from 'prop-types';
...
+  defaultActiveKey: PropTypes.oneOfType([PropTypes.string, PropTypes.array]),
+  /**
+    Renders accordion edge-to-edge with its parent container
+  */
+  flush: PropTypes.bool,
   UNSAFE_className: PropTypes.string,
+  /**
+    Callback fired when the active item changes
+  */
+  onSelect: PropTypes.func,
```
AlexAlt, line 56: this documentation is tight

src/Accordion/Accordion.jsx
```diff
@@ -5,25 +5,65 @@ import PropTypes from 'prop-types';
...
 };
 
 Accordion.defaultProps = {
+  activeKey: undefined,
+  alwaysOpen: undefined,
```
AlexAlt, line 61: should the default here be true or false? same question for the other booleans. Will any of this get messed up if it's undefined?

src/Accordion/AccordionItem.jsx
```diff
@@ -7,30 +7,38 @@ import RBAccordionItem from 'react-bootstrap/AccordionItem';
 import './AccordionItem.scss';
 
 const AccordionItem = ({
+  as,
```
AlexAlt, line 10: I think this name should be longer and more descriptive of what it is, typically unless you are in like a loop or something I would avoid doing these tiny abbreviated variable names

src/Accordion/Accordion.jsx
jasonbasuil: 'react-bootstrap' doesn't have the defaults for these so I assume it's fine and it highly dependent on use case so makes sense that there is none

src/Accordion/AccordionItem.jsx
jasonbasuil: I kind of just followed the 'react-bootstrap' source code and it seems like they do this for all the components. Granted, I don't think we've even used this prop before to change the element. I was basically just taking all the available props and exposing them, but am open to removing if we might not end up needing it. https://github.com/react-bootstrap/react-bootstrap/blob/master/src/Accordion.tsx

AlexAlt: ah ok- i see where they include it also to match what is in bootstrap, that's fine.

src/Accordion/Accordion.jsx
AlexAlt: ahh ok

AlexAlt (approved):
this looks good to me! feels safer also

PR #727, opened 2022-08-30
update Option to allow for indeterminate CheckboxButton

### LINKED ISSUE #726: Allow for indeterminate checkbox on Select option

jasonbasuil:
closes #726 

You should be able to set an `indeterminate` state on the `CheckboxButton` that is used in the custom `Option` component. 

Screen Shot 2022-08-30 at 2 58 42 PM

Context exploration for an upcoming effort on Hub pod (this effort hasn't started yet, just fyi):
image

### DIFF
--- a/src/Select/SingleSelect.stories.jsx
+++ b/src/Select/SingleSelect.stories.jsx
@@ -1,4 +1,4 @@
-import React, { Fragment } from 'react';
+import React, { Fragment, useRef } from 'react';
 import { action } from '@storybook/addon-actions';
 
 import {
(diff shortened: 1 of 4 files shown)


careful-axe (approved):
Looks great to me and well documented! Thanks for doing this Jason.

src/Select/SingleSelect.stories.jsx
```diff
@@ -117,6 +117,48 @@ export const CustomOptionWithCheckbox = () => (
...
+    { label: 'Patty Participant', value: 2 },
+    { label: 'Patrick Participant (indeterminate)', value: 3 },
+    { label: 'Polly Participant (indeterminate)', value: 4 },
+  ];
+
+  let inputRef;
+
+  const createInputRef = () => {
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    inputRef = useRef();
```
rroppolo, line 132: I know this is just an example but as the [eslint rule](https://reactjs.org/docs/hooks-rules.html) is warning you need to call hooks at the top level to ensure they always get called in the same order. Could this do
```
const inputRef = useRef();
```
and
```
ref={inputRef}
```
without the need for the function?

rroppolo (approved):
I had one small comment on the story but this LGTM!

src/Select/SingleSelect.stories.jsx
jasonbasuil: I think I did give that a try at first, but I think doing so only created a single ref that only applied to the very last Option to make it indeterminate. But yeah I think there should better way to do handle multiple refs but didn't figure it out here and hopefully can be implemented later.

Also just noting that I hadn't handled the logic for clicking an option and it handling the `isSelected` or `!isSelected` state from props, so I'm hoping that can be handled later as well 🤔  

https://user-images.githubusercontent.com/37383785/187558137-6ce0ed21-6eb6-47d2-bdc0-fb0cf1c04a71.mov

wenzeng4 (approved):

PR #739, opened 2022-09-12
Fix color system JS variable order

### LINKED ISSUE #738: Fix color system order
Green and navy both have 500 values that are darker than they should be. It looks like the color values are correct but out of order. Will confirm with the component library and fix the order.

image

image

gabescarbrough:
closes #738

Green and navy both have 500 values that are darker than they should be currently. This is because the non-numbered variables were used but those don't always correspond to the 500 value. I set them all to the 500 variables for consistency.

image

image


jasonbasuil (approved):
LGTM!

PR #743, opened 2022-09-14
make modal prop work with all Selects

### LINKED ISSUE #742: fix modal prop AsyncSelect

jasonbasuil:
closes #742 

Our `modal` prop wasn't working with all of our Selects. It works for our `SingleSelect` which we fixed a while ago when we noticed that the ternary order in the `styles` prop was backwards, so just fixing it with all the other selects we have. Should be like this: 

```
styles={{
  ...defaultStyles({ size }),
  menuPortal: (base) => (
    modal ?
    { ...base, zIndex: zStack.zIndexModalBackdrop + 1 } :
      base
  ),
}}
```

https://user-images.githubusercontent.com/37383785/190243609-69fe3eba-2d0e-49f2-9a23-f425767e69a2.mov

### DIFF
--- a/src/Select/CreatableSelect.jsx
+++ b/src/Select/CreatableSelect.jsx
@@ -2,6 +2,8 @@ import React from 'react';
 import Creatable from 'react-select/creatable';
 import PropTypes from 'prop-types';
 
+import zStack from 'src/Styles/zStack';
+
 import { defaultTheme, defaultStyles, SELECT_SIZES } from './styles';
 
 const CreatableSelect = ({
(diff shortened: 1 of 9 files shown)


src/Select/CreatableSelect.jsx
```diff
@@ -2,6 +2,8 @@ import React from 'react';
 import Creatable from 'react-select/creatable';
 import PropTypes from 'prop-types';
 
+import zStack from 'src/Styles/zStack';
```
rsaris, line 5: Was this just broken before? Where was it pulling `zStack` from?

src/Select/SingleSelect.stories.jsx
```diff
@@ -92,9 +92,11 @@ export const InModal = () => (
         onChange={onChange}
       />
     </ModalBody>
-    <ModalFooter>
-      <button className="btn btn-transparent" type="button">Cancel</button>
-      <button className="btn btn-success" type="submit">Save</button>
+    <ModalFooter
```
rsaris, line 95: Nice clean up for consistency!

rsaris (approved):

domschab23 (approved):
Makes sense; thanks for taking a look at this @jasonbasuil !

src/Select/CreatableSelect.jsx
jasonbasuil: Probably and we just didn't notice it until we tested out the modal prop

PR #750, opened 2022-09-29
add Grid components (Container, Row, Col)

### LINKED ISSUE #749: Add Grid Container

jasonbasuil:
closes #749 

This PR brings in the Grid layout components from React Bootstrap. A few devs have asked for some basic layout components and thought it would be valuable to have. Hopefully it should be intuitive enough to align and manage basic content across all breakpoints, especially as we try to create better experiences on mobile screens.

Chromatic build: https://62d040e741710e4f085e0647-bcwpacclug.chromatic.com/?path=/story/layouts-container--default
(ask me for the join link if you don't already have access to Chromatic)

### DIFF
--- a/src/Card/Card.jsx
+++ b/src/Card/Card.jsx
@@ -44,7 +44,7 @@ const Card = ({
       ...props,
       className: classNames(
         'Card',
-        `Card--${size}`,
+        { [`Card--${size}`]: size },
         className,
         {
           'Card--divided': divided,
@@ -73,7 +73,7 @@ Card.defaultProps = {
   elementType: 'section',
   helperText: undefined,
   noPadding: false,
-  size: CardSizes.LARGE,
+  size: undefined,
   subTitle: undefined,
   title: undefined,
 };
--- a/src/Card/Card.scss
+++ b/src/Card/Card.scss
@@ -16,11 +16,7 @@ $card-width-xs: 15rem;
   outline-color: $ux-green-500;
   padding: $card-spacing;
 
-  width: calc(100% - #{$card-xs-spacing});
-
-  @include media-breakpoint-up(sm) {
-    width: calc(100% - 2rem);
-  }
+  width: 100%;
 
   &__divider {
     border-top: 1px solid rgba(0,0,0,.1);
--- a/src/Container/Col.jsx
+++ b/src/Container/Col.jsx
@@ -5,6 +5,7 @@ import { Col as ReactBootstrapCol } from 'react-bootstrap';
 const Col = ({
   as,
   children,
+  className,
   lg,
   md,
   sm,
@@ -17,6 +18,7 @@ const Col = ({
   <ReactBootstrapCol
     as={as}
     bsPrefix={bsPrefix}
+    className={className}
     lg={lg}
     md={md}
     sm={sm}
@@ -60,6 +62,7 @@ Col.propTypes = {
    This is an escape hatch for working with heavily customized bootstrap css.
   */
   bsPrefix: PropTypes.string,
+  className: PropTypes.string,
   /**
    The number of columns to span on large devices (≥992px)
    `boolean | "auto" | number | { span: boolean | "auto" | number,
@@ -101,6 +104,7 @@ Col.propTypes = {
 Col.defaultProps = {
   as: undefined,
   bsPrefix: 'col',
+  className: undefined,
   lg: undefined,
   md: undefined,
   sm: undefined,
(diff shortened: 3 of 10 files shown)


src/Card/Card.scss
```diff
@@ -16,11 +16,7 @@ $card-width-xs: 15rem;
   outline-color: $ux-green-500;
   padding: $card-spacing;
 
-  width: calc(100% - #{$card-xs-spacing});
-
-  @include media-breakpoint-up(sm) {
-    width: calc(100% - 2rem);
```
jasonbasuil, line 22: I'm trying to think through how much of an impact it would be to remove this calc and just set `width: 100%` since I know people have been having trouble aligning elements outside of Card and need to offset the margins. My hope is that when we layout any base components, we don't have to worry about where a component lays itself and we can just expect that they will fall into place and be edge to edge. 

Going along the principle of [Components should probably not position themselves](https://www.gabe.pizza/notes-on-component-libraries/#:~:text=Components%20should%20probably%20not%20position%20themselves.) and [Components should usually take up all horizontal space given](https://www.gabe.pizza/notes-on-component-libraries/#:~:text=Components%20should%20usually%20take%20up%20all%20horizontal%20space%20given.)

src/Card/Card.jsx
```diff
@@ -44,7 +44,7 @@ const Card = ({
       ...props,
       className: classNames(
         'Card',
-        `Card--${size}`,
+        { [`Card--${size}`]: size },
```
jasonbasuil, line 47: If Card is going to take up the full width of it's container, I suppose that we don't need to give one of the card sizes by default?

src/Card/Card.scss
```diff
@@ -16,11 +16,7 @@ $card-width-xs: 15rem;
...
-  }
+  width: 100%;
```
AlexAlt, line 19: reminder (and i am so sorry) when we add this to the app, we should double check the apply page I was just working on because I have some calculations similar to this on the sibling components. And we should check the accordions probably

src/Container/Col.jsx
```diff
@@ -0,0 +1,116 @@
...
+  >
+    { children }
+  </ReactBootstrapCol>
+  );
+
+const colSize = PropTypes.oneOfType([
+  PropTypes.bool,
+  PropTypes.number,
+  PropTypes.string,
+  PropTypes.oneOf(['auto']),
```
AlexAlt, line 38: ❓ what is this doing? are there supposed to be multiple types in here or is this the only option?

jasonbasuil:
Going to explore potentially updating the `CardContainer` component or maybe a different layout component that has the intention and guidance of aligning and stacking cards, but also has the flexibility to put other stuff in between. Just thinking about deciding if this is a layout, determining the proper breakpoints and columns. 

https://user-images.githubusercontent.com/37383785/193342821-45ba2597-0447-4b61-a4e0-3e1034ccbf68.mov

example code looks something like this: 
```
 <Container>
    <Row>
      <Col sm={{ span: 8, offset: 2 }} md={{ span: 8, offset: 2 }}>
        <h2>A heading</h2>
        <Card title='Gain command over your panel of thousands (or millions) of users'>
        Customize your brand voice everywhere — emails, sender profiles, landing pages, opt-in forms, and screeners. Your customers will have a top-notch experience every time they interact with a researcher on your team.
        </Card>
        <div style={{textAlign: 'right', margin: '1rem 0'}}>
          <Button variant='primary'>Submit</Button>
        </div>
        <Card title='Card Stack 2'></Card>
        <br/>
        <p>Some arbitrary text</p>
        <p>More text</p>
        <Card title='Card Stack 3'></Card>
      </Col>
    </Row>
  </Container>
```

src/Card/Card.jsx
AlexAlt: so will this be a required prop now? or just be wild depending on the container?

AlexAlt (approved):
seems frickin legit, excited/scared to try this out in the app

rroppolo:
Re: the CardContainer with things other than Cards example: is there a way to set the gap space on the grid so that each item doesn't need to specify its own margin? Curious because margins can be annoying. We have a problem in forms where the form elements have a margin-bottom, but then the last in the list we usually have to remove the margin-bottom. So wondering if our layouts can handle consistent spacing between items w/out margin

jasonbasuil:
> Re: the CardContainer with things other than Cards example: is there a way to set the gap space on the grid so that each item doesn't need to specify its own margin? Curious because margins can be annoying. We have a problem in forms where the form elements have a margin-bottom, but then the last in the list we usually have to remove the margin-bottom. So wondering if our layouts can handle consistent spacing between items w/out margin

@rroppolo I think it's possible to use the utility classes from bootstrap to create gaps on the parent container (i don't think react-bootstrap has built it into their API yet tho). https://getbootstrap.com/docs/5.1/utilities/spacing/#gap Not sure if that's the route we would want to take since I think we're shying away from utility classes, but I've heard cases for and against

PR #756, opened 2022-10-03
change width to max-width on Card

### LINKED ISSUE #755: Change max-width on Card

jasonbasuil:
closes #755 

I think it needs to be `max-width: 100%` instead of `width: 100%` now that `Card` takes up the entire width, but we don't want it to overflow its parent container on mobile. Examples below:

Screen Shot 2022-10-03 at 11 37 01 AM

Screen Shot 2022-10-03 at 11 36 50 AM

Screen Shot 2022-10-03 at 12 05 43 PM

Screen Shot 2022-10-03 at 12 06 01 PM


rroppolo (approved):
👍

PR #761, opened 2022-10-06
Input label tooltip color

kyleshike:
Removes the `-gray` css suffix from the tooltips in `<InputLabel/>` and `<InputLegend/>`


jasonbasuil (approved):
Thanks!

PR #763, opened 2022-10-07
add CardStack and CardStackContainer

### LINKED ISSUE #762: update CardContainer to be flexible

jasonbasuil:
closes #762 

[Chromatic build](https://62d040e741710e4f085e0647-wbkcwiwiih.chromatic.com/?path=/story/layouts-cardstack--default): let me know if you need an invite to view 

Handling some of the flex / spacing issues with CardContainer by making some changes to Card, spacing, and layout in general. As a reminder, we're moving away from [Cards positioning themselves](https://www.gabe.pizza/notes-on-component-libraries/#:~:text=Components%20should%20probably%20not%20position%20themselves.) and moving responsibility to containers.

Introduces:

`<CardStack>` which is a layout container for vertical stacks of Cards. It sets vertical margins on Cards only. Hope to deprecate `CardContainer` with this.

- CardStack can also be responsible for Card sizing via the `size` prop (same as Card)
- It doesn't horizontally align cards on its own. You'll need a separate container (was thinking of bringing a general use Flex component for this later)

~~`<CardStackContainer>` is a different way to horizontally center Cards using our new grid components (`Container`, `Row`, `Col`) under the hood. It is a responsive layout where you can use any of the available breakpoints (xs, sm, md ... etc) to set column widths. (not too sure yet on guidelines for design on when to use vs. our fixed width Cards we're getting from `CardStack` now, but it's here anyways)~~

Cutover to rails-server notes:
- Replace `CardContainer` with `CardStack`.
- Will need ensure `CardStack` has a container with `display: flex; justify-content: center;`  

Some preliminary testing on RS (ignore weird buttons):

https://user-images.githubusercontent.com/37383785/196770974-9497ff1b-4e4a-4ee8-b921-f3e85931d1c4.mov

https://user-images.githubusercontent.com/37383785/196771913-04bf3be5-c5fd-4d37-88d9-04e485994611.mov

Could get rid of some of the margin/padding overrides we've been doing with CardContainer: 

Screen Shot 2022-10-19 at 11 29 11 AM

Screen Shot 2022-10-19 at 11 28 45 AM


src/CardStack/CardStack.jsx
```diff
@@ -0,0 +1,36 @@
...
+import PropTypes from 'prop-types';
+
+import './CardStack.scss';
+
+const CardStack = ({
+  children,
+  // eslint-disable-next-line camelcase
+  UNSAFE_className,
+  size,
+  ...props
```
jasonbasuil, line 12: one thing I noticed is that if we want to use `UNSAFE_className` to keep track of overrides...in this case seeing if anyone ever needs to adjust the `CardStack` spacing... , we would need to get rid of the `...props` since you can still set `className` if it's there.

src/CardStack/CardStack.jsx
```diff
@@ -0,0 +1,36 @@
...
+}) => (
+  <div
+    className={classNames(
+      UNSAFE_className,
+      'CardStack',
+      { [`CardStack--${size}`]: size },
```
jasonbasuil, line 18: by giving sizing responsibility to `CardStack`.... you can just set the `size` prop on it and then the `Card` will fill the container by default

src/CardStack/CardStack.scss
```diff
@@ -0,0 +1,31 @@
+@import '../../scss/theme';
+
+$card-stack-child-card-vertical-margin: $card-spacing / 2;
+$card-stack-horizontal-padding: $card-spacing / 2;
+
+.CardStack {
+  display: flex;
+  flex-direction: column;
+
+  padding: $card-stack-horizontal-padding;
```
jasonbasuil, line 10: added this padding just to provide spacing on the sides of cards when viewing on mobile

src/CardStack/CardStack.scss
```diff
@@ -0,0 +1,31 @@
+@import '../../scss/theme';
+
+$card-stack-child-card-vertical-margin: $card-spacing / 2;
```
jasonbasuil, line 3: setting only vertical margin on the top and bottom of Card so that if you ever need to put a different element in between cards (e.g. some text or a button or something)... it'll be spaced evenly

src/CardStack/CardStackContainer.jsx
```diff
@@ -0,0 +1,53 @@
...
+  lg,
+  md,
+  xl,
+  xxl,
+  xs,
+  ...props
+}) => (
+  <Container className={className} fluid={fluid} {...props}>
+    <Row>
+      <Col />
```
jasonbasuil, line 22: is this weird to do this? it'll render two empty `<div class="col"></div>`

src/CardStack/CardStack.scss
```diff
@@ -0,0 +1,31 @@
...
+  flex-direction: column;
+
+  padding: $card-stack-horizontal-padding;
+  
+  .Card {
+    margin: $card-stack-child-card-vertical-margin 0;
+  }
+
+  &--xs {
+    width: $card-width-xs
```
kyleshike, line 17: **question:** ❓Should these be `max-width`?

Otherwise I think `CardStack` runs the risk of overflowing it's parent
```suggestion
    width: 100%;
    max-width: $card-width-xs;
```

src/CardStack/CardStack.scss
```diff
@@ -0,0 +1,31 @@
...
+
+.CardStack {
+  display: flex;
+  flex-direction: column;
+
+  padding: $card-stack-horizontal-padding;
+  
+  .Card {
+    margin: $card-stack-child-card-vertical-margin 0;
+  }
```
kyleshike, line 14: **thought:** 💭 Won't this result in extra space inside the bounding box of `.CardStack`?  What if instead, each direct descendant of the class (excluding the first element) gets a margin-top instead? eg:
```suggestion
  > *:not(:first-child) {
    margin: $card-spacing 0 0;
  }
```

What's also kind of cool about this suggested approach is that it then everything child of `CardStack` is spaced correctly regardless of what it is

src/CardStack/CardStack.scss
kyleshike: **question:** ❓ Should this use a breakpoint?

If this padding is to give some horizontal breathing room to the component in mobile specifically, should it only be applied at or below the mobile breakpoint?  Also, unless we need this padding vertically as well, could the padding just be applied horizontally? eg:
```suggestion
  @include media-breakpoint-down(sm) {
      padding: 0 $card-stack-horizontal-padding;
  }
```

kyleshike: Or maybe this is a high-level layout concern that should be addressed in a separate component and possibly used in a number of places? eg: Let `CardStack` take up (at most) the full width of it's parent, and have another component like `<PageLayout/>` that specifies the horizontal padding for all breakpoints.

src/CardStack/CardStackContainer.jsx
kyleshike: It does seem _a little_ odd, especially if the provided size is `12`, in which case we'd end up with empty cols above & below. If the purpose here is to ensure it's centered horizontally, maybe best to just apply that with a stylesheet? _OR_ I know that bootstrap has the utility class `justify-content-center`, maybe that would work here? https://getbootstrap.com/docs/5.0/layout/columns/#horizontal-alignment

src/CardStack/CardStack.jsx
kyleshike: Could we alternatively provide an optional `spacing` prop to the component?  Something like 1 -5 (or whatever) and then have eg:

```scss
.CardStack--spacing-1 {}
.CardStack--spacing-2 {}
// ... etc
```

src/CardStack/CardStackContainer.jsx
jasonbasuil: I think I made this `CardStackContainer` first before introducing the `size` prop to  `CardStack` when all I wanted was just to horizontally align a stack of Cards. And tbh, maybe we might not need to introduce `CardStackContainer` with all its responsive breakpoints when `CardStack` gets the job done and you just need to wrap it in a flex container of some sorts.  

I do think generally we are trying to shy away from using utility classes from bootstrap.

src/CardStack/CardStack.jsx
```diff
@@ -0,0 +1,36 @@
...
+    )}
+    {...props}
+  >
+    {children}
+  </div>
+);
+
+CardStack.propTypes = {
+  size: PropTypes.string,
```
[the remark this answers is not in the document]
jasonbasuil: 👍

src/CardStack/CardStack.scss
jasonbasuil: I think that makes sense. I believe our original `Card` used a breakpoint for this padding purpose on mobile. I don't know quite yet what a higher level layout might look like. I know it's a bit varied with the current page containers we have that might need some consolidation.

src/CardStack/CardStack.scss
jasonbasuil: Ooh I do like this! The spacing looks equal and the bounds of the CardStack are still up against the edge. 
Screen Shot 2022-10-20 at 1 06 46 PM

jasonbasuil: I do imagine for the most common use case of `CardStack` ... it would mostly be Cards. Maybe the occasional heading or button or other element. It falls neatly in place with equal spacing. I imagine there may be times the margin might need adjustment depending on what non-card elements fall into it, but I'm all for making everything equidistant from the get go and then adjusting if needed.

kyleshike: > the margin might need adjustment depending on what non-card elements fall into it, but I'm all for making everything equidistant from the get go and then adjusting if needed

Yea I completely agree --let the most common use case be the easiest to implement while still facilitating the edge/corner cases via classes or props on the children

src/CardStack/CardStack.scss
kyleshike: Since the margins are done with `$card-spacing` can this be removed?
```suggestion
```

src/Container/Col.jsx
```diff
@@ -31,24 +31,24 @@ const Col = ({
   </ReactBootstrapCol>
   );
 
-const colSize = PropTypes.oneOfType([
+export const colSizePropTypes = PropTypes.oneOfType([
```
kyleshike, line 34: I could be wrong, but weren't these changes done in support of importing them in`<CardStackContainer/>`?  Since that component no longer exists, can these changes also be reverted?

kyleshike (approved):
Two small cleanup Q's but otherwise this is awesome. Excited to try this out in rails-server

src/Container/Col.jsx
jasonbasuil: ah yes! Good catch 👍

src/CardStack/CardStack.scss
```diff
@@ -1,6 +1,6 @@
 @import '../../scss/theme';
 
-$card-stack-horizontal-padding: $card-spacing / 2;
+$card-stack-horizontal-padding: .75rem;
```
jasonbasuil, line 3: got this console warning. Will wait till we need to upgrade

```
DEPRECATION WARNING: Using / for division is deprecated and will be removed in Dart Sass 2.0.0.
09:31:14 webpack.1   | 
09:31:14 webpack.1   | Recommendation: math.div($card-spacing, 2)DEPRECATION WARNING: Using / for division is deprecated and will be removed in Dart Sass 2.0.0.
09:31:14 webpack.1   | 
09:31:14 webpack.1   | Recommendation: math.div($card-spacing, 2)
```

PR #772, opened 2022-11-03
remove CardContainer and CardList

### LINKED ISSUE #771: remove CardContainer

jasonbasuil:
closes #771 

`CardContainer` is being removed since we cutover to use `CardStack`
`CardList` preceded CardContainer but was never used so removing that too.


rroppolo (approved):
🎉

PR #774, opened 2022-11-03
add OptionWithDescription for selects

### LINKED ISSUE #773: Add Select option with description

jasonbasuil:
closes #773 

Screen Shot 2022-11-08 at 12 11 35 PM

[Chromatic link](https://62d040e741710e4f085e0647-yrbkxavpbt.chromatic.com/?path=/story/components-selects-single--custom-option-with-description)

- Adds custom option component, `OptionWithDescription` for role picker select.
- Updates `form_group.scss` to account for Select width (which we were doing on rails-server)
- Add `menuWidth` prop for cases where you want to adjust the menu width.


scss/forms/form_group.scss
```diff
@@ -88,4 +88,9 @@
     @include font-type-30;
     height: auto;
   }
+
+  .AsyncSelect, .CreatableSelect, .SingleSelect {
```
jasonbasuil, line 92: looks like we handle this in `ui_form.scss` on rails-server so I think it makes sense moving this over here

src/Select/SingleSelect.jsx
```diff
@@ -22,6 +22,7 @@ const SingleSelect = ({
   isLoading,
   isMulti,
   isSearchable,
+  menuWidth,
   modal,
   name,
   options,
```
domschab23, line 28: It might make sense to add `components` as a prop for all select components that can use `Option` or `OptionWithDescription`

domschab23 (approved):
Makes sense! I was able to test this locally with `SingleSelect` component and custom `menuWidth` prop, and it works as expected. Great job @jasonbasuil!
Screen Shot 2022-11-09 at 1 26 44 PM

src/Select/styles.js
```diff
@@ -56,6 +56,10 @@ const defaultStyles = ({ size }) => ({
       color: systemColors.UX_BLUE_700,
       borderRadius: '.25rem',
     }),
+    menu: (styles) => ({
+      ...styles,
+      width: !menuWidth ? '100%' : menuWidth,
```
rroppolo, line 61: We may just want to do a quick pass over these selects when QAing. I know we had the width: 100% in ui_form but not sure if any of these exist outside a FormGroup that might change with this.

rroppolo (approved):
LGTM!

PR #781, opened 2022-11-15
add Flex component

### LINKED ISSUE #780: Add Flex component

jasonbasuil:
closes #780 

Screen Shot 2022-11-14 at 4 11 24 PM

The `<Flex>` layout component makes use of `styled-components`. I thought it would make more sense to use this instead of the usual classnames convention since there are so many possible flex properties that you can use. This is probably only my second time using styled-components so let me know if there are some best practices or conventions I'm overlooking.

Also I added what I thought were the most commonly used flex properties, but let me know if you want another one added or changed! Play around with the Storybook knobs in the build below:

Chromatic build: https://62d040e741710e4f085e0647-ukafrrrmop.chromatic.com/?path=/story/layouts-flex--default

### DIFF
--- a/src/Flex/Flex.jsx
+++ b/src/Flex/Flex.jsx
@@ -18,9 +18,9 @@ Flex.propTypes = {
     If `true`, `display: flex;` otherwise `display: block;`
   */
   container: PropTypes.bool,
-  direction: PropTypes.oneOf(Object.values(FLEX_PROPS.direction)),
   flex: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
   flexBasis: PropTypes.string,
+  flexDirection: PropTypes.oneOf(Object.values(FLEX_PROPS.flexDirection)),
   flexGrow: PropTypes.number,
   flexShrink: PropTypes.number,
   flexWrap: PropTypes.oneOf(Object.values(FLEX_PROPS.flexWrap)),
@@ -49,9 +49,9 @@ Flex.defaultProps = {
   alignSelf: undefined,
   className: undefined,
   container: undefined,
-  direction: undefined,
   flex: undefined,
   flexBasis: undefined,
+  flexDirection: undefined,
   flexGrow: undefined,
   flexShrink: undefined,
   flexWrap: undefined,
--- a/src/Flex/Flex.jsx
+++ b/src/Flex/Flex.jsx
@@ -25,10 +25,6 @@ Flex.propTypes = {
   flexShrink: PropTypes.number,
   flexWrap: PropTypes.oneOf(Object.values(FLEX_PROPS.flexWrap)),
   /**
-    row-gap column-gap (e.g. '10px 20px' => `gap: 10px 20px;`)
-  */
-  gap: PropTypes.string,
-  /**
     rem or px
   */
   height: PropTypes.string,
@@ -59,7 +55,6 @@ Flex.defaultProps = {
   flexGrow: undefined,
   flexShrink: undefined,
   flexWrap: undefined,
-  gap: undefined,
   height: undefined,
   justifyContent: undefined,
   justifySelf: undefined,
--- a/src/Flex/Flex.jsx
+++ b/src/Flex/Flex.jsx
@@ -1,0 +1,91 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { FlexWrapper } from './FlexWrapper.styles';
+
+const Flex = ({ className, children, ...props }) => (
+  <FlexWrapper className={className} {...props}>
+    {children}
+  </FlexWrapper>
+);
+
+Flex.propTypes = {
+  alignItems: PropTypes.oneOf([
+    'stretch',
+    'center',
+    'flex-start',
+    'flex-end',
+    'baseline',
+    'initial',
+    'inherit',
+  ]),
+  alignSelf: PropTypes.oneOf([
+    'stretch',
+    'center',
+    'start',
+    'end',
+  ]),
+  className: PropTypes.string,
+  /**
+    If `true`, `display: flex;` otherwise `display: block;`
+  */
+  container: PropTypes.bool,
+  direction: PropTypes.oneOf(['column', 'row']),
+  flex: PropTypes.string,
+  flexBasis: PropTypes.string,
+  flexGrow: PropTypes.number,
+  flexShrink: PropTypes.number,
+  flexWrap: PropTypes.oneOf(['wrap', 'nowrap', 'reverse']),
+  /**
+    rem or px
+  */
+  height: PropTypes.string,
+  justifyContent: PropTypes.oneOf([
+    'flex-start',
+    'flex-end',
+    'space-between',
+    'space-around',
+    'center',
+    'initial',
+    'inherit',
+  ]),
+  justifySelf: PropTypes.oneOf([
+    'stretch',
+    'center',
+    'start',
+    'end',
+  ]),
+  /**
+    rem or px
+  */
+  maxHeight: PropTypes.string,
+  /**
+    rem or px
+  */
+  maxWidth: PropTypes.string,
+  /**
+    rem or px
+  */
+  width: PropTypes.string,
+};
+
+Flex.defaultProps = {
+  alignItems: undefined,
+  alignSelf: undefined,
+  className: undefined,
+  container: undefined,
+  direction: undefined,
+  flex: undefined,
+  flexBasis: undefined,
+  flexGrow: undefined,
+  flexShrink: undefined,
+  flexWrap: undefined,
+  height: undefined,
+  justifyContent: undefined,
+  justifySelf: undefined,
+  maxHeight: undefined,
+  maxWidth: undefined,
+  width: undefined,
+};
+
+export default Flex;
--- a/src/Flex/Flex.jsx
+++ b/src/Flex/Flex.jsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
+import { FLEX_PROPS } from './Flex.types';
 import { FlexWrapper } from './FlexWrapper.styles';
 
 const Flex = ({ className, children, ...props }) => (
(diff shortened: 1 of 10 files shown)


src/Flex/Flex.jsx
```diff
@@ -0,0 +1,96 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { FlexWrapper } from './FlexWrapper.styles';
+
```
kyleshike, line 5: **suggestion:** 💭 What do you think about adding exported consts for the flex properties (`alignItems, alignSelf, flexDirection, flexWrap, justifyContent, justifySelf`)?

That way consumers of the component won't have to worry (as much) about the correct syntax & spelling.

```suggestion
export const alignItems = {
  baseline: 'baseline',
  center: 'center',
  ...etc
};

...etc

File: src/Flex/Flex.jsx [hunk 2]
```diff
@@ -0,0 +1,96 @@
...
+    'stretch',
+    'center',
+    'start',
+    'end',
+  ]),
+  className: PropTypes.string,
+  /**
+    If `true`, `display: flex;` otherwise `display: block;`
+  */
+  container: PropTypes.bool,
```
kyleshike, line 32: **question:** ❓I'm a little confused about this prop.  What does it do exactly and how will it get used in practice? In what context would we want to apply flex properties to an element with `display: block`?

src/Flex/Flex.jsx
```diff
@@ -0,0 +1,96 @@
...
+  direction: PropTypes.oneOf(['column', 'row']),
+  flex: PropTypes.string,
```
kyleshike, line 34: **question:** ❓Can't this be a number as well? https://developer.mozilla.org/en-US/docs/Web/CSS/flex#syntax
```suggestion
  flex: PropTypes.oneOfType([PropTypes.number, PropTypes.string])
```

src/Flex/Flex.jsx
```diff
@@ -0,0 +1,96 @@
...
+  flexBasis: PropTypes.string,
+  flexGrow: PropTypes.number,
+  flexShrink: PropTypes.number,
+  flexWrap: PropTypes.oneOf(['wrap', 'nowrap', 'reverse']),
+  /**
+    row-gap column-gap (e.g. '10px 20px' => `gap: 10px 20px;`)
+  */
+  gap: PropTypes.string,
```
kyleshike, line 42: **thought:** 💭 I feel like we should be careful about this one because it's not fully supported yet by Safari https://caniuse.com/flexbox-gap.  Until we can drop support for Safari 14 and iOS Safari 14.4 & under, this won't do what we want... 😢

src/Flex/Flex.jsx
```diff
@@ -0,0 +1,96 @@
...
+    'center',
+    'start',
+    'end',
+  ]),
+  className: PropTypes.string,
+  /**
+    If `true`, `display: flex;` otherwise `display: block;`
+  */
+  container: PropTypes.bool,
+  direction: PropTypes.oneOf(['column', 'row']),
```
kyleshike, line 33: **nitpick:** 💅 Prefix this prop with `flex` for consistency

```suggestion
  flexDirection: PropTypes.oneOf(['column', 'row']),
```

src/Flex/Flex.jsx
```diff
@@ -0,0 +1,96 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { FlexWrapper } from './FlexWrapper.styles';
+
+const Flex = ({ className, children, ...props }) => (
+  <FlexWrapper className={className} {...props}>
```
kyleshike, line 7: **thought:** 🤔 I don't know much about styled components,  but it seems like most if not all of these props (excluding `container`) could be passed directly to an html element's `style` prop, which reduces some complexity here.

eg: 
```suggestion
const Flex = ({ className, children, ...styleProps }) => (
  <div className={className} style={...styleProps}>
```

src/Flex/Flex.stories.jsx
```diff
@@ -0,0 +1,171 @@
...
+  parameters: {
+    docs: {
+      page: mdx,
+    },
+  },
+};
+
+const Box = ({ children }) => (
+  <div style={{
+    height: '100px',
```
kyleshike, line 28: **Suggestion:** 🐛 When the knobs for `height` and `maxHeight` are set to a value less than 100px, these boxes end up overflowing the container

```suggestion
    height: '100%',
```

kyleshike (commented):
I'm so excited for this one!  I left a few comments throughout, happy to chat synchronously about any of them if needed

src/Flex/Flex.jsx
jasonbasuil: Good question. I think I've seen some Flex components in other design systems do this where you won't always need to use `display: flex` as default. Maybe it's just me In my head thinking it's helpful when writing "hey, I'm making this <Flex> as a container... a display: flex container. Parent container. Oh, what properties are available to parents? Oh. What if I want to add non-parent (flex-item) children, but they can just be block elements? What flex-item properties could I still add to them?".

This might be a dumb, not often used example I tried thinking of where you want might want `display: block` on a `<Flex>` with available flex properties:

Screen Shot 2022-11-15 at 12 02 39 PM

 ```
<Flex container direction='row'>
    <Flex flexBasis='300px'>
      <Box>1A</Box>
      <Box>1B</Box>
      <Box>1C</Box>
    </Flex>
    <Flex>
      <Box>2A</Box>
      <Box>2B</Box>
    </Flex>
    <Flex>
      <Box>3A</Box>
    </Flex>
  </Flex>
```

File: src/Flex/Flex.jsx (hunk 4 above)
Comment (jasonbasuil) [reply src/Flex/Flex.jsx:42]: ah no! Good catch thinking about this. Would've been nice to have full support

File: src/Flex/Flex.types.js [hunk 8]
```diff
@@ -0,0 +1,41 @@
...
+    end: 'end',
+  },
+  direction: {
+    column: 'column',
+    row: 'row',
+  },
+  flexWrap: {
+    wrap: 'wrap',
+    noWrap: 'no-wrap',
+    reverse: 'reverse',
```
rroppolo, line 24: Is this meant to be 'wrap-reverse'? Didn't seem like it was honoring `reverse`

src/Flex/Flex.stories.jsx
```diff
@@ -0,0 +1,112 @@
...
+    <h1 style={{ fontSize: '1.5rem', fontWeight: '700' }}>Properties for the children (flex item)</h1>
+    <Flex container>
+      <Flex
+        alignSelf={radios('alignSelf', Object.values(FLEX_PROPS.alignSelf))}
+        flex={text('flex')}
+        flexBasis={text('flexBasis')}
+        flexGrow={text('flexGrow')}
+        flexShrink={text('flexShrink')}
+        height="100px"
+        justifySelf={radios('justifySelf', Object.values(FLEX_PROPS.justifySelf))}
```
rroppolo, line 100: Hmm is there a trick tog getting these knobs to work? I wasn't able to do much with the flex prop since the other two items don't have flex set.. Would it be worth setting flex="1" on the other two in the row? Or maybe have a few choices to select from vs. a free-text field, like how they have here https://developer.mozilla.org/en-US/docs/Web/CSS/flex ?

src/Flex/Flex.mdx
```diff
@@ -0,0 +1,46 @@
+import { ArgsTable, Story, Canvas } from '@storybook/addon-docs/blocks';
+import Flex from './Flex';
+
+# Flex
+
+##
+
+<h4>
+  Flex provides a layout container using CSS Flexbox.
```
rroppolo, line 9: ❓ My only thought is how `Container` sounds so broad, is it confusing when someone should use `Container` and when they should use `Flex`? The description of `Container` also mentions it is built with flexbox. Like would it make sense to call that one a `GridContainer` or something to further distinguish when to use one vs the other? Or something we can flesh out more in the documentation?

src/Flex/FlexWrapper.styles.js
```diff
@@ -0,0 +1,95 @@
+import styled, { css } from 'styled-components';
+
+export const FlexWrapper = styled.div`
```
rroppolo, line 3: Would we ever want to use something other than a div here? I noticed the `Container` component has an `as` prop that lets us control the html node it applies the flexbox styles to. Not totally sure how that would extend to styled components though..

rroppolo (approved):
Had a few comments/questions but I think this makes sense! I wonder if adding clarity between this component and `Container` will help with understanding the different applications, as well as maybe flushing out the stories to give more practical examples. But I think this is a good start!

src/Flex/Flex.jsx
kyleshike: I wonder if my confusion is stemming from the fact that in flexbox, some css properties apply to the parent when display is `flex` or `inline-flex`, and some apply to it's children. For example:

#### parent properties
```jsx
<Flex flexDirection='reverse'>
  <p>one</p>
  <p>two</p>
  <p>three</p>
</Flex>
```
Since the parent does not have `display: flex` it doesn't recognize the `flex-direction` property and the children aren't reversed


#### child properties
```jsx
<Flex>
  <Flex flexBasis="300px">one</Flex>
  <Flex flexBasis="300px">two</Flex>
  <Flex flexBasis="300px">three</Flex>
</Flex>
```

Since the parent doesn't have `display: flex`, the children aren't flex items and won't recognize the `flex-basis` property.

src/Flex/Flex.jsx
```diff
@@ -0,0 +1,66 @@
...
+
+Flex.propTypes = {
+  alignItems: PropTypes.oneOf(Object.values(FLEX_PROPS.alignItems)),
+  alignSelf: PropTypes.oneOf(Object.values(FLEX_PROPS.alignSelf)),
+  className: PropTypes.string,
+  /**
+    If `true`, `display: flex;` otherwise `display: block;`
+  */
+  container: PropTypes.bool,
+  direction: PropTypes.oneOf(Object.values(FLEX_PROPS.direction)),
```
kyleshike, line 21: Looks like this one flew under the radar maybe:

```suggestion
  flexDirection: PropTypes.oneOf(Object.values(FLEX_PROPS.flexDirection)),
```

src/Flex/Flex.types.js
```diff
@@ -0,0 +1,41 @@
...
+  },
+  alignSelf: {
+    stretch: 'stretch',
+    center: 'center',
+    start: 'start',
+    end: 'end',
+  },
+  direction: {
+    column: 'column',
+    row: 'row',
```
kyleshike, line 19: There are a couple other props here we should probably add:
```suggestion
    columnReverse: 'column-reverse',
    row: 'row',
    rowReverse: 'row-reverse',
```

kyleshike (approved):
A few small things but I don't think anything that should block a new release.  Perhaps over time and use in rails-server, we'll get a better sense of how to best represent flex box here in the DS

PR #785, opened 2022-11-22
Add Typography component

### LINKED ISSUE #783: add Typography component

jasonbasuil:
closes #783 

Screen Shot 2022-12-08 at 1 10 25 PM

Screen Shot 2022-12-08 at 1 10 34 PM

Created a `Text` and `Heading` component which helps establish hierarchy and presents content as clearly and efficiently as possible. It also helps create some default mapping of our `font-type` typography mixins to a reusable component as well as provide some best practices and documentation on accessibility & hierarchy.


src/Typography/Typography.stories.jsx
```diff
@@ -0,0 +1,120 @@
...
+          <TableCell><Pill color="blue">{typography.mixin}</Pill></TableCell>
+        </TableRow>
+      )))}
+    </TableBody>
+  </Table>
+);
+
+export const Heading = () => (
+  <>
+    <Typography as="h1" variant="heading3xl">The fastest way to recruit research participants</Typography>
```
jasonbasuil, line 88: I'm curious if these `variant` names I've used make enough sense and will be easy to consume. I took some inspiration from Shopify's Polaris Design System: https://polaris.shopify.com/components/text

src/Typography/Typography.types.js
```diff
@@ -0,0 +1,33 @@
+export const TYPOGRAPHY_PROPS = {
+  as: {
+    h1: 'h1',
```
jasonbasuil, line 3: I feel like the available `as` props here should be sufficient? You can always add any other valid element.

gabescarbrough:
This is just an initial thought but what would you think of breaking this into two components `Heading` and `Text`? I think if we took that route then the props could be cleaner, we wouldn't have to use `as` as much and it feels a bit more abstracted so there is less temptation to just write HTML.

For instance instead of
```jsx
<Typography as="h6" variant="headingXs">The fastest way to recruit research participants</Typography>
<Typography as="p" variant="bodyMd">Source from a pool of more than 2.1 million participants to reach nearly any target audience.</Typography>
```

you could do
```jsx
<Heading level={6} size="xs">The fastest way to recruit research participants</Heading>
<Text size="md">Source from a pool of more than 2.1 million participants to reach nearly any target audience.</Text>
```

You could still have the `as` prop for when you want `Text` to be an `aside` or something but it feels cleaner to me.

I could see some devs seeing the first option and thinking they might as well just do `<h6 class="Typography--headingXs">The fastest way to recruit research participants</h6>` not realizing the benefits of using the component.

Also I'm not sure how many devs know what typography means. Sure they could Google it but `Heading` and `Text` seem more obvious to me.

src/Typography/Typography.stories.jsx
gabescarbrough: I prefer `xxxl` to `3xl` but I don't have a real reason beyond it looks better to me.

jasonbasuil:
> This is just an initial thought but what would you think of breaking this into two components `Heading` and `Text`? I think if we took that route then the props could be cleaner, we wouldn't have to use `as` as much and it feels a bit more abstracted so there is less temptation to just write HTML.
> 
> For instance instead of
> 
> ```
> <Typography as="h6" variant="headingXs">The fastest way to recruit research participants</Typography>
> <Typography as="p" variant="bodyMd">Source from a pool of more than 2.1 million participants to reach nearly any target audience.</Typography>
> ```
> 
> you could do
> 
> ```
> <Heading level={6} size="xs">The fastest way to recruit research participants</Heading>
> <Text size="md">Source from a pool of more than 2.1 million participants to reach nearly any target audience.</Text>
> ```
> 
> You could still have the `as` prop for when you want `Text` to be an `aside` or something but it feels cleaner to me.
> 
> I could see some devs seeing the first option and thinking they might as well just do `<h6 class="Typography--headingXs">The fastest way to recruit research participants</h6>` not realizing the benefits of using the component.
> 
> Also I'm not sure how many devs know what typography means. Sure they could Google it but `Heading` and `Text` seem more obvious to me.

@gabescarbrough I was also thinking about breaking it up into `Heading` and `Text` which does seem to make sense as a consumer. I initially followed what Shopify Polaris did with their `Text` component https://polaris.shopify.com/components/text and did notice that they had deprecated their [Heading](https://polaris.shopify.com/components/heading) , [Subheading](https://polaris.shopify.com/components/subheading), and a couple other text related components in favor of just one broad `Text` component which made me wonder what were the cons to having them all be separated 🤔 

I am also just wanting to make sure that with developer handoff from designers that it is easy enough to look at their mocks from Figma and easily translate the exact font specs to the component w/o having too much guess work or having to refer to documentation. Since we are abstracting away the need to create a stylesheet and set a font mixin like `font-type-50--bold` for example. 

So for example if you were to see a Figma file like this (which isn't exposing what variable font token is being used... though designers said they try to expose the font token as much as possible), is it easy enough to translate to the react component with its props?

Screen Shot 2022-12-07 at 12 19 38 PM

gabescarbrough:
> @gabescarbrough I was also thinking about breaking it up into `Heading` and `Text` which does seem to make sense as a consumer. I initially followed what Shopify Polaris did with their `Text` component https://polaris.shopify.com/components/text and did notice that they had deprecated their [Heading](https://polaris.shopify.com/components/heading) , [Subheading](https://polaris.shopify.com/components/subheading), and a couple other text related components in favor of just one broad `Text` component which made me wonder what were the cons to having them all be separated 🤔

The first thing it makes me think of is that React Native has one base `Text` component so I guess if you're using both React and React Native that is nice. Though you could always write your own `Heading` component for React Native. For what it's worth when I did RN I didn't like using `Text` for both body text and heading. My brain groups h1, h2, h3, etc as variations on the same thing (headings) and p, ol, ul, etc as another group (text content).

I guess the other pro to one component is if they have all the same props. If you implemented it like my example they wouldn't since text elements don't have levels and headings do, but if they did then it would seem like unnecessary duplication to split them. I do think `<Heading level={6}>` feels nicer than `<Typography as="h6">` but that could just be me.

> So for example if you were to see a Figma file like this (which isn't exposing what variable font token is being used... though designers said they try to expose the font token as much as possible), is it easy enough to translate to the react component with its props?

I think so. My inclination would be to open the React Dev Tools and fiddle with the props until the computed CSS lined up with the Figma file.

gabescarbrough:
> @gabescarbrough I was also thinking about breaking it up into Heading and Text which does seem to make sense as a consumer. I initially followed what Shopify Polaris did with their Text component https://polaris.shopify.com/components/text and did notice that they had deprecated their [Heading](https://polaris.shopify.com/components/heading) , [Subheading](https://polaris.shopify.com/components/subheading), and a couple other text related components in favor of just one broad Text component which made me wonder what were the cons to having them all be separated 🤔

The first thing it makes me think of is that React Native has one base Text component so I guess if you're using both React and React Native that is nice. Though you could always write your own Heading component for React Native. For what it's worth when I did RN I didn't like using Text for both body text and heading. My brain groups h1, h2, h3, etc as variations on the same thing (headings) and p, ol, ul, etc as another group (text content).

I guess the other pro to one component is if they have all the same props. If you implemented it like my example they wouldn't since text elements don't have levels and headings do, but if they did then it would seem like unnecessary duplication to split them. I do think <Heading level={6}> feels nicer than <Typography as="h6"> but that could just be me.

> So for example if you were to see a Figma file like this (which isn't exposing what variable font token is being used... though designers said they try to expose the font token as much as possible), is it easy enough to translate to the react component with its props?

I think so. My inclination would be to open the React Dev Tools and fiddle with the props until the computed CSS lined up with the Figma file.

gabescarbrough (approved):
Looks awesome! Thanks for working on this. It will be super helpful.

PR #786, opened 2022-11-22
add labelDescription to OptionWithDescription

### LINKED ISSUE #784: Add label helper to OptionWithDescription

jasonbasuil:
closes #784 

Chromatic: https://62d040e741710e4f085e0647-yopxegfste.chromatic.com/?path=/story/components-selects-single--custom-option-with-description

- adds `labelDescription` in cases where the `label` needs additional context
- uses `hideDescription` in cases where you need to hide the description when viewing on mobile

Screen Shot 2022-11-22 at 12 52 06 PM

### DIFF
--- a/src/Select/OptionWithDescription.jsx
+++ b/src/Select/OptionWithDescription.jsx
@@ -22,9 +22,11 @@ const OptionWithDescription = ({ ...props }) => (
         {props.label}&nbsp;
         {props.data.labelDescription}
       </label>
-      <div className="OptionWithDescription__description">
-        {props.data.description}
-      </div>
+      {!props.data.hideDescription && (
+        <div className="OptionWithDescription__description">
+          {props.data.description}
+        </div>
+      )}
     </div>
   </components.Option>
   );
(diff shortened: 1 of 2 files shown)


src/Select/OptionWithDescription.jsx
```diff
@@ -19,11 +19,14 @@ const OptionWithDescription = ({ ...props }) => (
...
     <div className="OptionWithDescription">
       <label className="OptionWithDescription__label">
-        {props.label}
+        {props.label}&nbsp;
+        {props.data.labelDescription}
       </label>
-      <div className="OptionWithDescription__description">
-        {props.data.description}
-      </div>
+      {!props.data.hideDescription && (
```
domschab23, line 25: Now that I think about, can we make `props.data.hideDescription` be `props.hideDescription`?

Pretty much, I'd like to set it up in this fashion

```    <SingleSelect
      components={{
        Option: props => <OptionWithDescription {...props} hideDescription />,
      }}
```
instead of having `hideDescription` on `data` object (which I think are pretty much `options` from `SingleSelect`, right?)

jasonbasuil: good point!

domschab23 (approved):
Looks good to me! Tested with local DS 

Mobile hides description when using media query hook
Screen Shot 2022-11-22 at 4 31 22 PM

`labelDescription` is being displayed
Screen Shot 2022-11-22 at 4 31 04 PM

PR #800, opened 2022-12-19
add LoadingSkeleton

### LINKED ISSUE #798: add Skeleton loading

jasonbasuil:
closes #798 

[Chromatic link](https://62d040e741710e4f085e0647-aautmrrylg.chromatic.com/?path=/docs/components-loadingskeleton--default#props)

- Add `react-loading-skeleton`
- Create guidelines on when to use `LoadingSkeleton` vs. `LoadingOverlay`

Future: possibly build `LoadingSkeleton` into existing components (e.g ProfileCell, Button, Table) to make it easier to automatically handle skeleton states during loading. Mostly wanting to make it as easy as possible to use LoadingSkeleton without having to [create many dedicated loading skeleton screens.](https://www.npmjs.com/package/react-loading-skeleton#:~:text=Don%27t%20make%20dedicated%20skeleton%20screens)

What it could look like: 

https://user-images.githubusercontent.com/37383785/208517987-2bbd7955-f6bf-420d-b546-0cc2831d1242.mov

Currently:

https://user-images.githubusercontent.com/37383785/208518000-291be00a-1f03-4e84-9385-e5f7506cdca8.mov

### DIFF
--- a/src/LoadingSkeleton/LoadingSkeleton.jsx
+++ b/src/LoadingSkeleton/LoadingSkeleton.jsx
@@ -7,29 +7,10 @@ import 'react-loading-skeleton/dist/skeleton.css';
 
 import colors from '../Styles/colors/palette';
 
-const LoadingSkeleton = ({
- borderRadius,
- circle,
- className,
- containerClassName,
- containerTestId,
- count,
- height,
- inline,
- width,
- ...props
-}) => (
+const LoadingSkeleton = ({ className, ...props }) => (
   <SkeletonTheme baseColor={colors.UX_GRAY_300}>
     <Skeleton
-      borderRadius={borderRadius}
-      circle={circle}
       className={classNames('LoadingSkeleton', className)}
-      containerClassName={containerClassName}
-      containerTestId={containerTestId}
-      count={count}
-      height={height}
-      inline={inline}
-      width={width}
       {...props}
     />
   </SkeletonTheme>
@@ -46,7 +27,7 @@ LoadingSkeleton.propTypes = {
   circle: PropTypes.bool,
   className: PropTypes.string,
   /**
-  A custom class name for the <span> that wraps the individual skeleton elements.
+  A custom class name for the `<span>` that wraps the individual skeleton elements.
   */
   containerClassName: PropTypes.string,
   /**
@@ -64,7 +45,7 @@ LoadingSkeleton.propTypes = {
   */
   height: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
   /**
-  By default, a <br /> is inserted after each skeleton so that each skeleton gets its own line.
+  By default, a `<br />` is inserted after each skeleton so that each skeleton gets its own line.
   When inline is true, no line breaks are inserted.
   */
   inline: PropTypes.bool,
@@ -75,15 +56,15 @@ LoadingSkeleton.propTypes = {
 };
 
 LoadingSkeleton.defaultProps = {
-  borderRadius: undefined,
+  borderRadius: '0.25rem',
   className: undefined,
-  circle: undefined,
+  circle: false,
   containerClassName: undefined,
   containerTestId: undefined,
-  count: undefined,
-  height: 20,
-  inline: undefined,
-  width: undefined,
+  count: 1,
+  height: undefined,
+  inline: false,
+  width: '100%',
 };
 
 export default LoadingSkeleton;
--- a/src/LoadingSkeleton/LoadingSkeleton.jsx
+++ b/src/LoadingSkeleton/LoadingSkeleton.jsx
@@ -8,27 +8,82 @@ import 'react-loading-skeleton/dist/skeleton.css';
 import colors from '../Styles/colors/palette';
 
 const LoadingSkeleton = ({
- className, count, height, ...props
+ borderRadius,
+ circle,
+ className,
+ containerClassName,
+ containerTestId,
+ count,
+ height,
+ inline,
+ width,
+ ...props
 }) => (
   <SkeletonTheme baseColor={colors.UX_GRAY_300}>
     <Skeleton
+      borderRadius={borderRadius}
+      circle={circle}
       className={classNames('LoadingSkeleton', className)}
+      containerClassName={containerClassName}
+      containerTestId={containerTestId}
       count={count}
       height={height}
+      inline={inline}
+      width={width}
       {...props}
     />
   </SkeletonTheme>
 );
 
 LoadingSkeleton.propTypes = {
+  /**
+  The border radius of the skeleton.
+  */
+  borderRadius: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
+  /**
+  Makes the skeleton circular by setting border-radius to 50%.
+  */
+  circle: PropTypes.bool,
   className: PropTypes.string,
+  /**
+  A custom class name for the <span> that wraps the individual skeleton elements.
+  */
+  containerClassName: PropTypes.string,
+  /**
+  A string that is added to the container element as a data-testid attribute.
+  Use it with screen.getByTestId('...') from React Testing Library.
+  */
+  containerTestId: PropTypes.string,
+  /**
+  The number of lines of skeletons to render. If count is a decimal number like 3.5,
+  three full skeletons and one half-width skeleton will be rendered.
+  */
   count: PropTypes.number,
-  height: PropTypes.number,
+  /**
+  The height of the skeleton.
+  */
+  height: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
+  /**
+  By default, a <br /> is inserted after each skeleton so that each skeleton gets its own line.
+  When inline is true, no line breaks are inserted.
+  */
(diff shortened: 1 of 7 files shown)


src/LoadingSkeleton/LoadingSkeleton.jsx
```diff
@@ -0,0 +1,89 @@
...
+  The border radius of the skeleton.
+  */
+  borderRadius: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
+  /**
+  Makes the skeleton circular by setting border-radius to 50%.
+  */
+  circle: PropTypes.bool,
+  className: PropTypes.string,
+  /**
+  A custom class name for the <span> that wraps the individual skeleton elements.
```
kyleshike, line 49: 💅 I'm a little unsure how these comments ultimately get used by Chromatic, but it looks like the `<span>` here is getting interpreted as real html:

Screen Shot 2023-01-04 at 4 10 43 PM



```suggestion
  A custom class name for the `<span>` that wraps the individual skeleton elements.
```

src/LoadingSkeleton/LoadingSkeleton.jsx
```diff
@@ -0,0 +1,89 @@
...
+  The number of lines of skeletons to render. If count is a decimal number like 3.5,
+  three full skeletons and one half-width skeleton will be rendered.
+  */
+  count: PropTypes.number,
+  /**
+  The height of the skeleton.
+  */
+  height: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
+  /**
+  By default, a <br /> is inserted after each skeleton so that each skeleton gets its own line.
```
kyleshike, line 67: 💅 Similar issue here:

Screen Shot 2023-01-04 at 4 11 32 PM

```suggestion
  By default, a `<br />` is inserted after each skeleton so that each skeleton gets its own line.
```

src/LoadingSkeleton/LoadingSkeleton.mdx
```diff
@@ -0,0 +1,77 @@
...
+### When to use 
+- For representing all content on initial page load 
+- For representing large or multiple content elements on a page (avoiding multiple LoadingOverlays at one time)
+- For components that are data-intesive and may take a few seconds to fully load 
+- Offering a simplified preview of loading content
+
+### When not to use
+- Do not use a LoadingSkeleton and LoadingOverlay together. Use the one that best fits the context.
+
+### LoadingSkeleton vs. LoadingOverlay?
```
kyleshike, line 26: 🙌 This is so great. 

Thinking ahead about how this component might interact with RTK-Q's `isLoading` and `isFetching`, it seems like we'll want to use skeletons for `isLoading`, and use `LoadingOverlay` only when _refetching_ (eg: `!isLoading && isFetching`)

src/LoadingSkeleton/LoadingSkeleton.jsx
```diff
@@ -0,0 +1,89 @@
...
+  When inline is true, no line breaks are inserted.
+  */
+  inline: PropTypes.bool,
+  /**
+  The width of the skeleton.
+  */
+  width: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
+};
+
+LoadingSkeleton.defaultProps = {
```
kyleshike, line 77: 🤔  Would it make sense to have parity between the default props of our `<LoadingSkeleton/>` and `<Skeleton/>` from `react-loading-skeleton`?  For example, their default border radius is 0.25rem, whereas this one is `undefined`.

Similar for the `height` prop.  It seems really neat that react loading skeleton's height will inherit from the parent font size by default (or at least that's what the docs seem to indicate). Is there anyway to keep that behavior for ours? I'm thinking of use-cases like the one layed out in their docs [here](https://github.com/dvtng/react-loading-skeleton#adapts-to-the-styles-you-have-defined)

kyleshike (commented):
So excited for this one.  Left a few thoughts regarding the default props, I'm curious what you think

src/LoadingSkeleton/LoadingSkeleton.mdx
gabescarbrough: I don't know if the current `LoadingOverlay` is a good fit for refetching either since it hides the existing content and doesn't give any indication much progress has been made. `LoadingOverlay` doesn't really give you anymore information than `LoadingSkeleton` does so I'm can't think of a case where showing a spinner is actually better.

I think I'd need a better understanding of exactly when `isFetching` happens to have a real opinion of what the best solution is, but existing content disappearing for a loading state that doesn't tell me anything is a pet peeve of mine.

src/LoadingSkeleton/LoadingSkeleton.mdx
```diff
@@ -0,0 +1,77 @@
...
+| Form submission                                                        | No                                                | ✅ Yes (typically full-screen, possibly with page refresh)            |
+| Page saving                                                            | No                                                | ✅ Yes (typically full-screen)                                        |
+| Card                                                                   | ✅ Yes (for Card content, not the container itself| No (unless there is a user action that changes that data in the card) |
+| Table                                                                  | ✅ Yes (for initial page load)                    | No                                                                    |
+| Avatar or ProfileCell                                                  | ✅ Yes                                            | No                                                                    |
+| Alert or Toast                                                         | No                                                | No
+
```
gabescarbrough, line 41: To be play devil's advocate, I don't think the cases for `LoadingOverlay` actually provide any benefit over `LoadingSkeleton`.

The cases where you're saying to use `LoadingOverlay` are user initiated which does make them different than an initial page load, so `LoadingSkeleton` probably isn't the right choice, but I think `LoadingOverlay` is still the wrong solution too.

I'd prefer the component that initiated the action (almost always a button) to show the loading state (preferably as a progress bar but a spinner wouldn't be any worse than it is now) and for the page to still show the old content until the new content is loaded.

This opinion is totally non-blocking for this PR but I would love if we never used `LoadingOverlay`.

src/LoadingSkeleton/LoadingSkeleton.jsx
kyleshike: Yea I believe it would, but the default prop for height here is 20: https://github.com/user-interviews/ui-design-system/blob/53bf52a83400f747edf5ee574695be32b54531f8/src/LoadingSkeleton/LoadingSkeleton.jsx#L77-L84

src/LoadingSkeleton/LoadingSkeleton.mdx
jasonbasuil: Yeah I don't have a perfect solution for this just yet and we're still trying to align with design on what works best in all different situations. This is all documentation I've pulled from other places just to get something out. I do like the progress bar idea tho and am a fan of it.

We have talked about incorporating the loading spinner just inside the Button component. I've seen it in a few places before.

src/LoadingSkeleton/LoadingSkeleton.jsx
kyleshike: Unfortunately I don't think my suggestion did the trick 😢

jasonbasuil: Sorry, I had to build another [Chromatic](https://62d040e741710e4f085e0647-aautmrrylg.chromatic.com/?path=/docs/components-loadingskeleton--default#props) so that was it!

src/LoadingSkeleton/LoadingSkeleton.mdx
kyleshike: From the [docs](https://redux-toolkit.js.org/rtk-query/usage/queries#frequently-used-query-hook-return-values), `isLoading` is true only on the first request, whereas `isFetching` will also be true whenever a cached response is revalidated. This occurs when a mutation invalidates a tag, which is typically initiated by a user's action.

What I'm saying is that we probably don't want to use the skeleton in these cases, and instead provide a different indication the app is processing their request.

src/LoadingSkeleton/LoadingSkeleton.mdx
kyleshike: > incorporating the loading spinner just inside the Button component

Just want to add a ⬆️ to this idea. It's such a great way of providing immediate feedback to the user that their button press, submit action, etc was registered, without significantly changing the overall page render

kyleshike (approved):

src/LoadingSkeleton/LoadingSkeleton.mdx
gabescarbrough: Thanks for the link to the docs.

[This comment](https://github.com/user-interviews/ui-design-system/pull/800#discussion_r1062645926) probably gets across what I was saying better.

Totally agree that user initiated reloading is different from initial load and makes sense to be represented differently. My point was that the main reason `LoadingSkeleton` doesn't make sense for reloading (it hides all the content) is also a problem with `LoadingOverlay`.

Not blocking for this at all. You're right that `LoadingOverlay` is the correct option of the two for reloading that given what we have right now since people associate skeletons with initial load. Just pointing out that `LoadingOverlay` isn't great for reloading either.

PR #806, opened 2023-01-09
import skeleton.css directly

### LINKED ISSUE #805: import skeleton.css directly

jasonbasuil:
closes #805 

There's a conflicting order webpack issue when used on RS that I'm thinking is due to `import 'react-loading-skeleton/dist/skeleton.css';`. I just copied that code and made a separate `LoadingSkeleton.scss` 

Screen Shot 2023-01-09 at 11 49 25 AM

I ran into a similar issue with [ToggleInput](https://github.com/user-interviews/ui-design-system/blob/develop/src/ToggleInput/ToggleInput.scss#L81) a while back where I did the same thing to resolve the issue. I'm hoping this will be the same fix 🤞

### DIFF
--- a/spec/__snapshots__/Storyshots.test.js.snap
+++ b/spec/__snapshots__/Storyshots.test.js.snap
@@ -7679,6 +7679,268 @@ exports[`Storyshots Components/LoadingOverlay Default 1`] = `
 </div>
 `;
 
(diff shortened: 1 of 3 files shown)


kyleshike (approved):
LGTM 🚢 

Kinda neat to see how they styled this thing under the hood

PR #811, opened 2023-01-13
Adding optional props to Tabs component

### LINKED ISSUE #809: Adjust Tabs and Tab components
There are 2 things I'd like to change in `Tabs` and/or `Tab` component

1. No wrapping tab components. Either delete `flex-wrap: wrap` from the css class or have an option not to have
Screen Shot 2023-01-12 at 2 41 01 PM

2. Once the above is fixed, then the border bottom should always go to the lowest one. Something similar to what I showed @jasonbasuil during our call, but I'll take a deeper look!
Screen Shot 2023-01-12 at 2 42 51 PM

domschab23:
Fix #809

Current state and without `flexWrapUnset` nor `navItemButtonFullHeight`

https://user-images.githubusercontent.com/54962889/212375185-2120164d-b375-4c07-8ffb-bc8fc8c96fbc.mov

With `flexWrapUnset` but without `navItemButtonFullHeight`

https://user-images.githubusercontent.com/54962889/212375298-69fc33ab-e0f7-4b4e-822e-57568e303494.mov

With `flexWrapUnset` and `navItemButtonFullHeight`

https://user-images.githubusercontent.com/54962889/212375375-16d6fa38-6dd0-475d-85ff-c463e6184752.mov

I think there's a case to be made that both of these css changes should be by default added to this component without those props, but it's up to @jasonbasuil to decide

### DIFF
--- a/spec/__snapshots__/Storyshots.test.js.snap
+++ b/spec/__snapshots__/Storyshots.test.js.snap
@@ -20626,7 +20626,7 @@ exports[`Storyshots Components/Tabs Controlled 1`] = `
 >
   <div>
     <span
-      className="sc-dkPtRN klomrB"
+      className="sc-dkPtRN gfzzIq"
     >
       <ul
         className="nav nav-tabs"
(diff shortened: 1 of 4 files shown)


brianCollinsUI (approved):
**praise**: 🙌

Nice - thanks for sharing the videos of this behavior- looks great!

jasonbasuil (approved):
Thanks @domschab23 ! I think this works for your case and we can decide on defaults once we get a bit more design clarity. 👍

PR #829, opened 2023-02-06
Bigger text for SingleSelect

brianCollinsUI:
Added a note to readme about first time setup

Screen Shot 2023-02-06 at 6 24 02 AM

Screen Shot 2023-02-06 at 6 24 08 AM


jasonbasuil:
Hey @brianCollinsUI ! I chatted with the designers this morning and there was a general consensus that we will hold off on the bigger text for the SingleSelect for now (@SydneyCEmerson said we're good) but can revisit if we decide that it would be more applicable in other places. Thanks for taking the time to explore this!

brianCollinsUI:
Great! thanks for the update

PR #838, opened 2023-02-09
Update setMessage to accept object in useToast.js and add Toast.mdx doc file

### LINKED ISSUE #836: Update setMessage to accept object in useToast.js and update Toast docs
## Problem
_What is the problem we are trying to solve?_

The expected order of parameters for incoming messages in `useToast.js` gets scrambled in the flow from the dispatch of `setMessage` —> 
https://github.com/user-interviews/ui-design-system/blob/c8b2d3864565c78adb39fbf9ffa1a5590bfdb1be/src/Toast/useToast.js#L46

to the invocation of createMessage in the `messagesReducer`  —>
https://github.com/user-interviews/ui-design-system/blob/c8b2d3864565c78adb39fbf9ffa1a5590bfdb1be/src/Toast/useToast.js#L23-L28

to `createMessage`'s definition where the parameters are ordered differently  —> 
https://github.com/user-interviews/ui-design-system/blob/c8b2d3864565c78adb39fbf9ffa1a5590bfdb1be/src/Toast/useToast.js#L4

It goes from `action, type, text, title` in the first two examples, to `type, text, action, title` in `createMessage`'s definition. 

## Strategy
_What is the general approach we are taking to solve this problem?_

Update `setMessage` and `createMessage` to accept an object with `message`, `title`, `action`, and `type` keys and become order agnostic.

gretahayes19:
Resolves #836 

Update `setMessage` in `useToast.jsx`  to expect an object with `message`, `title`, `action`, and `type` keys and become order agnostic.

Add Toast.mdx with Props table and `useToast` documentation:
Screen Shot 2023-02-08 at 4 20 42 PM

### DIFF
--- a/src/Toast/Toast.jsx
+++ b/src/Toast/Toast.jsx
@@ -43,6 +43,10 @@ Toast.propTypes = {
   header: PropTypes.bool,
   messages: PropTypes.arrayOf(
     PropTypes.shape({
+      action: PropTypes.shape({
+        content: PropTypes.string,
+        url: PropTypes.string,
+      }),
       id: PropTypes.string,
       message: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
       title: PropTypes.string,
--- a/src/Toast/Toast.mdx
+++ b/src/Toast/Toast.mdx
@@ -6,8 +6,12 @@ import Toast from './Toast';
 ##
 
 <h4>
-	Toasts communciate important messages and provide immediate feedback without disrupting the user experience.
+	Toasts communicate important messages and provide immediate feedback without disrupting the user experience.
 </h4>
+
+<Canvas>
+  <Story id="components-toast--default" />
+</Canvas>
 
 ## Props
 
@@ -22,7 +26,7 @@ This custom hook returns the following four values to aid the implementation of 
 - Array of objects with shape:
 ```
 {
-  type: 'success' (one of MessageTypes),
+  type: MessageTypes.SUCCESS, (one of MessageTypes)
   title: 'A good title',
   message: 'A good message',
   action: {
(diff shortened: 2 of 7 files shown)


src/Toast/Toast.jsx
```diff
@@ -43,6 +43,10 @@ Toast.propTypes = {
   header: PropTypes.bool,
   messages: PropTypes.arrayOf(
     PropTypes.shape({
+      action: PropTypes.shape({
```
jasonbasuil, line 46: nice!

src/Toast/Toast.mdx
```diff
@@ -0,0 +1,53 @@
+import { ArgsTable, Story, Canvas } from '@storybook/addon-docs/blocks';
+import Toast from './Toast';
+
+# Toast
+
+##
+
+<h4>
+	Toasts communciate important messages and provide immediate feedback without disrupting the user experience.
```
jasonbasuil, line 9: typo on communicate

src/Toast/Toast.mdx
```diff
@@ -0,0 +1,53 @@
...
+</h4>
+
```
jasonbasuil, line 11: might be good to add in the first Toast story here after the title description just as an example 

Something like:
```
<Canvas>
  <Story id="components-toast--toast" />
</Canvas>
```

And actually, it might be good to rename the Story to "Default" just to follow how we also setup stories for other components just to show what the default component looks like.

src/Toast/Toast.mdx
```diff
@@ -0,0 +1,53 @@
...
+## useToast
+
+This custom hook returns the following four values to aid the implementation of toast components.
+
+#### messages
+- Array of objects with shape:
+```
+{
+  type: 'success' (one of MessageTypes),
```
jasonbasuil, line 25: can we say `type: MessageTypes.SUCCESS` just to encourage people to use the MessageTypes instead of the string

jasonbasuil (approved):
I think this is really great and will make it much easier to set toast messages! I gave it a quick spin locally and we definitely will need to make updates on RS to accept the new object format. But other than that, I just made some small doc recommendations. LGTM 🙌

PR #845, opened 2023-02-18
Add bugsnag config to DS and handle toast arg error

### LINKED ISSUE #844: Add error handling in useToast setMessage
Per [this comment](https://github.com/user-interviews/rails-server/pull/16379#discussion_r1108633316) add error handling to handle incorrect/old arg format for `setMessage`

```javascript
function setMessage(...args) {
  if(args && typeof args[0] === 'string') {
    console.error("hey don't do it this way anymore");
    // do it the old way anyway
  }
  // do it the new way
}
```

gretahayes19:
Resolves #844

Add `bugsnag` configuration to DS in order to handle toast arg error. 

If users make a function call to `useToast` with the incorrect arg signature, it will trigger a bugsnag error report.  Eventually we hope to remove this error handling once we are confident we have correctly updated all use cases within `rails-server`.

### DIFF
--- a/src/Toast/__snapshots__/withToast.test.jsx.snap
+++ b/src/Toast/__snapshots__/withToast.test.jsx.snap
@@ -13,7 +13,7 @@ Array [
       >
         <svg
           aria-hidden="true"
-          className="svg-inline--fa fa-circle fa-w-16 "
+          className="svg-inline--fa fa-circle "
           data-icon="circle"
           data-prefix="fas"
           focusable="false"
@@ -45,7 +45,7 @@ Array [
         </svg>
         <svg
           aria-hidden="true"
-          className="svg-inline--fa fa-check fa-w-16 "
+          className="svg-inline--fa fa-check "
           data-icon="check"
           data-prefix="fas"
           focusable="false"
@@ -97,7 +97,7 @@ Array [
       >
         <svg
           aria-hidden="true"
-          className="svg-inline--fa fa-times fa-w-11 "
+          className="svg-inline--fa fa-times "
           data-icon="times"
           data-prefix="fas"
           focusable="false"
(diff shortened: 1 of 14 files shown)


src/Toast/useToast.js
```diff
@@ -48,18 +49,34 @@ const useToast = (initialMessages = []) => {
...
-    message,
-    title,
-  }) => {
-    dispatch({
- type: ACTIONS.SET_MESSAGE,
-payload: {
- action, type, message, title,
-},
-});
+  const setMessage = useCallback((...options) => {
```
gretahayes19, line 52: The `arguments` object cannot be referenced in an arrow function, I opted to use rest args here with `...options` so that I can conditionally check for the correct arg signature.

src/hoc/withTrackedClick.jsx
```diff
@@ -8,13 +8,13 @@ export default function withTrackedClick(Target) {
...
+    const handleClick = (clickEvent) => {
       props.tracking.trackEvent({ ...eventData, event });
 
       if (props.onClick) {
         props.onClick(clickEvent);
       }
-    }
+    };
 
     return <Target {...rest} onClick={handleClick} />;
```
gretahayes19, line 19: Made the change to the `handleClick` above because I was getting a `jsx props should not use functions` linter error on line 19.

src/Toast/__snapshots__/withToast.test.jsx.snap
```diff
@@ -13,7 +13,7 @@ Array [
       >
         <svg
           aria-hidden="true"
-          className="svg-inline--fa fa-circle fa-w-16 "
+          className="svg-inline--fa fa-circle "
```
gretahayes19, line 16: I _think_ these snapshots required changing as a result of a Fontawesome update ? Not 100% certain, but the changes don't seem problematic.

jasonbasuil (approved):
This LGTM! And very exciting! 🙌 

I already tested locally linking to 'rails-server' and it works as expected both with accepting the different parameters as well as bugsnag logging the errors. Thanks so much for leading this!

PR #864, opened 2023-03-15
enforce storybook v6.2.1

### LINKED ISSUE #863: enforce Storybook v6.2.1

jasonbasuil:
closes #863


spec/__snapshots__/Storyshots.test.js.snap
```diff
@@ -21208,7 +21208,7 @@ exports[`Storyshots Components/Tabs Controlled 1`] = `
 >
   <div>
     <span
-      className="sc-dmyDGi goryrA"
+      className="sc-dmqHEX hCuTMq"
```
jasonbasuil, line 21211: might've been some downstream dependency changes... probably to styled-components or another library which changed the snapshots from the [original bugsnag PR](https://github.com/user-interviews/ui-design-system/pull/845/files#diff-d0b8495a07015b3459796609ae72cddb42e598b9cc3640412c5fb75a9c66c0f7)

jasonbasuil: was curious about why that was... but this seems to have reverted it back

rroppolo: Looking at the yarn.lock I think styled-components may have upgraded from 5.3.8 to 5.3.9. Should be fine (unless they pulled a storybook 🙄 )

rroppolo (approved):
Thanks!!

PR #869, opened 2023-03-21
add TableLoadingSkeleton

### LINKED ISSUE #868: add Table loading skeleton

jasonbasuil:
closes #868 

https://user-images.githubusercontent.com/37383785/226713838-d1aca3ba-244d-4ba8-a6c6-e84cbe496f03.mov

[Chromatic](https://62d040e741710e4f085e0647-nxxozkjmox.chromatic.com/?path=/story/components-table--loading-default)

Concept derived from creating a loading skeleton for the participants table in workspace, so making it reusable here. https://github.com/user-interviews/rails-server/pull/17159

- Incorporate the LoadingSkeleton into the top-level `Table` component
- Allows for customizing the number of `columns` and their widths. Also allow customizing number of `rows`.
- Make `TableLoadingSkeleton` available from exports if you need to use it elsewhere (perhaps where it's not as easy to hook up loading state directly into `Table`)


careful-axe (approved):
Fantastic Jason! Really cool that with one prop it can be added to the dope table and also is a stand alone easy to use component. 🙌

domschab23:
If we could find a way to keep the column names, then I think this would provide the best experience in the loading phase

PR #880, opened 2023-03-23
Revert "button size updates (previously reverted)"

jasonbasuil:
Reverting one more time so that we can remove bugsnag for the release tomorrow 😅 these are button updates that will need to be handled and QA'd on a separate release

Reverts user-interviews/ui-design-system#876


gretahayes19 (approved):
LGTM! 

https://www.youtube.com/watch?v=FGBhQbmPwH8

PR #894, opened 2023-04-10
Improve outline transparent button disabled state

### LINKED ISSUE #893: Improve `outline-transparent` button variant disabled state
Right now it isn't very clear when the `outline-transparent` button variant is disabled. Veronica mocked up a version that is more obvious that we want to switch to.

image

gabescarbrough:
closes #893

- Make `outline-transparent` disabled more obvious based on Veronica's design
- Add `disabled` Button examples in Storybook

image


jasonbasuil (approved):
LGTM! Thanks!

PR #900, opened 2023-04-19
Add drawer component

### LINKED ISSUE #897: Add drawer component

gabescarbrough:
closes #897

Adds a `Drawer` component and corresponding `DrawerHeader`, `DrawerBody` and `DrawerFooter` child components.

Chromatic: https://www.chromatic.com/build?appId=62d040e741710e4f085e0647&number=221

image

### DIFF
--- a/src/Drawer/Drawer.mdx
+++ b/src/Drawer/Drawer.mdx
@@ -37,56 +37,22 @@ When there is too much negative space with the content required. Consider a moda
   * The primary and secondary action buttons are encouraged but not required
   * If additional actions or other components are required, children of this component are included to the left of the footer (see example below)
 
-## Stories
+### Size
 
-There are three `size` options, `sm` (default), `md` and `lg`
-
-### Default
-
-<Canvas>
-  <Story id="components-drawer--default" />
-</Canvas>
-
-### Medium
-
-<Canvas>
-  <Story id="components-drawer--medium" />
-</Canvas>
-
-### Large
-
-<Canvas>
-  <Story id="components-drawer--large" />
-</Canvas>
+There are three `size` options, `sm` (default), `md` and `lg`. Choose size via the knobs to see examples.
 
 ### Orientation
 
 Pass `orientation="left"` to override the default behavior of sliding in from the right
 
-<Canvas>
-  <Story id="components-drawer--orientation" />
-</Canvas>
-
 ### Expandable
 
 Setting `expandable` allows the drawer to expand to the full size of the viewport
-
-<Canvas>
-  <Story id="components-drawer--expandable" />
-</Canvas>
 
 ### Additional Actions
 
 If additional components are required to be included, pass them as children to the `DrawerFooter`. This example also doesn't have a title or a border on the header, and the background overlay is turned off to allow interaction with outside elements.
 
-<Canvas>
-  <Story id="components-drawer--additional-actions" />
-</Canvas>
-
 ### Empty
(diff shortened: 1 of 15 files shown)


src/Drawer/Drawer.mdx
```diff
@@ -0,0 +1,102 @@
...
+<h4>
+  Drawers are container elements which slide in and sit on top of an application's main window. To return to the main interface users must click outside of the drawer, click an action within the drawer such as a close or cancel button, or press the escape key.
+</h4>
+
+### When to use 
+When we want to focus the user’s attention on an interactive interface or a large amount of information
+
+### When to not use
+When there is too much negative space with the content required. Consider a modal in those cases.
+
```
jasonbasuil, line 20: We can probably add an ArgsTable here

```
## Props

<ArgsTable of={Drawer} />

```

jasonbasuil: when you add  `subcomponents` to the stories config, it should populate all the component props. Keeps it a little cleaner and puts all the props in one easy place. 

Screen Shot 2023-04-19 at 3 05 32 PM

src/Drawer/Drawer.mdx
```diff
@@ -0,0 +1,102 @@
...
+## Anatomy:
+* **Drawer**
+    * This the parent component and the only one truly required, but its child components are highly encouraged
+    * If you do need to do something truly custom, Drawer can be used as a general purpose container for your own components
+    * Drawer features padding at the top as it is intended to sit behind a navigation bar
```
jasonbasuil, line 25: Non-blocking. I imagine the `48px` which is meant to add padding that reflects our navigation bar won't change anytime soon which I think is fine. I wonder if the Drawer might be possible be used on pages where a navigation bar isn't present and we might want to adjust that, but an override would be okay in that situation if it comes up.

src/Drawer/Drawer.mdx
```diff
@@ -0,0 +1,102 @@
...
+* **DrawerHeader**
+    * Provides a header with a close button, a title (optional) and a button to expand the drawer to the viewport's full width (optional based on a prop to Drawer)
+    * There is also a prop for whether there is a border separating the header from the body. Turning the border off is a good option for drawer's without titles
```
jasonbasuil, line 28: Non-blocking. Could maybe expose the `bordered` prop on the stories via a knob just so people know it's available. Might not be as apparent via the docs.

src/Drawer/Drawer.scss
```diff
@@ -0,0 +1,77 @@
+@import '../../scss/theme.scss';
+
+$drawer-width-small: 32rem;
+$drawer-width-medium: 40rem;
+$drawer-width-large: 55.625rem;
```
jasonbasuil, line 5: I think we could possibly place these in a `drawer.scss` file in `scss` top level folder. We do something similar with card sizes and place them there so that they can be used in RS if needed to be referenced.

src/Drawer/Drawer.scss
```diff
@@ -0,0 +1,77 @@
...
+  &--left {
+    left: -100%;
+
+    &#{$block}--visible {
+      left: 0;
+    }
+  }
+
+  &--visible {
+    box-shadow: -4px 4px 8px rgba(0,0,0,.1);
```
jasonbasuil, line 75: maybe add this to `scss/box_shadow.scss` ?

PR #908, opened 2023-04-24
Drawer/Fix primary action button

DevStarks:
This change uses onPrimaryAction as a click handler for the primary action button.


gabescarbrough:
Thanks for catching and fixing this!

The tests that are failing are Storyshots snapshot tests that are missing the onClick handler you added. If you run `yarn test -u` and commit the new snapshots they will pass.

gabescarbrough (approved):
Looks good!

jasonbasuil (approved):
Good catch! Thank you.

PR #916, opened 2023-05-01
add TableFoot, update TableCell hover, and allow stickyRowBottom

### LINKED ISSUE #915: updates to Table components

jasonbasuil:
closes #915 

Related to this [PR](https://github.com/user-interviews/rails-server/pull/18062)

- Add `TableFoot` component 
- Optionally remove the default hover state on a TableCell
- Add `stickyRow` prop for TableFoot to be sticky at the bottom of the table.

[Table No Hover State](https://62d040e741710e4f085e0647-sluzecrtod.chromatic.com/?path=/story/components-table--table-no-hover-state)

[Table With Footer ](https://62d040e741710e4f085e0647-sluzecrtod.chromatic.com/?path=/story/components-table--table-with-footer)


src/Table/TableCell.scss
```diff
@@ -52,4 +56,10 @@
     top: 0;
     z-index: $z-index-1;
   }
+
+  &--sticky-row-bottom {
+    position: sticky;
+    bottom: 0;
+    z-index: 11;
```
domschab23, line 63: I see we have constants for other z-index values in this file. Maybe we want to follow similar pattern? Not a blocker though

src/Table/Table.stories.jsx
```diff
@@ -484,6 +529,40 @@ export const TableWithMultipleStickyColumnsAndHeader = () => {
...
+          <TableCell>{row.participant}</TableCell>
+          <TableCell>{row.incentive}</TableCell>
+          <TableCell>{row.processingFee}</TableCell>
+          <TableCell>{row.recruitFee}</TableCell>
+          <TableCell>{row.amount}</TableCell>
+        </TableRow>
+        )))}
+    </TableBody>
+    <TableFoot>
+      <TableRow stickyRowBottom>
```
domschab23, line 555: So I think this actually might have some issues where you want to have 2 rows at the bottom be sticky (which is what we're trying to do in the component from rails-server repo that this PR mentioned). With current implementation this will just put 2 rows on top of each other and one will no longer be visible because they both have css prop `bottom: 0;`. This will work under the assumption that there could only be one sticky row at the bottom. I'd vote to move this over to the `TableFoot` component so that multiple rows at the bottom could be sticky

domschab23 (commented):
This is great! Thanks for the quick turnaround on adding this to the DS. Left some comments. Not approving because of sticky row bottom (need to have capability to have multiple rows be sticky at the bottom. Threw a suggestion in the comment)

src/Table/Table.stories.jsx
jasonbasuil: Ah I see what you mean. Missed that part and was assumption of just having one row. 

So you can probably do something like this eventually?

https://user-images.githubusercontent.com/37383785/235744274-f31db528-44a5-41a5-9876-f40925862ad8.mov

michaeldewolf85 (approved):

domschab23 (approved):

PR #942, opened 2023-05-22
Add loading props to DrawerFooter actions

gabescarbrough:
1. Add isLoading and loadingText props for primary and secondary actions
2. Make order of props and propTypes consistent


jasonbasuil (approved):
LGTM!

PR #944, opened 2023-05-22
change Chromatic build deploy on main branch

### LINKED ISSUE #943: change chromatic build deploy on main

jasonbasuil:
closes #943 

We have about a 5,000 snapshot limit on Chromatic every month and I find myself getting closer to the limit more often / I have to think about hitting the limit. 

Changing the deploy step on `main` makes a little more sense at this point since we really just need to inspect the visual diffs before a release anyways. Also just reduces the noise on the Chromatic dashboard with all the pushes to `develop`

Screenshot 2023-05-22 at 12 59 54 PM

Screenshot 2023-05-22 at 12 59 39 PM

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -1,13 +1,13 @@
 # .github/workflows/chromatic.yml
 
 # Workflow name
-name: 'Chromatic build on develop'
+name: 'Chromatic build on main'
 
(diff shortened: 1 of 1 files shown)


brianCollinsUI (approved):
**praise**: 🙌

Good idea to reduce usage

PR #950, opened 2023-05-28
Move deploy to GitHub actions

rsaris:
In looking into what other ways we were able to deploy Storybook I found out that Github has opened up the ability to just deploy via actions (vs. having to use a branch or anything like that). 

I did this via a fork so I could test this w/o disrupting the other work. You can see this page here: https://user-interviews.github.io/ui-design-system-deploy-test/

This still doesn't solve the ability to deploy for a single branch (though I did add the ability to manually deploy this so we could test it and then redeploy from main, though that would obviously be weird for anyone looking at the pages).


.github/workflows/gh-pages-deploy.yml
```diff
@@ -1,30 +1,46 @@
...
+    contents: read
+    pages: write
+    id-token: write
+
   jobs:
     gh-pages-deploy:
       name: Deploying to gh-pages
       runs-on: ubuntu-latest
-      env:
-        FONTAWESOME_NPM_AUTH_TOKEN: ${{ secrets.FONTAWESOME_NPM_AUTH_TOKEN }}
```
rsaris, line 11: Moved this to just be inline.

yarn.lock
```diff
@@ -10162,28 +10120,6 @@ parse-json@^5.0.0, parse-json@^5.2.0:
...
-  integrity sha512-qOpH55/+ZJ4jUu/oLO+ifUKjFPNZGfnPJtzvGzKN/4oLMil5m9OH4VpOj6++9/ytJcfks4kzH2hhi87GL/OU9A==
-  dependencies:
-    protocols "^2.0.0"
-
-parse-repo@^1.0.4:
-  version "1.0.4"
-  resolved "https://registry.yarnpkg.com/parse-repo/-/parse-repo-1.0.4.tgz#74b91d2cb8675d11b99976a0065f6ce17fa1bcc8"
-  integrity sha512-RdwYLh7cmxByP/BfeZX0QfIVfeNrH2fWgK1aLsGK+G6nCO4WTlCks4J7aW0O3Ap9BCPDF/e8rGTT50giQr10zg==
-
-parse-url@^7.0.2:
```
rsaris, line 10177: burninate

jeffbax (approved):
I think @jasonbasuil indicated that individual branches deploy to Chromatic so this seems like it would be good but I'll let him confirm

PR #951, opened 2023-06-01
add MoneyInput component

### LINKED ISSUE #736: Add money input to DS
### How we take money values as input now

- Number Input with step value of `any`
- `app/javascript/researcher/project_builder/project_compensation.jsx` 
- `app/javascript/researcher/project_workspace/details/compensation_fields/compensation_amount_input.jsx`
- `app/views/prepaid_balances/prepaid_balance_transactions/edit.html.erb`

- Number Input with step value of `0.01`
- `app/javascript/researcher/project_workspace/staff/settings/project_settings.jsx`
- `app/views/prepaid_balances/prepaid_balance_transactions/new.html.erb`

### Why we should have a money input in the design system

- Neither step value for number inputs is very good
  - `any` lets you enter whatever decimal point number you want including values much more precise than we generally expect money values to be, and how we display them elsewhere in the user interface.
  - `0.01` stops you from entering values while more decimal points and increments by cents as you would probably expect but
    - The browser based validation is unituitive and doesn't behave how you would expect for money
    - Values such as `2.50` are shown as `2.5` which is unexpected
  - They don't play nice with eachother. If an input of step `any` takes a value of `4.3333` and then you go to edit it on a page that uses `0.01` then `4.33` would be considered invalid because the validation considers both starting value and step when validating
  - Neither masks for separators or considers how to format different currencies, including that different currencies expect different decimal places. While most countries use two decimal places, the Japanese Yen uses none and countries that use the Dinar use three.
- Consistency. We shouldn't have inputs for the same type of value behave differently in different places

### Requirements

- Easy to use input that takes decimals as input and formats them for the appropriate number of decimal places according the locale provided
- Validates the input according to the format provided
- Validates for invalid values such as negative numbers

### Suggestion

- [React Currency Input Field](https://cchanxzy.github.io/react-currency-input-field/) looks like a good option which formats according to the [Internationalization API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl) and allows for custom validation

### Feedback

- Are there other requirements to consider?
- Are we using money inputs in places I missed?

jasonbasuil:
closes #736 

## Context

Chatted with Nat on RX Workflow and decided that it would be helpful to bring in a MoneyInput component which aligns with their current efforts relating to incentive editing (as well as improving the overall validation experience for existing currency inputs)

## Reviewer Expectations
Primary Reviewer: Gabe
Secondary Reviewer(s): Dom, Stephen, Nat

## Testing

You can view a live build in Chromatic here: https://62d040e741710e4f085e0647-ppfritpjca.chromatic.com/?path=/docs/components-moneyinput--docs

One example test from RS in `app/javascript/researcher/project_workspace/details/compensation_fields/index.jsx`

Before:

https://github.com/user-interviews/ui-design-system/assets/37383785/8310a0e9-af02-4e05-9a6a-f185522f14ab

After:

https://github.com/user-interviews/ui-design-system/assets/37383785/c34426af-5bf9-49a5-8e51-3ed8f9ae1a9e

### DIFF
--- a/src/MoneyInput/MoneyInput.jsx
+++ b/src/MoneyInput/MoneyInput.jsx
@@ -20,36 +20,8 @@ MoneyInput.propTypes = {
   */
   allowNegativeValue: PropTypes.bool,
   className: PropTypes.string,
-  /**
-    Specify decimal scale for padding/trimming eg. 1.5 -> 1.50 or 1.234 -> 1.23 if decimal scale 2
-  */
-  decimalScale: PropTypes.number,
-  /**
-    Separator between integer part and fractional part of value
-  */
-  decimalSeparator: PropTypes.string,
-  /**
-    Limit length of decimals allowed
-  */
-  decimalsLimit: PropTypes.number,
   defaultValue: PropTypes.number,
-  /**
-    Disable abbreviations eg. 1k -> 1,000, 2m -> 2,000,000
-  */
-  disableAbbreviations: PropTypes.bool,
   disabled: PropTypes.bool,
-  /**
-    Disable auto adding the group separator between values, eg. 1000 -> 1,000
-  */
-  disableGroupSeparators: PropTypes.bool,
-  /**
-    Value will always have the specified length of decimals
-  */
-  fixedDecimalLength: PropTypes.number,
-  /**
-    Separator between thousand, million and billion
-  */
-  groupSeparator: PropTypes.string,
   /**
     International locale config
   */
@@ -63,18 +35,6 @@ MoneyInput.propTypes = {
   */
   placeholder: PropTypes.string,
   /**
-    Include a prefix eg. £ or $
-  */
-  prefix: PropTypes.string,
-  /**
-    Incremental value change on arrow down and arrow up key press
-  */
-  step: PropTypes.number,
-  /**
-    Include a suffix eg. € or %
-  */
-  suffix: PropTypes.string,
-  /**
     Transform the raw value from the input before parsing. Needs to return string.
   */
   transformRawValue: PropTypes.func,
@@ -87,23 +47,13 @@ MoneyInput.defaultProps = {
   allowNegativeValue: true,
   className: undefined,
   defaultValue: undefined,
+  disabled: false,
+  intlConfig: { locale: 'en-US', currency: 'USD' },
+  maxLength: undefined,
+  placeholder: undefined,
+  transformRawValue: undefined,
   value: undefined,
   onValueChange: undefined,
-  placeholder: undefined,
-  decimalsLimit: 2,
-  decimalScale: undefined,
-  fixedDecimalLength: undefined,
-  prefix: '$',
-  suffix: undefined,
-  decimalSeparator: undefined,
-  groupSeparator: undefined,
-  intlConfig: undefined,
-  disabled: false,
-  disableAbbreviations: false,
-  disableGroupSeparators: false,
-  maxLength: undefined,
-  step: undefined,
-  transformRawValue: undefined,
 };
 
 export default MoneyInput;
--- a/src/MoneyInput/MoneyInput.jsx
+++ b/src/MoneyInput/MoneyInput.jsx
@@ -93,7 +93,7 @@ MoneyInput.defaultProps = {
   decimalsLimit: 2,
   decimalScale: undefined,
   fixedDecimalLength: undefined,
-  prefix: undefined,
+  prefix: '$',
   suffix: undefined,
   decimalSeparator: undefined,
   groupSeparator: undefined,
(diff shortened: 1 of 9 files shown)


natkasman:
The dollar sign finally aligns with the numbers 😂

src/MoneyInput/MoneyInput.jsx
```diff
@@ -0,0 +1,109 @@
...
+  /**
+    Include a suffix eg. € or %
+  */
+  suffix: PropTypes.string,
+  /**
+    Transform the raw value from the input before parsing. Needs to return string.
+  */
+  transformRawValue: PropTypes.func,
+  value: PropTypes.number,
+  onValueChange: PropTypes.func,
```
gabescarbrough, line 82:
```suggestion
  /**
    Allow decimals in input
  */
  allowDecimals: PropTypes.bool,
  /**
    Allow user to enter negative value
  */
  allowNegativeValue: PropTypes.bool,
  className: PropTypes.string,
  defaultValue: PropTypes.number,
  disabled: PropTypes.bool,
  /**
    International locale config
  */
  intlConfig: PropTypes.object,
  /**
    Maximum characters the user can enter
  */
  maxLength: PropTypes.number,
  /**
    Placeholder if no value
  */
  placeholder: PropTypes.string,
  /**
    Transform the raw value from the input before parsing. Needs to return string.
  */
  transformRawValue: PropTypes.func,
  value: PropTypes.number,
  onValueChange: PropTypes.func,
```

**suggestion (blocking)**: 💡 my instinct is that we should expose less of the props that `react-currency-input-field` offers so that it is simpler to use our MoneyInput consistently.

`react-currency-input-field` is really flexible but a lot of the props overlap with `intlConfig` which I think is the more future proof way to configure this. The prefix, suffix, group separator and decimal separator options vary by currency. Setting those via `intlConfig` makes it easier to support currencies besides dollars in the future because you're setting one prop instead of multiple. And if we only support dollars now, we can set `intlConfig` to USD in the defaults.

Other options like `disableGroupSeparators` I don't think we should expose because we would want them to be consistent across our app.

Similarly I could see `step` being useful but I'd prefer to choose a value and pass it here as opposed to exposing that prop, again with the goal of our MoneyInput being consistent wherever it is used.

src/MoneyInput/MoneyInput.jsx
```diff
@@ -0,0 +1,109 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+import CurrencyInput from 'react-currency-input-field';
+
+import './MoneyInput.scss';
+
+const MoneyInput = ({ className, ...props }) => (
+  <CurrencyInput className={classNames(className, 'MoneyInput', 'form-control')} {...props} />
```
gabescarbrough, line 10: **nitpick**: 💅  I lean towards not spreading the props here and instead explicitly passing specific props we want to make available. Because there are so many options we run the risk of MoneyInput being configured differently in many places if we expose all of the options, and then we would be in a similar situation to now where money inputs feel different in different parts of the app.

gabescarbrough (commented):
This is great addition to the design system.

This is the rare case where I think we should make this component less flexible. Since the main pain point we have right now is inconsistency I think we should limit the props we expose so that this input is less configurable and "just works" as much as possible.

Also not exposing the props that override `intlConfig` would leave a clear path to support currencies besides USD in the future without making a lot of changes.

src/MoneyInput/MoneyInput.stories.jsx
```diff
@@ -0,0 +1,82 @@
...
+  title: 'Components/MoneyInput',
+  component: MoneyInput,
+  parameters: {
+    docs: {
+      page: mdx,
+    },
+  },
+};
+
+export const Default = (args) => {
```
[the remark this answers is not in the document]
jasonbasuil: Can't really think of one, so makes sense to just always have the currency symbol displayed.

src/MoneyInput/MoneyInput.jsx
jasonbasuil: Makes sense to me. It does feel easier to handle all of the settings through `intlConfig` and make the default the USD. Good callout 👍

src/MoneyInput/MoneyInput.jsx
jasonbasuil: I go back and forth between wanting to prop spread or not prop spread on a component (teetering between flexibility or strict prop enforcement). I'm not 100% sure what all the other `Input` components or `input` elements that we currently use for currency inputs have as their props/attributes, but it's just nice to have the flexibility to accommodate those when switching them out for the `MoneyInput` just in case we need them (e.g. I saw some of them were using an `aria-label`, `ref`, `name` attribute etc.) 

I think we'll be okay with what we have exposed through the PropTypes declarations and hiding the other ones (unless you dig through the official `react-currency-input-field` docs). The default should be configured and docs shouldn't point anyone to modify hopefully.

domschab23 (approved):
Makes sense

gabescarbrough (approved):
Looks good!

PR #953, opened 2023-06-06
add IconButton

### LINKED ISSUE #952: add IconButton

jasonbasuil:
closes #952 

Screenshot 2023-06-07 at 2 19 11 PM

Chromatic: https://62d040e741710e4f085e0647-rikhiwebhn.chromatic.com/?path=/docs/components-iconbutton--docs

The IconButton component is used for creating clickable buttons with icons, commonly found in app bars, toolbars, and toggleable actions.

We also are wanting to help encourage and enforce a set of pre-defined common actions that will be available with the `IconButton`. We're hoping that this will create a bit more consistency with which icons we use since devs are able to choose from the entire FontAwesome library and it might not always be apparent which icon weight to use.

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -11,6 +11,7 @@ const config = {
     },
     '@storybook/addon-a11y',
     '@storybook/addon-actions',
+    '@storybook/addon-backgrounds',
     '@storybook/addon-controls',
     '@storybook/addon-links',
     '@storybook/addon-knobs',
--- a/src/IconButton/IconButton.jsx
+++ b/src/IconButton/IconButton.jsx
@@ -64,7 +64,10 @@ const IconButton = ({
       variant={variant}
       {...props}
     >
-      <FontAwesomeIcon className="fa-fw" icon={action ? IconButtonActions[action]?.icon : icon} />
+      <FontAwesomeIcon
+        className={classnames('fa-fw', size === ButtonSizes.MEDIUM && 'fa-lg')}
+        icon={action ? IconButtonActions[action]?.icon : icon}
+      />
     </Button>
   );
 };
--- a/src/IconButton/IconButton.jsx
+++ b/src/IconButton/IconButton.jsx
@@ -3,14 +3,15 @@ import classnames from 'classnames';
 import PropTypes from 'prop-types';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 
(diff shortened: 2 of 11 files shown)


.storybook/main.js
```diff
@@ -11,6 +11,7 @@ const config = {
     },
     '@storybook/addon-a11y',
     '@storybook/addon-actions',
+    '@storybook/addon-backgrounds',
```
jasonbasuil, line 14: Addon backgrounds for testing components that lay on top of our different backgrounds.

This PR isn't handling `IconButton` on different backgrounds yet.

src/IconButton/IconButton.jsx
```diff
@@ -0,0 +1,91 @@
...
+      aria-label={getAriaLabel()}
+      className={classnames('IconButton', className)}
+      size={size}
+      variant={variant}
+      {...props}
+    >
+      <FontAwesomeIcon
+        className={classnames('fa-fw', size === ButtonSizes.MEDIUM && 'fa-lg')}
+        icon={action ? IconButtonActions[action]?.icon : icon}
+      />
```
jasonbasuil, line 70: FontAwesome has [relative sizing](https://fontawesome.com/docs/web/style/size#relative-sizing) for their icons. Even though we're setting `size='md'` on `IconButton`, FontAwesome only has `fa-lg` and not `fa-md`. I think the visual distinction between the two sizes is noticeable enough and it seemed like we were using `fa-lg` in a number of places in the app anyways.

src/IconButton/IconButton.jsx
```diff
@@ -0,0 +1,91 @@
...
+  },
+  NEXT: {
+    ariaLabel: 'Next',
+    icon: faChevronRight,
+  },
+  PREVIOUS: {
+    icon: faChevronLeft,
+    ariaLabel: 'Previous',
+  },
+};
```
jasonbasuil, line 47: Any other common actions I might be missing here? 

Also just noting that that we're using all `regular` icon weights here. IIRC we generally stick to this weight. Will need to confirm with @vgordaz

kyleshike: I think close & expand would be good additions:
Screenshot 2023-06-07 at 5 11 01 PM

kyleshike (approved):
Just added a couple icons that I think could be useful but otherwise this looks good

PR #972, opened 2023-07-17
Add RichTextEditor component

### LINKED ISSUE #970: Add Rich Text Editor

gabescarbrough:
closes #970 

Adds a `RichTextEditor` component to input and edit text with formatting. Currently supports bold, italic, link, unordered list, and ordered list.

Built on top of [TipTap](https://tiptap.dev) which supports easily adding further actions including custom extensions. See the [nodes](https://tiptap.dev/api/nodes), [marks](https://tiptap.dev/api/marks) and [extensions](https://tiptap.dev/api/extensions) sections of their docs for existing extensions.

Chromatic: https://www.chromatic.com/build?appId=62d040e741710e4f085e0647&number=255

image

image


src/RichTextEditor/actionHandlers.js
```diff
@@ -0,0 +1,33 @@
+const hasNoHttpProtocol = (url) => url.indexOf('http://') !== 0 && url.indexOf('https://') !== 0;
```
kyleshike, line 1: We could probably use regex here to simplify this a bit if you'd like:
```js
const hasNoHttpProtocol = (url) => !!url.match(/^http(s?):\/\//)
```

src/RichTextEditor/actionHandlers.js
```diff
@@ -0,0 +1,33 @@
...
+
+    // empty
+    if (url === '') {
+      editor.chain().focus().extendMarkRange('link').unsetLink()
+        .run();
+
+      return;
+    }
+
+    const absoluteUrl = hasNoHttpProtocol(url) ? `//${url}` : url;
```
kyleshike, line 22: this might be a bit more simple if we just define the regex here and use it, eg:
```js
const protocolRegex = /^http(s?):\/\//;
const absoluteUrl = url.match(protocolRegex) ? url : `//${url}`;
```

src/RichTextEditor/actionHandlers.js
```diff
@@ -0,0 +1,33 @@
...
+
+    // update link
+    editor.chain().focus().extendMarkRange('link').setLink({ href: absoluteUrl })
+      .run();
```
kyleshike, line 26: Nitpick: does this break line because of an EsLint rule? I typically like to break line for each function call once at least one break is required.  I feel like they scan easier this way, eg:
```js
editor
  .chain()
  .focus()
  .extendMarkRange('link')
  .setLink({ href: absoluteUrl })
  .run();
```

src/RichTextEditor/actionHandlers.js
```diff
@@ -0,0 +1,33 @@
...
+  },
+  unlink: () => editor.chain().focus().unsetLink().run(),
+  bold: () => editor.chain().focus().toggleBold().run(),
+  italic: () => editor.chain().focus().toggleItalic().run(),
+  unorderedList: () => editor.chain().focus().toggleBulletList().run(),
+  orderedList: () => editor.chain().focus().toggleOrderedList().run(),
```
kyleshike, line 32: Total nitpick, but this could be DRY'ed a bit by setting `editor.chain().focus()` to a variable eg:
```js
const chainedFocus = editor.chain().focus();
// ...
  unlink: () => chainedFocus.unsetLink().run(),
  bold: () => chainedFocus.toggleBold().run(),
  italic: () => chainedFocus.toggleItalic().run(),
  unorderedList: () => chainedFocus.toggleBulletList().run(),
  orderedList: () => chainedFocus.toggleOrderedList().run(),
```

src/RichTextEditor/oneLineLimit.js
```diff
@@ -0,0 +1,10 @@
+// a TipTap extension which forces the top node of the editor to not have children
+// which enforces the content being one line long
+
+import { Node } from '@tiptap/core';
```
kyleshike, line 4: I hate that this is called `Node`. Can you imagine if this were also a node app?

src/RichTextEditor/oneLineLimit.js
```diff
@@ -0,0 +1,10 @@
...
+
+export const OneLineLimit = Node.create({
```
kyleshike, line 6: Do you envision a need for further custom nodes in the future? If so, do you think that'd be a thing we'd allow RS to control, or would it be controled here? Not an actionable question, just sort of curious how you envision this component in the long term

src/RichTextEditor/index.js
```diff
@@ -0,0 +1,9 @@
+import RichTextEditor from './RichTextEditor';
+import { AllRteActions, DefaultRteActions, RteActions } from './richTextEditorActions';
+
+export {
+  AllRteActions,
+  DefaultRteActions,
+  RichTextEditor,
+  RteActions,
```
kyleshike, line 8: I feel like these might be a bit more clear if we nix the abreviation here, eg: `AllRichTextEditorActions`, etc.  Also, I'm a little wary of each of these having the `Actions` suffix when they aren't all the same type.  It looks like `RteActions` is currently a simple object of strings, whereas the others are arrays of strings, is that right?  I wonder if we can better differentate the two, or standardize the two formats into a single one instead.

src/RichTextEditor/RichTextEditor.scss
```diff
@@ -0,0 +1,58 @@
...
+  }
+}
+
+.ProseMirror {
+  min-height: 5rem;
+  background: $ux-neutral-100;
+  border: 1px solid $ux-gray-400;
+  border-radius: 0.25rem;
+  border-top-left-radius: 0;
+  border-top-right-radius: 0;
```
kyleshike, line 26: Can we use the same styles here that we do for the toolbar, just reversed?  eg:  `border-radius: 0 0 0.25rem 0.25rem;`?

src/RichTextEditor/RichTextEditor.test.jsx
```diff
@@ -0,0 +1,28 @@
...
+    expect(asFragment()).toMatchSnapshot();
+  });
+
+  describe('given an initial value', () => {
+    it('deserializes value correctly', () => {
+      render(<Setup initialValue="<p>hello world</p>" />);
+
+      expect(screen.getByText('hello world')).toBeInTheDocument();
+    });
+  });
```
kyleshike, line 27: Can we get some specs here to cover the supported actions? eg: bold, etc.

src/RichTextEditor/RichTextEditorMenuBar.jsx
```diff
@@ -0,0 +1,110 @@
...
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+
+import {
+  faBold,
+  faItalic,
+  faLink,
+  faListOl,
+  faListUl,
+  faUnlink,
+} from '@fortawesome/pro-regular-svg-icons';
```
kyleshike, line 19: @jasonbasuil just want to make sure this the icon set that we'll want to be using. I feel like we definitely want to get down to just one icon set in the future if possible.

src/RichTextEditor/RichTextEditorMenuBar.jsx
```diff
@@ -0,0 +1,110 @@
...
+}) => {
+  const actionHandlers = createActionHandlers(editor);
+
+  const actions = [
+    {
+      label: 'Bold',
+      name: RteActions.BOLD,
+      disabled: availableActions.includes(RteActions.BOLD) && !editor.can()
+        .chain()
+        .focus()
```
kyleshike, line 36: We're doing something pretty similar in `actionHandlers`.  Is it possible to consolidate that & these calls to `chain`, `focus`, etc?

src/RichTextEditor/RichTextEditor.scss
```diff
@@ -0,0 +1,58 @@
+@import '../../scss/theme.scss';
+
+.RichTextEditor {
+  width: 100%;
+
+  .RichTextEditor__character-count {
+    font-size: 0.75rem;
+    line-height: 1rem;
```
jasonbasuil, line 8: suggestion: could probably use `font-type-20` here

src/RichTextEditor/RichTextEditor.jsx
```diff
@@ -0,0 +1,215 @@
...
+          availableActions={availableActions}
+          editor={editor}
+        />
+        <EditorContent
+          className={classNames(
+            className,
+            'RichTextEditor__field',
+            { 'RichTextEditor__field--error': hasErrors },
+          )}
+          editor={editor}
```
kyleshike, line 148: Can we make this stronger somehow? I could be mistaken but it looks like we're allowing `RichTextEditor` to essentially accept anything for props and pass it to `EditorContent`

src/RichTextEditor/RichTextEditor.scss
```diff
@@ -0,0 +1,58 @@
...
+  .ProseMirror {
+    border-color: $ux-red-600;
+  }
+}
+
+.ProseMirror {
+  min-height: 5rem;
+  background: $ux-neutral-100;
+  border: 1px solid $ux-gray-400;
+  border-radius: 0.25rem;
```
jasonbasuil, line 24: suggestion: `$ux-border-radius`

kyleshike (commented):
This is awesome.  I left a bunch of comments, happy to chat or huddle to talk about them if needed

src/RichTextEditor/RichTextEditor.jsx
```diff
@@ -0,0 +1,215 @@
...
+      >
+        <RichTextEditorMenuBar
+          availableActions={availableActions}
+          editor={editor}
+        />
+        <EditorContent
+          className={classNames(
+            className,
+            'RichTextEditor__field',
+            { 'RichTextEditor__field--error': hasErrors },
```
jasonbasuil, line 146: Just for my own curiosity, what would we be expecting for the error state? Would we be validating on character count (maybe empty char count?) or something else?

src/RichTextEditor/RichTextEditorMenuBar.jsx
```diff
@@ -0,0 +1,110 @@
...
+          .map((action) => (
+            <Button
+              aria-label={action.label}
+              className={classNames({ 'Button--active': editor.isActive(action.name) })}
+              disabled={action.disabled}
+              key={action.name}
+              variant="transparent"
+              onClick={action.onClick}
+            >
+              <FontAwesomeIcon icon={action.icon} />
```
jasonbasuil, line 97: Suggestion non-blocking: Wondering if we can make all of these icons the same width.

```
<FontAwesomeIcon className="fa-fw" icon={action.icon} />
```

Alternatively we do have the `IconButton` component that does the same thing (though the common `action` options don't currently include the icons for this RichTextEditor component but could be added via the `icon` prop), but those icons probably wouldn't be used outside of the text editor most likely? 

Screenshot 2023-07-18 at 12 50 20 PM

Screenshot 2023-07-18 at 12 50 03 PM

src/RichTextEditor/RichTextEditorMenuBar.scss
```diff
@@ -0,0 +1,23 @@
+@import '../../scss/theme.scss';
+
+.RichTextEditorMenuBar {
+  background-color: $ux-gray-100;
+  border: 1px solid $ux-gray-400;
+  border-bottom: 0;
+  border-radius: 0.25rem 0.25rem 0 0;
```
jasonbasuil, line 7: `$ux-border-radius`

src/index.js
```diff
@@ -59,6 +59,12 @@ import Popper from 'src/Popper';
 import ProfileCell, { ProfileCellSkeleton } from 'src/ProfileCell';
 import RadioButton from 'src/RadioButton';
 import RadioButtonGroup from 'src/RadioButtonGroup';
+import {
+  AllRteActions,
+  DefaultRteActions,
+  RichTextEditor,
+  RteActions,
```
jasonbasuil, line 66: Suggestion non-blocking:

Curious if it's possible to rename these to all start with Rte or RichTextEditor. Just thinking from an import perspective, I might not readily remember what is all available for the RichTextEditor, so by having it all start with the same name I can see what is available.

jasonbasuil (commented):
This is stellar and really excited to see this in here and used in other places in the app! ⭐ Left a few non-blocking suggestions to begin with, but otherwise looking really good.

src/RichTextEditor/RichTextEditorMenuBar.jsx
jasonbasuil: For the vast majority of icons we use in the app, we will be using the regular weight icons. I agree that parring down our icon set is something we're wanting to do in a separate effort with Design.

src/RichTextEditor/actionHandlers.js
gabescarbrough: Good call. I'll make this tweak.

package.json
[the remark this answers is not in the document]
gabescarbrough: I'm not sure how to compare bundle size to be honest. How would you go about measuring that?

src/RichTextEditor/RichTextEditor.jsx
gabescarbrough: Where this is currently used it is just being validated to not be empty.

src/RichTextEditor/RichTextEditor.test.jsx
kyleshike: Either blocking or as a fast follow, yea. In my experience specs that aren't written with the code they cover often end up unwritten.

src/RichTextEditor/RichTextEditorMenuBar.jsx
kyleshike: yea I think I do.  Let me try and mess around with it tomorrow morning and offer a suggestion

src/RichTextEditor/RichTextEditor.jsx
kyleshike: At the least, filtering the rest down to valid DOM props, or in this case I suppose valid props of `EditorContent`.

src/RichTextEditor/RichTextEditor.jsx
```diff
@@ -0,0 +1,214 @@
...
+  /**
+   HTML provided to initialize the editor's content
+  */
+  initialValue: propTypes.string,
+  isOneLine: propTypes.bool,
+  placeholder: propTypes.string,
+  /**
+    Callback function to call with sanitized HTML when editor state changes
+  */
+  onChange: propTypes.func.isRequired,
```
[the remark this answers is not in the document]
kyleshike: Blur is commonly used to present any input errors if any are present. I think we should by default include it in any input component we add to the DS

package.json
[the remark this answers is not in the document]
kyleshike: https://www.npmjs.com/package/webpack-bundle-analyzer

You can see how it's implemented in RS here, though the implementation will be different as RS config is built from shaka: https://github.com/user-interviews/rails-server/blob/main/config/webpack/base.js#L117-L129

src/RichTextEditor/RichTextEditorMenuBar.jsx
gabescarbrough: Great call out. I was thinking about this before and wasn't sure which way to go.

> (though the common action options don't currently include the icons for this RichTextEditor component but could be added via the icon prop), but those icons probably wouldn't be used outside of the text editor most likely?

This was my reasoning for not using `IconButton` before.

After thinking about it a bit I went ahead and switched to `IconButton`. I was basically recreating it, so it makes for these buttons to get any future updates to `IconButton` like if we add hover tooltips in the future or something.

src/RichTextEditor/RichTextEditorMenuBar.jsx
kyleshike: So, rather than having to run the same set of commands both here and in `createActionHandlers`, I think it will be better to at the least co-locate the command sequence in `createActionHandlers`, eg:
```js
  italic: {
    onClick: () => editor.chain().focus().toggleItalic().run(),
    disabled: !editor.can().toggleItalic(),
  },
```

Then here in the component it'd just be:
```jsx
  {
      label: 'Italic',
      name: RichTextEditorActions.ITALIC,
      disabled: actionHandlers.italic.disabled,
      onClick: actionHandlers.italic.onClick,
      icon: faItalic,
    },
```

Also, what's the need for checking `availableActions.includes(RichTextEditorActions.ITALIC)` in the disabled state? As far as I can tell, `createActionHandlers` will _always_ return an object keyed with `link, unlink, bold, italic, unorderedList, orderedList`.  Am I missing something or can this just be removed like in my pseudocode above?

src/RichTextEditor/index.js
gabescarbrough: I made this change.

src/RichTextEditor/RichTextEditor.jsx
gabescarbrough: > At the least, filtering the rest down to valid DOM props

As far as I know React doesn't differentiate between HTML attributes and props as far as passing them to child components. I'm not aware of a pattern to pass down HTML attributes but not props.

The main patterns I've seen for passing HTML attributes to child components are prop spreading and adding explicit props (such as `ariaName` for `aria-name`).

Explicit props are a pain because if the specific ARIA attribute you want to use isn't supported you just can't use it - which was actually a pain point with our old WYSIWYG component and affects accessibility. It would be quite an effort to add a prop for every possible ARIA attribute to all of our DS components.

> or in this case I suppose valid props of EditorContent

TipTap actually supports this exact use-case [the same way in EditorContent](https://github.com/ueberdosis/tiptap/blob/develop/packages/react/src/EditorContent.tsx#L142).

I know prop spreading is generally discouraged, but in the case of form elements that should be able to support any ARIA attribute it is probably the best option. And if we decide that it isn't then we should decide on an alternative for all DS components and have an effort to switch so that all of them are consistent.

src/RichTextEditor/RichTextEditor.jsx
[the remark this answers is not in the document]
gabescarbrough: > I think we should by default include it in any input component we add to the DS, even we don't use it (yet) in RS.

I disagree with this. I think any new input component should have consistent event handlers with the existing input components.

I'm not opposed to adding `onBlur`. `onFocus`, etc. but I think if we do so they should be added to all of the input components within the same release.

You should be able to assume that all of the input components support the same event handlers.

src/RichTextEditor/actionHandlers.js
gabescarbrough: I made this change.

src/RichTextEditor/RichTextEditor.jsx
kyleshike: > As far as I know React doesn't differentiate between HTML attributes and props

React will log a warning about an unsupported prop and recommend removing it: https://codesandbox.io/s/wispy-voice-vmddzn?file=/src/App.js

You can read an interesting conversation about why this is probably not a good thing to do here: https://gist.github.com/statianzo/91234d26094ad89c1ff742f41b25d737#gistcomment-1838594.

If we want to forward props from our component into the child, then I'd prefer we be explicit about it and use a prop eg `editorContentProps` or if it's just for a11y then `ariaProps: propTypes.shape({ // ...etc }}` and not spread literally anything and everything from `RichTextEditor` into `EditorContent` just to make aria props easier.

src/RichTextEditor/RichTextEditor.test.jsx
gabescarbrough: I'd prefer to add more tests as a fast follow. Merging this PR is blocking removing the old WYSIWYG from RS so I'd rather follow "fast beats perfect" in this case.

package.json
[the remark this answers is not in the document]
gabescarbrough: Does the bundle size affect whether this PR can be merged?

We're already importing only the TipTap packages that we actually use so we wouldn't be able to build the same component with less dependencies.

This PR also blocks removing Plate, Slate and TipTap from RS so merging this PR will cause our overall bundle size to go down for the main repo.

Measuring bundle size for the DS is a great idea but I think editing our Webpack config is out of scope for this PR.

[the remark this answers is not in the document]
kyleshike: I'm not suggesting that you commit changes to our webpack config in this PR, but I think understanding the consequence of adding a new dependency is _absolutely_ within scope of the PR that introduces it, and this one is adding 17. If you don't want to tinker with webpack to get this information, you can probably use something like https://bundlephobia.com/ to do the job.  I don't know a ton about that site, I just did a little googling.

src/RichTextEditor/RichTextEditorMenuBar.jsx
gabescarbrough: > Also, what's the need for checking availableActions.includes(RichTextEditorActions.ITALIC) in the disabled state? As far as I can tell, createActionHandlers will always return an object keyed with link, unlink, bold, italic, unorderedList, orderedList. Am I missing something or can this just be removed like in my pseudocode above?

It can't be removed, or at least if it was there would need to be an alternative that serves the same need.

Only the extensions which correspond to `availableActions` are initialized. Otherwise you would be able to used the keyboard shorts for extensions which aren't in the menu bar. That check is avoiding a `null` error.

package.json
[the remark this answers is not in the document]
gabescarbrough: According to Bundlephobia:

### DS without TipTap
* 261.9 kB MIN
* 86.9 kB MIN + GZIP

### DS with TipTap
* 600.6 kB MIN
* 217.8 kB MIN + GZIP

So between 2 to 2.5x the bundle size.

### Slate and Plate dependencies in RS are:
* 679 kB MIN
* 203.4 kB MIN + GZIP

So the entire DS with TipTap is less or equal to just the size of our Plate/Slate dependencies that it allows us to get rid of.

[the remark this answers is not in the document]
kyleshike: Awesome, thank you.  This is exactly what I was hoping to confirm. If it were significantly _larger_ than our deps before, it'd probably be something to give a little more thought before we double down on TipTap.

Just for clarity: does this data comparison includes the extensions for TipTap in addition to `@tiptap/core`?

src/RichTextEditor/RichTextEditor.jsx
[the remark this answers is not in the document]
gabescarbrough: It makes sense to me to add an issue to add more event handlers to our input components, including this one.

I think it makes more sense to include a checklist of those components in the issue, than to add a `TODO` here, since we would want to add the new handlers to all of the input components to be consistent.

src/RichTextEditor/RichTextEditorMenuBar.jsx
gabescarbrough: It doesn't make sense to me to move `disabled` into `createActionHandlers` because `disabled` is only relevant to the menu bar (whether the button is disabled). `createActionHandlers` is also used in `RichTextEditor` when registering a keyboard shortcut for links.

src/RichTextEditor/RichTextEditor.jsx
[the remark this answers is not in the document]
kyleshike: As long as it's tracked and we are actively working to take care of it, I don't really care if there are TODO comments or not I suppose.

src/RichTextEditor/RichTextEditorMenuBar.jsx
kyleshike: My primary suggestion here is to reduce the redundancy here.  Without being co-located, drift between the two is much higher. Also it seems to me co-locating an event handler with a disabled state is perfectly reasonable.  Unless of course we're calling these tip tap actions in ways that aren't in response to user input, but that seems unlikely.

> is also used in RichTextEditor when registering a keyboard shortcut for links.

Wouldn't we also want to register keyboard shortcuts if and only if `editor.can()...` is true for the given action? Seems like we don't check that on links, is there a reason why some of these actions check if the command can be executed and others do not?

src/RichTextEditor/RichTextEditor.jsx
gabescarbrough: > React will log a warning about an unsupported prop and recommend removing it

For sure, and I agree that prop spreading should be avoided when it comes to passing down component specific props. The reason for it in this case is to be able to pass down ARIA attributes for accessibility. There are a lot of ARIA attributes so it wouldn't make sense to add a prop for each possible ARIA attribute. In this case I think the cost of devs having to be mindful of what props they pass is worth the reward of it being possible to provide any ARIA attribute when you need to for accessibility purposes.

> You can read an interesting conversation about why this is probably not a good thing to do here: https://gist.github.com/statianzo/91234d26094ad89c1ff742f41b25d737#gistcomment-1838594.

I've read this discussion before and agree with it when it comes to passing props. It doesn't offer a better solution for passing HTML attributes such as ARIA attributes without explicitly providing a prop for each one.

> If we want to forward props from our component into the child, then I'd prefer we be explicit about it and use a prop eg editorContentProps or if it's just for a11y then ariaProps: propTypes.shape({ // ...etc }} and not spread literally anything and everything from RichTextEditor into EditorContent just to make aria props easier.

What would the shape of `ariaProps` look like that allows all ARIA attributes? Would we have to maintain a list of all possible ARIA attributes? I haven't been able to find a package that provides that and is actively maintained.

I realize prop spreading is a compromise here, and that in other situations it is a bad choice. In this case I don't see a better choice. This isn't being done to make passing down ARIA attributes _easier_ it is being done to make passing down ARIA attributes _possible_.

It's totally possible there is a better solution to this problem. If so I think we should have a separate PR to move to that solution throughout the DS as opposed to blocking this PR.

package.json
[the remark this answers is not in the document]
gabescarbrough: It includes all TipTap packages we are using.

src/RichTextEditor/RichTextEditorMenuBar.jsx
gabescarbrough: > Wouldn't we also want to register keyboard shortcuts if and only if editor.can()... is true for the given action? 

I think you're conflating two concepts here. Keyboard shortcuts are registered on initialization. Just because they are initialized doesn't mean they are always available no matter what you're doing in the editor. For instance, you can't make text bold if you're currently within a code block. If you do `CMD-B` inside a code block it still won't let you made the text bold. `disabled` in this case is checking specifically if an action in the menu bar should be disabled because it currently is not allow based on where you are in the rich text in the editor. It fully has to do with the menu bar actions and not with the action handlers.

> Seems like we don't check that on links, is there a reason why some of these actions check if the command can be executed and others do not?

We're registering a custom keyboard shortcut for links because a default isn't provided. The others have defaults. It is unrelated to the conversation above.

> My primary suggestion here is to reduce the redundancy here. Without being co-located, drift between the two is much higher.

I get what you're saying, but I think putting the code in the place that relates to its purpose makes more sense than putting it next to similar looking code that is used for a different purpose. Moving it into `actionHandlers` makes it look like `disabled` has to do with how actions are handled when it really has to do with with whether a specific button in a specific place should be disabled.

I also don't think there is a high likelihood that we will change that code frequently since they are the exact method chains recommended by the docs. The only case I could see us changing it if is TipTap has a breaking change to their API, in which case we would want to vigilant about making changes anyway.

src/RichTextEditor/RichTextEditor.jsx
kyleshike: > It doesn't offer a better solution

It actually does: use a named prop for the things we're forwarding and spread _it_ instead:
```jsx
const ContactBadgeBetter = ({ user, contactMethod, divProps }) => (
  <div className='Badge' {...divProps} />
}
```
as opposed to
```jsx
const ContactBadge = ({ user, contactMethod, ...rest }) => (
  <div className='Badge' {...rest} />
}
```

This way at least, the props that are being forwarded are very intentional by separating the component’s props and the props you want to pass through.  This helps prevent any consumer of the component from inadvertently forward things they don't intend to.

Down the line, it's not too hard to build a tool to filter invalid dom props.  I've done it before.  The html spec doesn't change _that_ frequently but yea you'd be on the hook to keep it up to date.  That updating could theoretically be done lazily, eg: when a dev is trying to forward a valid html prop but the filtered list is incorrect, but all of that is for another day and probably not worth it.

> What would the shape of ariaProps look like that allows all ARIA attributes

Ideally yea it'd be exhaustive.  But I'd settle for just `propTypes.Object` if we can put these forwarded props into their own namespace.

gabescarbrough: I made this change

kyleshike (approved):
Well, we definitely went back and forth quite a bit on some things here. I'm going to work on standardizing some expectations around prop spreading & html props, avoiding code repetition, what should be done when adding a new dependency, etc so they needn't be discussions on pull request, but rather expectations as part of the work done.

Nice work

jasonbasuil (approved):
Thanks for all the thought that went into this!

PR #980, opened 2023-08-01
upgrade FontAwesome v6.4.0

### LINKED ISSUE #979: upgrade FontAwesome v6.4.0

jasonbasuil:
closes #979 

Updating FontAwesome on the Design System to match rails-server.

We shouldn't have anything breaking here since we exclusively use the `<FontAwesomeIcon />` React component here and use imports, not classes.


careful-axe (approved):
LGTM!

PR #992, opened 2023-08-31
WIP - add TypeScript

jasonbasuil:


src/EmptyState/EmptyState.tsx
```diff
@@ -1,24 +1,25 @@
...
 
 import './EmptyState.scss';
 
-export const EmptyStateMarginTopSizes = {
-  SMALL: 'sm',
-  MEDIUM: 'md',
-  LARGE: 'lg',
-  NONE: 'none',
-};
+export interface IEmptyStateProps {
```
jasonbasuil, line 10: I think prefixing the interface with `I` could be helpful with recognizing at a glance... at least that was the convention I followed when I last used TS.

src/EmptyState/EmptyState.tsx
```diff
@@ -1,24 +1,25 @@
...
 import { Text } from 'src/Text';
 
 import './EmptyState.scss';
 
-export const EmptyStateMarginTopSizes = {
-  SMALL: 'sm',
-  MEDIUM: 'md',
-  LARGE: 'lg',
-  NONE: 'none',
-};
```
jasonbasuil, line 16: this isn't consumed on RS so should be fine to remove and is taken care of with the interface.

src/EmptyState/EmptyState.tsx
```diff
@@ -1,24 +1,25 @@
...
+  title?: ReactNode;
+}
 
-const EmptyState = ({
+const EmptyState: React.FC<IEmptyStateProps> = ({
   className,
-  fullWidth,
-  marginTop,
+  fullWidth = false,
+  marginTop = 'sm',
```
jasonbasuil, line 22: put default props here that used to be on propTypes

src/EmptyState/EmptyState.tsx
```diff
@@ -53,31 +54,4 @@ const EmptyState = ({
...
-    Sets a predefined margin-top depending on spacial context.
-  */
-  marginTop: propTypes.oneOf(Object.values(EmptyStateMarginTopSizes)),
-  /**
-    Accepts a React node, in most cases a Button component.
-  */
-  primaryAction: propTypes.node,
-  subtitle: propTypes.oneOfType([propTypes.string, propTypes.node]),
-  title: propTypes.node,
-};
```
jasonbasuil, line 72: I do want to follow up on documentation practices in this new format with TS since I have followed this pattern of commenting on props here and having it show on the Docs pages for the components.

package.json
```diff
@@ -105,6 +109,8 @@
     "@testing-library/react": "^12.1.2",
     "@testing-library/react-hooks": "^3.2.1",
     "@testing-library/user-event": "^13.5.0",
+    "@types/react": "16.14.0",
+    "@types/react-dom": "^18.0.11",
```
jeffbax, line 113: are react libraries not version-matched? ie. react-dom would also be 16.x

src/EmptyState/EmptyState.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+import { Heading } from '../Heading';
+import { Text } from '../Text';
+
+import './EmptyState.scss';
+
+export interface EmptyStateProps {
+  className?: string;
+  fullWidth?: boolean;
+  marginTop?: 'sm' | 'md' | 'lg' | 'none';
```
jeffbax, line 13: nice that TS builds in enums like this

src/Layout/SidebarView/SidebarView.jsx
```diff
@@ -145,6 +145,7 @@ export default class SidebarView extends Component {
     const sidebarClassName = classNames(
       this.props.className,
       {
+        // eslint-disable-next-line @typescript-eslint/camelcase
```
jeffbax, line 148: is this going to get sprinkled all over or is there a rule we need to adjust

jeffbax (approved):
looks good as far as I can tell

src/EmptyState/EmptyState.tsx
kyleshike: https://storybook.js.org/docs/react/configure/typescript might be helpful.  It looks like 
```ts
typescript: {
    reactDocgen: 'react-docgen-typescript',
    reactDocgenTypescriptOptions: {
      compilerOptions: {
        allowSyntheticDefaultImports: false,
        esModuleInterop: false,
      },
      propFilter: () => true,
    },
  },
```

Might do the trick, but I'm fine with having that investigation be a fast follow

src/EmptyState/EmptyState.tsx
kyleshike: Yea this is part of what is going to make TS in the DS incredibly useful over in rails server.

FYI this is actually a union of literals, and an enum exactly:

https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#union-types
https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#literal-types

_But_ TS does have some awesome enums as well:
https://www.typescriptlang.org/docs/handbook/enums.html

src/Layout/SidebarView/SidebarView.jsx
kyleshike: Hopefully it won't need to be used a lot.. I haven't looked to deeply at how styles are implemented here in the DS, but if we use CSS modules we can do away with BEM here & elsewhere.

src/EmptyState/EmptyState.tsx
jasonbasuil: Noted, and yeah I'll plan on doing this as a follow up

package.json
jasonbasuil: Not entirely sure. I followed what we were already using on RS, but maybe @kyleshike would know?

kyleshike: Yikes yea these should be version matched.  I guess it hasn't come up because all the react-dom stuff is happening under the hood with `react-rails` over in rails-server

kyleshike (approved):
Looks great!  Just need to bring down the version of `@types/react-dom` to match the package itself

PR #993, opened 2023-09-05
Fixes up TS config

kyleshike:


kyleshike:
@jasonbasuil one key part of this is to remove this line in rails server: https://github.com/user-interviews/rails-server/blob/main/app/javascript/lib/ts_module_shims.ts#L47

package.json
```diff
@@ -18,8 +18,8 @@
     "@tiptap/extension-text": "^2.0.3",
     "@tiptap/pm": "^2.0.3",
     "@tiptap/react": "^2.0.3",
-    "@typescript-eslint/eslint-plugin": "^6.6.0",
-    "@typescript-eslint/parser": "^6.6.0",
+    "@typescript-eslint/eslint-plugin": "^2",
```
jasonbasuil, line 21: 🤔 Just curious what you found with the downgrade since it's a noticeable version difference?

stories/Spacing/Spacing.stories.jsx
```diff
@@ -52,3 +42,13 @@ export const Spacing = () => (
...
+
+export default {
+  title: 'Foundations/Spacing',
+  component: Spacing,
+  parameters: {
+    docs: {
+      page: mdx,
+    },
+  },
+};
```
jasonbasuil, line 54: I feel like I recall seeing some order issue that Storybook was not happy about. Is this why this is moved to the bottom?

.eslintrc
```diff
@@ -122,6 +122,7 @@
     }
   },
   "rules": {
+    "@typescript-eslint/explicit-function-return-type": "off",
```
jasonbasuil, line 125: Do we imagine we would turn this back on at some point?

jasonbasuil (approved):
Just a couple clarifying comments, but looks great!

package.json
kyleshike: I had to bring it down to v2 to make it compliant with the version of eslint, which is pretty far behind, hence this diff

stories/Spacing/Spacing.stories.jsx
kyleshike: In updating to use the typescript parser for eslint, a few new rules violations were popping up.  This change specifically addresses this error:
Screenshot 2023-09-06 at 3 45 44 PM

.eslintrc
kyleshike: I don't think so, no.  The newer versions of `typescript-eslint` actually don't even have this rule on by default.  

Generally speaking, it's a lot better to let TS do the work for you and infer a function return type.  If you're interested, here's a great video that does a better job explaining how its useful: https://www.youtube.com/watch?v=I6V2FkW1ozQ&feature=youtu.be

jasonbasuil: Nice, thanks for sharing.

PR #1011, opened 2023-09-15
Adds necessary final step for unlinking rs

kyleshike:
Occasionally when running `yarn rs:unlink`, rails-server would then fail to compile.  Running `yarn --check-files` after unlinking should prevent this from happening moving forward.  Will need to monitor to confirm this is fully resolved.


jasonbasuil (approved):
🙌

PR #1018, opened 2023-09-20
convert Flex component to TypeScript

### LINKED ISSUE #1007: [UIDS] Flex components should be able to define their own html element
## Description

There are situations where we want something other than a div to be the primary flex container (ie: `<nav />`).

## User Stories

- As an engineer, I want to use `<Flex />` on elements other than a div so that I can benefit from flex alignment on other html elements..
- 

## Acceptance Criteria

- a prop is added to `<Flex />` so that devs can use html elements other than a div

### LINKED ISSUE #1016: [UIDS] TS conversion - Flex
## Description

`Flex` should be converted to use TS.

jasonbasuil:
closes #1016 
closes #1007 

Converts `Flex` to use TypeScript
Add `as` prop to `Flex` so you can define any other element other than a `div`

Here's a screenshot of the Storybook. (apologies, I would I have provided a Chromatic link, but there's currently an issue with the build process that I need to resolve)

https://github.com/user-interviews/ui-design-system/assets/37383785/81103537-aae2-4464-bb49-e440e8946b68

Reviewers

Primary: Gabe, Jeff
Secondary: Kyle


src/Flex/Flex.tsx
```diff
@@ -0,0 +1,87 @@
+import React, { ReactNode, ElementType } from 'react';
+import { FlexWrapper } from './FlexWrapper.styles';
+
+export interface FlexProps {
+  alignItems?: 'stretch' | 'center' | 'flex-start' | 'flex-end' | 'baseline' | 'initial' | 'inherit';
+  alignSelf?: 'stretch' | 'center' | 'start' | 'end';
+  /**
+    You can use a custom element for this component other than the default `div`
+  */
+  as?: ElementType;
```
jasonbasuil, line 10: works as expected with creating descriptions for prop tables in Storybook
Screenshot 2023-09-20 at 3 01 39 PM

src/Flex/Flex.jsx
```diff
@@ -1,66 +0,0 @@
-import React from 'react';
```
jasonbasuil, line 1: Using the Ratchet tool is a good start, but if proptypes objs are defined in another file it won't pull. Still neat tho

Screenshot 2023-09-20 at 3 16 19 PM

src/Flex/Flex.tsx
```diff
@@ -0,0 +1,87 @@
...
+  flexWrap,
+  height,
+  justifyContent,
+  justifySelf,
+  maxHeight,
+  maxWidth,
+  width,
+  ...props } : FlexProps) => (
+  <FlexWrapper
+    as={as}
```
jasonbasuil, line 64: `as` prop works as expected

Screenshot 2023-09-20 at 3 28 22 PM

Screenshot 2023-09-20 at 3 28 41 PM

jeffbax (approved):
seems like these are going to be pretty straightforward -- 

fwiw, I'm fine with leaving oversight to gabe & kyle just was wondering if there was any tooling to help you along. 

I only found those others w/ ~5 mins, maybe there are better ones the others are aware of. Kind of surprised MS doesn't have one, given they had built some others in the past for things like VB -> C#

jasonbasuil:
> seems like these are going to be pretty straightforward --
> 
> fwiw, I'm fine with leaving oversight to gabe & kyle just was wondering if there was any tooling to help you along.
> 
> I only found those others w/ ~5 mins, maybe there are better ones the others are aware of. Kind of surprised MS doesn't have one, given they had built some others in the past for things like VB -> C#

Sounds good. Yeah I think the actual component transition is straightforward and its more of an opportunity to revisit how stories are written, updating docs, etc that haven't been revisited in a while.

PR #1038, opened 2023-09-29
Migrates Tabs component to TS  and CSS modules

kyleshike:
This converts the Tabs component to use css modules (which plays nicer with SSR) and also converts the component to TS.

While in the process of converting to TS, I found a few things that slipped through the cracks in the initial conversion, namely eslint and jest types.  These have both been corrected as well.


.eslintrc
```diff
@@ -23,7 +23,7 @@
       "env": {
         "browser": true
       },
-      "parser": "babel-eslint",
+      "parser": "@typescript-eslint/parser",
```
kyleshike, line 26: This was an oversight when initially setting up TS.  Some changes in this PR are fix linting issues as a consequence.

.eslintrc
```diff
@@ -112,6 +112,12 @@
         "react-hooks/rules-of-hooks": "error",
         "semi": 0
       }
+    },
+    {
+      "files": "**/*.{ts,tsx}",
+      "rules": {
+        "react/require-default-props": "off"
+      }
```
kyleshike, line 120: We don't really need default props for ts components. We can just provide default function arguments instead.

.storybook/webpack.config.js
```diff
@@ -3,7 +3,16 @@ const path = require('path');
...
+    use: [
+      'style-loader',
+      {
+        loader: 'css-loader',
+        options: {
+          modules: true
+        }
+      },
+      'sass-loader'
+    ],
```
kyleshike, line 15: This enables css modules

src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,37 @@
+import type { TabsProps as BootstrapTabProps } from 'react-bootstrap/Tabs';
+
+import React from 'react';
+import classNames from 'classnames';
+
+import ReactBootstrapTabs from 'react-bootstrap/Tabs';
+import styles from './tabs.module.scss';
+
+export interface TabsProps extends BootstrapTabProps {
```
kyleshike, line 9: @gabescarbrough this makes that issue we discussed regarding spreading props into another component go away 🎉

gabescarbrough: 🔥 that's awesome

src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,38 @@
...
+  flexWrapUnset?: boolean;
+  navItemButtonFullHeight?: boolean;
+}
+
+const Tabs = ({
+  children,
+  flexWrapUnset,
+  navItemButtonFullHeight,
+  ...props
+}: TabsProps) => (
```
gabescarbrough, line 19: Am I understanding correctly that because we're using the interface `TabsProps` the spreaded props are being type checked?

If someone uses this component in our main repo and passes props that aren't in the interface does TypeScript give a compiler error?

src/Tabs/__snapshots__/Tabs.test.tsx.snap
```diff
@@ -5,101 +5,6 @@ Object {
   "asFragment": [Function],
   "baseElement": <body>
     <div>
-      <span
```
gabescarbrough, line 8: What caused the changes in the snapshot?

gabescarbrough (approved):
Everything related to the goal of this PR looks good to me.

A few questions but nothing blocking.

There are also some `js-lint` warnings in the Files Changed tab for these files. We should resolve those.

src/Tabs/__snapshots__/Tabs.test.tsx.snap
kyleshike: It's from switch from styled components to css modules.  Previously, the tabs were styled by this span:
```js
export const StyledTabsWrapper = styled.span
```

jasonbasuil:
I think the behavior in the stories changed where the tab content isn't switching.

Docs:

https://github.com/user-interviews/ui-design-system/assets/37383785/684aa2c6-caec-4f10-ba6e-88d1737c4312


This PR:


https://github.com/user-interviews/ui-design-system/assets/37383785/eb4f16b2-64bf-4cd1-9d05-8085ef8e352f

jasonbasuil (approved):
Just the comment about the story behavior. Also, we could probably remove `styled-components` since it's no longer used and we don't want it included anymore.

src/Tabs/__snapshots__/Tabs.test.tsx.snap
gabescarbrough: Ah ok. That makes sense.

src/Tabs/Tabs.tsx
kyleshike: Yea that's exactly right. In this case specifically, the remaining `props` are straight from `BootstrapTabProps`:
Screenshot 2023-10-03 at 4 19 12 PM

PR #1043, opened 2023-10-12
re-write components without styled-components

### LINKED ISSUE #1041: Re-write Flex without styled-components

jasonbasuil:
closes #1041

Removing `styled-components` as there are some compatibility issues when using SSR on RS. This just re-writes a couple components to use CSS modules instead.

Chromatic: https://62d040e741710e4f085e0647-wmozfzrxln.chromatic.com/?path=/story/layouts-flex--default


src/Flex/Flex.module.scss
```diff
@@ -0,0 +1,125 @@
...
+  display: flex;
+}
+
+.align-items-stretch {
+  align-items: stretch;
+}
+
+.align-items-center {
+  align-items: center;
+}
```
jasonbasuil, line 15: I wanted to stick to the convention of using pascalCase with CSS modules here, but there were so many properties and it was a bit of an eye sore to parse and it seemed easier to read with Kebabcase

src/Flex/Flex.stories.tsx
```diff
@@ -25,7 +25,7 @@ export const Default: Story = {
     flexDirection: 'row',
     justifyContent: 'space-between',
   },
-  render: ({ children, ...args }) => (
```
jasonbasuil, line 28: gets rid of the unused children warning

src/Popover/PopoverCard.tsx
```diff
@@ -0,0 +1,36 @@
...
+  children?: ReactNode;
+  size?: 'sm' | 'md';
+}
+
+const PopoverCard = ({
+  children,
+  className,
+  size = 'sm',
+}: PopoverCardProps) => {
+  const widthClass = handleWidth(size);
```
jasonbasuil, line 27: we use PopoverCard once in RS. Verified it still works fine:

https://github.com/user-interviews/ui-design-system/assets/37383785/ecb5e6cd-7605-4a88-b9da-9f2eff80df78

src/Flex/Flex.tsx
kyleshike, line 1: I feel like we should revisit this component at some point.  Its a very thin wrapper over flexbox styles. I'm not sure that it really abstracts anything or adds any guardrails, etc.

kyleshike (approved):
Looks great.  Left a note on the flex component, but not blocking.

src/Popover/PopoverCard.tsx
careful-axe: I think we might want to use this for segments soon as well.

src/Flex/Flex.tsx
jasonbasuil: Yeah it is pretty simple, but if I can center basic stuff without more CSS that's a win.  Can revisit to make it more robust down the road.

careful-axe (approved):
Thanks for keeping me in the loop!

PR #1049, opened 2023-10-23
Fix tabs styles in RS

kyleshike:
Rails server includes bootstrap _after_ the DS, which means that bootstrap's own styles take precedence over our own that are built on top. By doubling up the class name selectors here, we ensure that our styles take precedence.

**This should be treated as an interim solution.** 

Ultimately, we should have the DS own bootstrap & react-bootstrap altogether, with rails server only consuming BS _through_ the DS.

### Before:
Screenshot 2023-10-23 at 11 23 15 AM

**Active tab state getting overridden**
Screenshot 2023-10-23 at 11 24 24 AM

**flexWrapUnset getting overidden** 
Screenshot 2023-10-23 at 11 25 18 AM

### After:
Screenshot 2023-10-23 at 11 23 42 AM

### DIFF
--- a/src/Tabs/tabs.module.scss
+++ b/src/Tabs/tabs.module.scss
@@ -1,6 +1,6 @@
 @import '../../scss/theme.scss';
 
-.tabs {
+.tabs.tabs {
   --border-width: 0.125rem;
 
   border-bottom: var(--border-width) solid $ux-gray-400;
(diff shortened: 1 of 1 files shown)


jasonbasuil (approved):
Thanks! More reason for the DS to own bootstrap.

PR #1056, opened 2023-11-01
add trailingText to Input component

jasonbasuil:
Adds `trailingText` to `Input`

Figma: https://www.figma.com/file/wU1NfHuJchd9xLirqBREQN/Research-Set-up?type=design&node-id=3569-7744&mode=design&t=aB8GD4SuTxyA5Mbs-0

Chromatic: https://62d040e741710e4f085e0647-nsyrdovjii.chromatic.com/?path=/story/components-form-elements-form-group--with-trailing-text

If `<Input type='number'`, add spacing for chevrons

https://github.com/user-interviews/ui-design-system/assets/37383785/bf820320-14a2-469e-9309-4ba4d1f5fd7e

If not `type='number'`, no spacing adjustment
Screenshot 2023-11-01 at 1 39 36 PM

### DIFF
--- a/src/Input/Input.jsx
+++ b/src/Input/Input.jsx
@@ -16,6 +16,7 @@ const Input = React.forwardRef((props, ref) => {
     trailingIconLabel,
     trailingIconOnClick,
     trailingIconOnClickSubmit,
+    trailingText,
     type,
     value,
     onChange,
@@ -47,6 +48,11 @@ const Input = React.forwardRef((props, ref) => {
         onChange={onChange}
         {...rest}
       />
+      {trailingText && (
+        <span className={`input-trailing-text ${type === 'number' ? 'input-trailing-text--input-type-number' : ''}`}>
+          {trailingText}
+        </span>
+      )}
       {(trailingIcon && trailingIconOnClick) && (
         <button
           aria-label={trailingIconLabel}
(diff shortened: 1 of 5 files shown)


src/Input/Input.scss
```diff
@@ -24,6 +26,19 @@
     background-color: $ux-gray-100;
   }
 
+  .input-trailing-text {
+    position: absolute;
+    right: $ux-spacing-40;
+    top: 50%;
+    transform: translateY(-50%);
+    pointer-events: none; 
+    z-index: 3;
```
jasonbasuil, line 35: bootstrap oddly adds a z-index: 3 so just adding here so it doesn't disappear on focus
Screenshot 2023-11-01 at 1 44 23 PM

src/Input/Input.jsx
```diff
@@ -47,6 +48,11 @@ const Input = React.forwardRef((props, ref) => {
         onChange={onChange}
         {...rest}
       />
+      {trailingText && (
```
domschab23, line 51: Any chance we can make this a little more bulletproof on smaller screens? 

https://github.com/user-interviews/ui-design-system/assets/54962889/bf04d4b3-63a3-4a91-bd56-17d382c4017f

jasonbasuil: made it `display: none` when reaching `200px`. Kind of an arbitrary length and doesn't really account for what the placeholder text might be. Lmk what you think.

https://github.com/user-interviews/ui-design-system/assets/37383785/e7fb27b6-ecc9-40b8-af10-900018bae0e5

domschab23: @natkasman any opinions here?

natkasman: I think this is fine considering the smallest device sizes is usually 300ish in width!

domschab23 (approved):

PR #1057, opened 2023-11-07
Allow drawer to be expanded by default

eric2523:
### Description
This is a feature that us at rx-workflow wanted as the contents that we were putting in was too big for the non-expanded state (needed to put the Calendar component inside). With that we wanted the drawer to be able to be expanded by default while not allowing for the user to click on a expanded button.

### Changes
Creates a new prop `defaultExpanded`, which I chose the naming based off react-select's `defaultMenuIsOpen` naming (open to changes tho), that will set the initial value of the expanded state

### DIFF
--- a/src/Drawer/Drawer.mdx
+++ b/src/Drawer/Drawer.mdx
@@ -15,7 +15,7 @@ import * as ComponentStories from './Drawer.stories';
 
 <Canvas of={ComponentStories.Default} />
 
-### When to use 
+### When to use
 When we want to focus the user’s attention on an interactive interface or a large amount of information
 
 ### When to not use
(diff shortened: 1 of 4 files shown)


src/Drawer/Drawer.stories.jsx
```diff
@@ -37,6 +37,7 @@ export const Default = () => {
       <Button onClick={toggleVisible}>Open</Button>
       <Drawer
         behindNav={(boolean('behindNav', true))}
+        defaultExpanded={boolean('defaultExpanded', false)}
```
eric2523, line 40: I think we want this for all these examples so our knobs work for every example?

jasonbasuil (approved):
LGTM!

src/Drawer/Drawer.mdx
```diff
@@ -50,7 +50,11 @@ Pass `orientation="left"` to override the default behavior of sliding in from th
 
 ### Expandable
 
-Setting `expandable` allows the drawer to expand to the full size of the viewport
+Setting `expandable` allows the drawer to expand to the full size of the viewport when the expand button is clicked.
+
+### Default Expanded
+
+Setting `defaultExpanded` allows the drawer to expand the full size of the viewport by default.
```
gabescarbrough, line 57: **nitpick**: 💅 

We might want to call out that `defaultExpanded` can be used whether `expandable` is true or false, to make it clear that you can use it for make an always full-width drawer, or one that opens full-width but can be minimized to the `size` specified.

gabescarbrough (approved):
Looks great! Thanks for doing this.

Just one minor suggestion on the docs.

PR #1062, opened 2023-11-10
use Inter font

### LINKED ISSUE #1061: [UIDS] - Change font to Inter
---
title: "[UIDS]"
assignees: ''

---

## Description

## User Stories

## Acceptance Criteria

- [Criteria 1]
- [Criteria 2]
- [Criteria 3]

jasonbasuil:
closes #1061 

- Updates font to `Inter`
- Pares down the available font types showcased in the Typography stories (similar to Synthesis DS in [Figma](https://www.figma.com/file/dmmTbSjuPXqt9m82YaewXO/Synthesis-Design-System?type=design&node-id=11-1435&mode=design&t=T5ZmVEjRAUsQdWo7-0))
- Uses `synth` namespace for the Synthesis Design System

https://62d040e741710e4f085e0647-xfgczigaif.chromatic.com/?path=/docs/foundations-typography--docs

### DIFF
--- a/scss/baseline.scss
+++ b/scss/baseline.scss
@@ -2,7 +2,7 @@
 @import './typography';
 
 body {
-  font-family: $font-family;
-  font-size: $font-size-base;
+  font-family: $synth-font-family;
+  font-size: $synth-font-size-base;
   color: $ux-gray-900;
 }
--- a/scss/forms/form_control_label.scss
+++ b/scss/forms/form_control_label.scss
@@ -3,14 +3,14 @@
 @import '../typography';
 
 .FormControlLabel {
-  @include font-type-30;
+  @include synth-font-type-30;
   align-items: center;
   color: $ux-gray-900;
   display: flex;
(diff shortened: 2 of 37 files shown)


kyleshike (commented):
Does this require updating all use of `font-type` vars in rails server? If so, what do you think about supporting both the old and the new variables until they're all ported over in RS? We can remove the old vars with the major release since it's a breaking change

jasonbasuil:
> Does this require updating all use of `font-type` vars in rails server? If so, what do you think about supporting both the old and the new variables until they're all ported over in RS? We can remove the old vars with the major release since it's a breaking change

Perhaps I'm just over optimistic that I could get these all replaced in one go, but supporting both probably is a good idea at least for a first pass (and it is likely others may be using the old ones while the new ones get swapped). 

I can add the old ones back to reference the new ones so we can support both.

kyleshike (approved):
🚀

PR #1063, opened 2023-11-13
Use modal backdrop z index on drawer

kyleshike:
### Before:

https://github.com/user-interviews/ui-design-system/assets/3161151/647112d1-f3ad-4180-9d62-23e0942ac7ba

### After:

https://github.com/user-interviews/ui-design-system/assets/3161151/2620f7f2-f5e0-4214-9f92-c2bf336fd374

### DIFF
--- a/src/Drawer/Drawer.scss
+++ b/src/Drawer/Drawer.scss
@@ -7,7 +7,7 @@
   bottom: 0;
   left: 0;
   background: black;
-  z-index: 1000;
+  z-index: $z-index-modal-backdrop;
   opacity: 0%;
   pointer-events: none;
   transition: 200ms linear;


gabescarbrough (approved):

gabescarbrough:
This might be a separate issue but if we're changing z-indexes the drawer's z-index should probably be 1030 [based on this discussion](https://userinterviews.slack.com/archives/C6ZH961V5/p1697046842587979)

kyleshike:
@gabescarbrough I think we actually want to leave it as is for the new navigation, where both the drawer and its backdrop should be on top of the sidebar (or mobile header/footer) navigation

PR #1068, opened 2023-11-20
remove window from Avatar classname

jasonbasuil:


domschab23 (approved):
Verified with local build! 🚢

PR #1069, opened 2023-11-20
Add Synthesis colors from Figma

gabescarbrough:
Adds Sass variables for the [new Synthesis colors from Figma](https://www.figma.com/file/dmmTbSjuPXqt9m82YaewXO/Synthesis-Design-System?type=design&node-id=0-1&mode=design&t=lkmh9m9vVaDMtER6-0)

Todo:
- [x] Discuss and finalize color names
- [ ] Add new colors to docs
- [ ] Mark legacy colors not included in Brand Colors as deprecated in docs

### DIFF
--- a/scss/colors/synthesis-palette.scss
+++ b/scss/colors/synthesis-palette.scss
@@ -1,23 +1,24 @@
 // stoplight color system
-$synth-green-1-success: #015029;
-$synth-green-2-success: #09914f;
-$synth-green-3-success: #d4f6e5;
+$synth-success-green-dark: #015029;
+$synth-success-green-medium: #09914f;
+$synth-success-green-light: #d4f6e5;
 
-$synth-red-1-error: #c71f1f;
-$synth-red-2-error: #e03131;
-$synth-red-3-error: #fae1e1;
+$synth-error-red-dark: #c71f1f;
+$synth-error-red-medium: #e03131;
+$synth-error-red-light: #fae1e1;
 
-$synth-yellow-1-warn: #fbda36;
-$synth-yellow-2-warn: #ffe356;
-$synth-yellow-3-warn: #fff1ab;
+$synth-warning-amber-dark: #fbda36;
+$synth-warning-amber-medium: #ffe356;
+$synth-warning-amber-light: #fff1ab;
+$synth-warning-amber-alternate: #f59c27;
 
 // neutrals
-$synth-alert-bg-gray: #f4f4f4;
-$synth-div-stroke-gray: #dedede;
-$synth-header-gray: #fbfbfb;
-$synth-page-white: #fdfdfd;
-$synth-text-black: #1b1b1b;
-$synth-unselected-gray: #5b5b5b;
+$synth-alert-bg-neutral: #f4f4f4;
+$synth-div-stroke-neutral: #dedede;
+$synth-header-neutral: #fbfbfb;
+$synth-page-neutral: #fdfdfd;
+$synth-text-neutral: #1b1b1b;
+$synth-unselected-neutral: #5b5b5b;
 
 // main schema - visual hierarchy
 $synth-accent-green: #158d71;
@@ -29,8 +30,7 @@ $synth-primary-cta-blue: #162c4e;
 $synth-selected-state-green: #c6e2db;
 
 // feedback and signifier colors
-$synth-hyperlink: #3f6dca;
-$synth-orange-1-warn: #f59c27;
+$synth-hyperlink-color: #3f6dca;
 
 $synth-hover-blue: #e5ecff;
 $synth-header-gray-blue: #eef2f8;
--- a/scss/colors/synthesis-palette.scss
+++ b/scss/colors/synthesis-palette.scss
@@ -1,0 +1,36 @@
+// stoplight color system
+$synth-green-1-success: #015029;
+$synth-green-2-success: #09914f;
+$synth-green-3-success: #d4f6e5;
+
+$synth-red-1-error: #c71f1f;
+$synth-red-2-error: #e03131;
+$synth-red-3-error: #fae1e1;
+
+$synth-yellow-1-warn: #fbda36;
+$synth-yellow-2-warn: #ffe356;
+$synth-yellow-3-warn: #fff1ab;
+
+// neutrals
+$synth-alert-bg-gray: #f4f4f4;
+$synth-div-stroke-gray: #dedede;
+$synth-header-gray: #fbfbfb;
+$synth-page-white: #fdfdfd;
+$synth-text-black: #1b1b1b;
+$synth-unselected-gray: #5b5b5b;
+
+// main schema - visual hierarchy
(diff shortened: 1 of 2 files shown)


scss/colors/synthesis-palette.scss
```diff
@@ -0,0 +1,36 @@
+// stoplight color system
```
gabescarbrough, line 1: All variable names are currently what is in Figma just formatted for Sass and with the `synth` prefix

scss/colors/synthesis-palette.scss
```diff
@@ -0,0 +1,36 @@
...
+$synth-green-1-success: #015029;
+$synth-green-2-success: #09914f;
+$synth-green-3-success: #d4f6e5;
```
gabescarbrough, line 4:
```suggestion
$synth-success-green-dark: #015029;
$synth-success-green-medium: #09914f;
$synth-success-green-light: #d4f6e5;
```

To me, this convention is easier to know what you're getting from just looking at the variable name.

scss/colors/synthesis-palette.scss
```diff
@@ -0,0 +1,36 @@
...
+
+$synth-red-1-error: #c71f1f;
+$synth-red-2-error: #e03131;
+$synth-red-3-error: #fae1e1;
```
gabescarbrough, line 8:
```suggestion
$synth-error-red-dark: #c71f1f;
$synth-error-red-medium: #e03131;
$synth-error-red-light: #fae1e1;
```

scss/colors/synthesis-palette.scss
```diff
@@ -0,0 +1,36 @@
...
+
+$synth-yellow-1-warn: #fbda36;
+$synth-yellow-2-warn: #ffe356;
+$synth-yellow-3-warn: #fff1ab;
```
gabescarbrough, line 12:
```suggestion
$synth-warning-yellow-dark: #fbda36;
$synth-warning-yellow-medium: #ffe356;
$synth-warning-yellow-light: #fff1ab;
```

scss/colors/synthesis-palette.scss
```diff
@@ -0,0 +1,36 @@
...
+
+// neutrals
+$synth-alert-bg-gray: #f4f4f4;
```
gabescarbrough, line 15: Should we use `neutral` instead of `gray` to avoid the `gray` vs `grey` issue?

gabescarbrough: Do we even need the word `gray`? Could it just be `$synth-alert-bg-color`? Is that it is gray important?

scss/colors/synthesis-palette.scss
```diff
@@ -0,0 +1,36 @@
...
+$synth-div-stroke-gray: #dedede;
+$synth-header-gray: #fbfbfb;
+$synth-page-white: #fdfdfd;
```
gabescarbrough, line 18: If we did switch to CSS custom properties I could see this being `--synth-page-bg-color` since depending on light or dark mode it might not be white.

scss/colors/synthesis-palette.scss
```diff
@@ -0,0 +1,36 @@
...
+$synth-text-black: #1b1b1b;
```
gabescarbrough, line 19: Similar to above I could see this being `--synth-text-color` and being black or white depending on dark/light mode in the future.

scss/colors/synthesis-palette.scss
```diff
@@ -0,0 +1,36 @@
...
+$synth-navbar-blue: #010812;
+$synth-primary-cta-blue: #162c4e;
+$synth-selected-state-green: #c6e2db;
+
+// feedback and signifier colors
+$synth-hyperlink: #3f6dca;
+$synth-orange-1-warn: #f59c27;
+
+$synth-hover-blue: #e5ecff;
+$synth-header-gray-blue: #eef2f8;
```
gabescarbrough, line 36: These two were over to the side. Not 100% sure these are final colors.

scss/colors/synthesis-palette.scss
```diff
@@ -0,0 +1,36 @@
...
+$synth-accent-green: #158d71;
+$synth-dark-background-selected-blue: #162c4e;
+$synth-hover-state: #eef9f6;
+$synth-indicator-stroke-green: #91cabb;
+$synth-navbar-blue: #010812;
+$synth-primary-cta-blue: #162c4e;
+$synth-selected-state-green: #c6e2db;
+
+// feedback and signifier colors
+$synth-hyperlink: #3f6dca;
```
gabescarbrough, line 32:
```suggestion
$synth-hyperlink-color: #3f6dca;
```

Might want to be more specific?

scss/colors/synthesis-palette.scss
```diff
@@ -0,0 +1,36 @@
...
+$synth-orange-1-warn: #f59c27;
```
gabescarbrough, line 33: This is an alternate of the stoplight yellow for a specific situation where orange looks better than yellow.

I could see a case for calling those amber instead (which includes yellow and orange) and calling this something like `$synth-warning-amber-alternate`

scss/colors/synthesis-palette.scss
jasonbasuil: I think we can keep this naming convention for now as it makes sense to me. As we talked about, as long as we have some primitives for these colors that get assigned to the semantic ones currently shown here we should be fine. We may end up adjusting or adding colors later to account for a sliding scale for brightness / a11y to make the system more logical on the primitive level.

`$synth-green-1` or `$synth-green-100`. I'm not too picky at this point, just as long as the primitive exists.

scss/colors/synthesis-palette.scss
jasonbasuil: I would be fine using `neutral`. Just looking at our current pallete.scss it does include white as well as black.

Screenshot 2023-11-21 at 11 42 06 AM

scss/colors/synthesis-palette.scss
jasonbasuil: should be fine to keep

scss/colors/synthesis-palette.scss
jasonbasuil: sounds fine to me

jasonbasuil (approved):
Echoing what Kyle said. Let's 🚀 and start getting all our components Synthesized™️

scss/colors/synthesis-palette.scss
gabescarbrough: > I think we can keep this naming convention for now as it makes sense to me.

Keep the original naming convention or the one I suggested?

jasonbasuil: We can keep your suggested one!

gabescarbrough: Cool. Just wanted to make sure I was reading that correctly before I made changes.

PR #1071, opened 2023-11-21
add border to Card under Synthesized class

### LINKED ISSUE #1070: remove elevations from Card

jasonbasuil:
closes #1070 

Adds border to `Card` under the `.Synthesized` class. This should only affect the `Card` component when `.Synthesized` class is applied to the `<body>` when `@is_new_navigation_enabled` is enabled.

Looks like this (removing the border already applied in `app/javascript/researcher/account_project_dashboard/account_project_dashboard.scss`):
 
Screenshot 2023-11-21 at 1 18 32 PM

Screenshot 2023-11-21 at 1 18 45 PM

When `@is_new_navigation_enabled` is off:

Screenshot 2023-11-21 at 1 21 18 PM

### DIFF
--- a/src/Card/Card.scss
+++ b/src/Card/Card.scss
@@ -1,5 +1,12 @@
 @import '../../scss/theme';
 
+.Synthesized {
+  .Card {
+    border: 1px solid $ux-gray-400;
+    box-shadow: none;
+  }
+}
+
 .Card {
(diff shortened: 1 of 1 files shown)


kyleshike (approved):
we-all-live-in-computer-world-computer-world

PR #1091, opened 2023-12-05
update DropdownItem hover state, remove danger variant

### LINKED ISSUE #1090: [UIDS] - Update DropdownItem hover state, remove danger variant
---
title: "[UIDS]"
assignees: ''

---

## Description

## User Stories

## Acceptance Criteria

- [Criteria 1]
- [Criteria 2]
- [Criteria 3]

jasonbasuil:
closes #1090 

- Update the `DropdownItem` hover state to synth hover color (also uses CSS modules to avoid bootstrap collision)
- removes danger variant (we only use in one place currently, but PD says we can get rid of it)

Before:

https://github.com/user-interviews/ui-design-system/assets/37383785/b009f61f-fa5e-48ff-90f7-4d753794ecec

After:

https://github.com/user-interviews/ui-design-system/assets/37383785/a98c28d4-9aff-4ab4-9694-1b20bffc7288

### DIFF
--- a/src/Dropdown/DropdownItem.jsx
+++ b/src/Dropdown/DropdownItem.jsx
@@ -36,7 +36,7 @@ const DropdownItem = ({
     onClick={onClick}
     {...props}
   >
-    { leadingIcon && <FontAwesomeIcon className="icon-left fa-fw" icon={leadingIcon} /> }
+    { leadingIcon && <FontAwesomeIcon className={classNames(styles.DropdownItemIcon, 'icon-left', 'fa-fw')} icon={leadingIcon} /> }
     { children }
   </RBDropdown.Item>
   );
--- a/src/Dropdown/DropdownItem.jsx
+++ b/src/Dropdown/DropdownItem.jsx
@@ -8,10 +8,6 @@ import { Dropdown as RBDropdown } from 'react-bootstrap';
 
 import styles from './DropdownItem.module.scss';
 
-const DropdownItemVariants = {
-  DANGER: 'danger',
-};
-
 const DropdownItem = ({
   active,
   as,
@@ -22,7 +18,6 @@ const DropdownItem = ({
   href,
   leadingIcon,
   onClick,
-  variant,
   bsPrefix,
   ...props
 }) => (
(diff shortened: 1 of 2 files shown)


kyleshike (commented):
Looks great, but there's a question about release strategy we'll want to consider here

gabescarbrough:
> Looks great, but there's a question about release strategy we'll want to consider here

That's my one point of feedback as well. Should this only be styled the new way if wrapped in `.Synthesized`? Or is this how we want it to look on all screen regardless?

Also dropping a variant is a breaking change. According to semantic versioning I'd expect it to be marked as deprecated in the docs and dropped in the next major version.

jasonbasuil:
> > Looks great, but there's a question about release strategy we'll want to consider here
> 
> That's my one point of feedback as well. Should this only be styled the new way if wrapped in `.Synthesized`? Or is this how we want it to look on all screen regardless?
> 
> Also dropping a variant is a breaking change. According to semantic versioning I'd expect it to be marked as deprecated in the docs and dropped in the next major version.

Good callout, totally forgot about that! Just dropped the Synthesized class in and will work for the old hover states. We don't have any docs on the danger variant, but we can def bump up the version on the next release.

gabescarbrough:
> We don't have any docs on the danger variant, but we can def bump up the version on the next release.

Since the `danger` variant isn't documented and only used in one place, I think it's fair to just bump a minor version then. The more I think about it, it isn't _really_ a breaking change since it will just fall back to the normal styles which is what we want.

gabescarbrough (approved):

jasonbasuil:
Opting to remove the `.Synthesized` class. Had to double check, but when both the Synthesized version and the non-Synthesized version were present, even if the `.Synthesized` was present on the body, it would still choose the non-Synthesized version. Once I got rid of the non-Synthesized version, then bootstrap took over with their defaults. 

Seeming as how this will get released to everyone next week, I think it's okay to just have this as is. 

scoped with `.Synthesized`

Screenshot 2023-12-11 at 11 48 09 AM

not scoped

Screenshot 2023-12-11 at 11 49 07 AM

PR #1096, opened 2023-12-07
give default sizes to levels, make weight and size optional

### LINKED ISSUE #1095: [UIDS] Give default sizes to Heading
---
title: "[UIDS]"
assignees: ''

---

## Description

## User Stories

## Acceptance Criteria

- [Criteria 1]
- [Criteria 2]
- [Criteria 3]

jasonbasuil:
closes #1095 

Chromatic: https://62d040e741710e4f085e0647-xcmihmyrti.chromatic.com/?path=/story/components-heading--sizes

- Makes only `level` and `size` required on `Heading`
- Changes default level from `2` to `1`
- Assigns default `size` props to every Heading `level`
- If `size` or `weight` are assigned, they will overtake the defaults 

I think this should simplify the need to assign each one of these props on the `Heading` component every time. Given a required `level`, it should be able to default to the a more logical sizing based on level. A dev can always override the size or weight based on visual hierarchy needs, but hopefully this reduces the friction to using the `Heading` component. 

Primary: Kyle, Gabe

### DIFF
--- a/src/Heading/Heading.tsx
+++ b/src/Heading/Heading.tsx
@@ -8,16 +8,27 @@ export interface HeadingProps {
   children?: ReactNode;
   className?: string;
   level: 1 | 2 | 3 | 4 | 5 | 6;
-  size: 'xs' | 'sm' | 'md' | 'lg' | 'xl' | 'xxl' | 'xxxl';
+  size?: 'xs' | 'sm' | 'md' | 'lg' | 'xl' | 'xxl' | 'xxxl';
   textAlign?: 'left' | 'center' | 'right';
-  weight: 'regular' | 'medium' | 'bold';
+  weight?: 'regular' | 'medium' | 'bold';
 }
+
+const LevelToSizeMap = {
+  1: 'xxxl',
+  2: 'xxl',
+  3: 'xl',
+  4: 'lg',
+  5: 'md',
+  6: 'sm',
+};
+
+const getHeadingSizeClassFromLevel = (level: number) => `Heading--${LevelToSizeMap[level]}`;
 
 const Heading = ({
   children,
   className,
-  level = 2,
-  size = 'xxl',
+  level = 1,
+  size,
   textAlign,
   weight = 'bold',
   ...props
@@ -28,6 +39,7 @@ const Heading = ({
       className: classNames(
         className,
         'Heading',
+        !size && getHeadingSizeClassFromLevel(level),
         {
           [`Heading--${size}`]: !!size,
           [`Heading--${weight}`]: !!weight,
--- a/src/Heading/Heading.tsx
+++ b/src/Heading/Heading.tsx
@@ -7,8 +7,22 @@ import './Heading.scss';
 export interface HeadingProps {
   children?: ReactNode;
   className?: string;
+  /**
+   * @type {number}
+   * @description Choose the appropriate heading level (h1 to h6) based on a
+   * page's semantic structure.
+   * Headings define the document structure, and screen readers use them for navigation.
+   * For accessibility, a page must contain at least one h1, headings should follow a
+   * logical hierarchy where each heading level represents a sublevel of the previous one,
+   * and avoid skipping levels.
+   */
   level: 1 | 2 | 3 | 4 | 5 | 6;
-  size?: 'xs' | 'sm' | 'md' | 'lg' | 'xl' | 'xxl' | 'xxxl';
+  /**
+   * @type {string}
+   * @description Sizes map to the available font-sizes from the defined list of font-types.
+   * Adjust for visual hierarchy.
+   */
+  size: 'xs' | 'sm' | 'md' | 'lg' | 'xl' | 'xxl' | 'xxxl';
   textAlign?: 'left' | 'center' | 'right';
   weight?: 'regular' | 'medium' | 'bold';
 }
--- a/spec/__snapshots__/Storyshots.test.js.snap
+++ b/spec/__snapshots__/Storyshots.test.js.snap
@@ -9054,8 +9054,8 @@ Array [
 `;
 
 exports[`Storyshots Components/Heading Default 1`] = `
-<h2
-  className="Heading Heading--xxl Heading--bold"
+<h1
+  className="Heading Heading--xxxl Heading--bold"
   style={
     Object {
       "textAlign": undefined,
@@ -9063,7 +9063,7 @@ exports[`Storyshots Components/Heading Default 1`] = `
   }
 >
   The fastest way to recruit research participants
-</h2>
+</h1>
 `;
 
 exports[`Storyshots Components/Heading Levels 1`] = `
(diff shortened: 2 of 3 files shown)


src/Heading/Heading.tsx
```diff
@@ -8,16 +8,27 @@ export interface HeadingProps {
...
+};
+
+const getHeadingSizeClassFromLevel = (level: number) => `Heading--${LevelToSizeMap[level]}`;
+
 const Heading = ({
   children,
   className,
-  level = 2,
-  size = 'xxl',
+  level = 1,
```
jasonbasuil, line 30: Open to different opinions on this, but when I originally built this I would have thought `h2` would be the most commonly used level for heading. Current spread in the app looks like: 9 `h1`, 7 `h2`, 5 `h3`, 5 `h4`, 1 `h5`, so seems like it makes sense for the default to be `level={1}`? And perhaps in my head I logically think, oh I want a heading which will start at 1, but I can adjust those levels down if I need.

gabescarbrough: I'd also like to push back a bit on assigning default `size` props for every `level`. In my experience one of the biggest misconceptions among devs less experiences with front-end code is that the reason to use h1-h6 tags is to get different heading sizes. Assigning default sizes for this component reinforces that assumption if you don't know better.

Having to explicitly set both `level` and `size` is less convenient but it makes it clear that the heading level and its size aren't dependent on each other. It is a bit more friction, but it is also harder to do the wrong thing on accident.

gabescarbrough: Also worth mentioning that we don't have any usages of `level={6}`. Kind of interesting if not that surprising.

gabescarbrough: For comparison, h1-h6 elements that could be using `Heading`:

```
>  ~/R/r/a/javascript  main  rg '<h1' | wc -l
     152
>  ~/R/r/a/javascript  main  rg '<h2' | wc -l
     344
>  ~/R/r/a/javascript  main  rg '<h3' | wc -l
     391
>  ~/R/r/a/javascript  main  rg '<h4' | wc -l
     196
>  ~/R/r/a/javascript  main  rg '<h5' | wc -l
     203
>  ~/R/r/a/javascript  main  rg '<h6' | wc -l
      33
```

kyleshike: > Having to explicitly set both level and size is less convenient but it makes it clear that the heading level and its size aren't dependent on each other. 

While it's definitely true that size & level of headings aren't as a rule dependent on one another, I'm not sure I agree that this is the best choice for our API. If headings are consistently sized in _our design spec_ then we have an opportunity to put up some guardrails to ensure they are used consistently in the app.

To put it another way, I see this is an opportunity to provide sane defaults, while still providing the ability to specify size in whatever edge cases.

Alternatively if we didn't map levels to a default size, then devs need to consult with Synthesis' typography map and choose the right one each time.  This is more or less what we have currently and it results in design drift.

gabescarbrough: > If headings are consistently sized in our design spec then we have an opportunity to put up some guardrails to ensure they are used consistently in the api

That's a big "if" though. What does "headings are consistently sized in our design spec" even mean, in a practical sense? `level` is tied to the broader context of the page and the other headings on it and their semantic importance. `size` is a design decision based on visual hierarchy. I don't see a clear way to correlate them in an app context. For something like a blog post it makes sense for higher levels to be smaller because the headings are usually nested into sections of text. That isn't generally the case for headings used in user interfaces.

It's only a guardrail if it keeps you from going off track. I think there is a high likelihood that correlating `level` and `size` leads devs to not even thinking about what the right track is, and choosing `level` based on what `size` is in the design since `level` is required and `size` isn't.

> To put it another way, I see this is an opportunity to provide sane defaults, while still providing the ability to specify size in whatever edge cases.

I'm totally in favor of sane defaults. I'm open to changing my mind but I think this is a case where there isn't an obvious default and we would be setting ourselves up for the same trap devs fall into with h1-h6 where they make the wrong choice because it feels right.

> Alternatively if we didn't map levels to a default size, then devs need to consult with Synthesis' typography map and choose the right pairing each time.

I wouldn't expect devs to have to consult the typography map. You would refer to the mockups from design to know which `size` to use and the structure of the page you're working on for the `level`. 

> This is more or less what we have currently and it results in design drift.

Can you give an example of design drift caused by `level` and `size` being separate? I haven't personally seen wrong heading sizes being used. In my experience they've matched the size in the mockup.

The design drift I have seen is from custom CSS setting the size of h1-h6 elements which would be addressed by switching to Heading. There are way more uses of h1-h6 than Heading at the moment and most of them have their own CSS.

jasonbasuil: > I'd also like to push back a bit on assigning default size props for every level. In my experience one of the biggest misconceptions among devs less experienced with front-end is that the reason to use h1-h6 tags is to get different heading sizes. Assigning default sizes for this component reinforces that assumption if you don't know better.

I would agree because I used to be guilty of doing this and that a dev less familiar might end up choosing a level because of the size that it is given by default. I do see the value in if `size` remained required that at least the dev using it would know that the visual presentation of the heading is not tightly coupled with `level` and can be changed. 

I think inevitably we're not going to be perfect with heading levels on every single page and that's certainly not the case today (e.g. ensuring there is only 1 `h1` on a page / using `h2` - `h6` in descending order of importance and not skipping levels). It may be hard to predict how an app like ours with so many different moving pieces could follow the best practices on semantic heading structure (unless we had some automated a11y tests).

gabescarbrough:
I figured I should fact check myself on this statement, at least as far as `Heading` usages in our app

> I would bet that there is rarely a correlation between size and level in app user interfaces

### `level` + `size` combination counts

| level | size  | count |
| ------|-------|------ |
| 1     | xxxl  | 7     |
| 1     | xl    | 2     |
| 1     | default (xxl)  | 3     |
| 2     | xxxl  | 1     |
| 2     | xxl   | 1     |
| 2     | xl    | 4     |
| 2     | lg    | 2     |
| 2     | default (xxl)  | 2     |
| 3     | xxxl  | 1     |
| 3     | xxl   | 2     |
| 3     | lg    | 1     |
| 3     | default (xxl)  | 9     |
| 4     | xl    | 1     |
| 4     | lg    | 2     |
| 4     | md    | 2     |
| 4     | default (xxl)  | 5     |
| 5     | md    | 1     |

### Most common `size` per `level`
(including default with corresponding size)

| level | size  |
| ------|-------|
| 1     | xxxl  |
| 2     | xl    |
| 3     | xxl   |
| 4     | xxl   |
| 5     | md    |

We should take this with a grain of salt because 49 usages is a small sample size but there are some clear takeaways imo:
1. The most common sizes per level don't align with the mappings in this PR
2. The idea that a lower level means a bigger most common size is wrong 2/5 of the time
3. Most levels are used with 3-4 sizes that overlap, with many ties for second place

I don't think there is a strong correlation between `level` and `size` in our usages.

Analyzing the vanilla HTML/CSS heading tags to see the most common font sizes there would be more telling because it is a much bigger sample size at 27x more usages.

jasonbasuil:
> I don't think there is a strong correlation between level and size in our usages.

I would agree. Thanks for checking on the spread of the usages @gabescarbrough . It's definitely highly variable with levels and sizes. I see the `Heading` with defaults as just a starting point when using it and more often than not `size` will be changed. I don't have a super strong opinion if we feel like this detracts more from the developer experience or may be a trap for using the wrong level based on the default size shown at first. I'm up for whatever makes more sense based on these conversations. 

At the end of the day, more `Heading` components in usage is a win vs. plain HTML headings (which don't use our font types if not set in CSS).

src/Heading/Heading.tsx
kyleshike: I think this PR shouldn't be blocked any further on what has turned into a debate.  @jasonbasuil can you just re-request review when you've made whatever changes you feel are best here?

src/Heading/Heading.tsx
```diff
@@ -7,17 +7,42 @@ import './Heading.scss';
...
   className?: string;
+  /**
+   * @type {number}
+   * @description Choose the appropriate heading level (h1 to h6) based on a
+   * page's semantic structure.
+   * Headings define the document structure, and screen readers use them for navigation.
+   * For accessibility, a page must contain at least one h1, headings should follow a
+   * logical hierarchy where each heading level represents a sublevel of the previous one,
+   * and avoid skipping levels.
+   */
```
jasonbasuil, line 18: Wanted to add some description notes just for some extra context for devs who might need a reminder that level is not tightly coupled with size. Also adds some a11y notes.

Screenshot 2023-12-12 at 12 24 15 PM

src/Heading/Heading.tsx
```diff
@@ -7,17 +7,42 @@ import './Heading.scss';
...
   level: 1 | 2 | 3 | 4 | 5 | 6;
+  /**
+   * @type {string}
+   * @description Sizes map to the available font-sizes from the defined list of font-types.
+   * Adjust for visual hierarchy.
+   */
```
jasonbasuil, line 24: Screenshot 2023-12-12 at 12 24 29 PM

kyleshike (approved):
Let's keep an eye on usages of this component.  I'm willing to bet that we can still abstract away some sane defaults for size.

src/Heading/Heading.tsx
```diff
@@ -28,6 +53,7 @@ const Heading = ({
       className: classNames(
         className,
         'Heading',
+        !size && getHeadingSizeClassFromLevel(level),
```
gabescarbrough, line 56: I realize this has been merged already but just want to make sure I'm understanding correctly.

Now that `size` is required and doesn't have a default, will `getHeadingSizeClassFromLevel` ever be used? Wouldn't you get a type error if `size` isn't provided?

jasonbasuil: Yeah you will get a type error. I guess it's only getting used if `size` isn't present to set it initially, then forces you to set the size. Guess it defeats the purpose of having that function if it's required, but we can see how it plays out and change it back to optional if we want that abstraction back.

PR #1106, opened 2023-12-28
adds max height to template variable popup

kyleshike:
When the variable list is too long, it would overflow the window and the bottom items become inaccessible.  This just sets a reasonable max-height.  This is sort of a hot fix and the popup should be updated to calculate it's height based on its position, but for now this does the trick.


brianCollinsUI (approved):
lgtm!

PR #1115, opened 2024-01-02
update Toast dismiss timing

### LINKED ISSUE #1114: [UIDS] - Update Toast dismiss timing
---
title: "[UIDS]"
assignees: ''

---

## Description

## User Stories

## Acceptance Criteria

- [Criteria 1]
- [Criteria 2]
- [Criteria 3]

jasonbasuil:
closes #1114 

Chromatic: https://62d040e741710e4f085e0647-dgnqhmugfu.chromatic.com/?path=/story/components-toast--default

- Sets Success variant dismiss timing to 3 seconds. This should make sense for Success messages as they tend to be shorter in word length and require less time to read (e.g. "Message successfully sent", "Email template saved successfully." I read in a couple places that a success toast can vary between 2-5 seconds so we've probably been on the longer end. Nat thought 3 seconds was good length.
- Keeps all other Toast variants at 5 seconds by default as users may need a bit more time to read and comprehend these messages. 

We could probably refine more, but this seems to solve the issue Nat described in Workflow.

Screenshot 2024-01-02 at 2 51 14 PM

### DIFF
--- a/src/Alert/Alert.jsx
+++ b/src/Alert/Alert.jsx
@@ -52,7 +52,8 @@ const getAlertIcon = (type) => {
   }
 };
 
-const AUTO_DISMISS_TIMEOUT = 5000;
+const AUTO_DISMISS_TIMEOUT_SUCCESS = 3000;
+const AUTO_DISMISS_TIMEOUT_DEFAULT = 5000;
 
 const getAlertClassName = (type) => {
   if (!Object.values(MessageTypes).includes(type)) {
@@ -68,12 +69,14 @@ function Alert(props) {
   useEffect(() => {
     let timeout;
     if (autoDismiss) {
-      timeout = setTimeout(() => (onDismiss(id)), AUTO_DISMISS_TIMEOUT);
+      timeout = setTimeout(() => (onDismiss(id)),
+      props.type === MessageTypes.SUCCESS ?
+      AUTO_DISMISS_TIMEOUT_SUCCESS : AUTO_DISMISS_TIMEOUT_DEFAULT);
     }
     return () => {
       clearTimeout(timeout);
     };
-  }, [autoDismiss, onDismiss, id]);
+  }, [autoDismiss, onDismiss, id, props.type]);
 
   return (
     <div


gabescarbrough:
I'm really curious what ChatGPT's source is for this. I haven't been able to find similar advice while searching.

gabescarbrough:
I'd prefer all autodismissable toasts display for the same amount of time.

In the context of something displaying on screen, 2 seconds seems like a big difference to me. If you're used to seeing mostly success toasts, I think the other types would feel like they are slow or lagging in comparison.

I don't have a strong take on how long that should be. Looking at other design systems it seems like the default is usually 5 seconds for all types of toasts, and most offer the option to set an override which I personally don't like. I'd prefer consistency.

If we think our current toasts are on screen for too long then I'd be cool with all toasts being 3 seconds.

It might make sense for `error` toasts to not be autodismissable. There's nothing worse than trying to get a screenshot of an error and it disappearing.

That said, I'm not trying to block this if you guys agree with ChatGPT. I'm just wary of trusting AI on UX best practices unless we can find a reputable source with the same advice.

jasonbasuil:
> > wary of trusting AI on UX best practices unless we can find a reputable source with the same advice.
> 
> This is completely fair.

This is valid point. I did try to ask where the response's source was from but direct access to training data isn't available. I guess the other places where I saw some of the best practices came from other design systems or blog posts. As far as reputable, I would have expected Nielsen Norman to have something on the topic but couldn't find a direct answer.

Here's one from the [UX Movement blog on 3-5 sec](https://uxmovement.com/forms/best-practices-to-make-your-toasts-usable-and-accessible/#:~:text=Also%2C%20toasts%20will%20typically%20display%20for%202%2D5%20seconds.%20A%20time%20frame%20of%203%20seconds%20is%20ideal%20for%20most%20users.)

Though [HP's design system](https://design-system.hpe.design/templates/toast-notifications#:~:text=Toasts%20are%20intended%20to%20be%20non%2Dintrusive%20and%20brief.%20A%20user%20must%20be%20able%20to%20read%20and%20comprehend%20a%20toast%20message%20in%20~8%20seconds%20before%20the%20toast%20automatically%20closes%20itself.) recommends 8 seconds, so this could be highly opinionated.

> It might make sense for error toasts to not be auto-dismissable. There's nothing worse than trying to get a screenshot of an error and it disappearing.

I do agree that it would be good to have consistency for the timing, and Nat did mention that she was curious about the warning/errors toasts being dismissible by default. I think having the 3 seconds for success solves Nat's issue, and keeping the 5 seconds for the other existing toasts that are auto-dismissable for the time being until we make a decision on dismissible defaults, just so we don't shorten existing error or warning toasts that disappear and you need to screenshot like you'd mentioned @gabescarbrough

PR #1121, opened 2024-01-05
Converts RTE to typescript, replaces templateVariables prop with extensions prop

kyleshike:
Template variables are a very rails-server specific concept. This PR removes them from this repo altogether, instead allowing for our RichTextEditor to accept an array of extensions which are passed directly to the TipTap editor. In support of this, this PR also converts the RTE to typescript, to provide some safeguards around the validity of extensions passed to it.

https://github.com/user-interviews/rails-server/pull/22794 provides an example of how using this new prop can be used in rails server.


src/RichTextEditor/RichTextEditor.stories.jsx
```diff
@@ -53,15 +53,6 @@ export const OneLine = () => (
   />
 );
 
-export const TemplateVariables = () => (
```
kyleshike, line 56: Removed the template variables story from storybook. I opted not to add an extensions story as it seemed a bit contrived. Adding extensions should function exactly as described in TipTap's documentation.

src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -175,6 +199,9 @@ const RichTextEditor = ({
           )}
           editor={editor}
           role="textbox"
+          // TODO DS-1112
+          // onBlur={handleEditorContentBlur}
+          // eslint-disable-next-line react/jsx-props-no-spreading
```
kyleshike, line 204:
```suggestion
```

src/RichTextEditor/RichTextEditorMenuBar.tsx
```diff
@@ -15,14 +15,20 @@ import {
   faListUl,
   faUnlink,
 } from '@fortawesome/pro-regular-svg-icons';
+import IconButton from '../IconButton';
```
kyleshike, line 18: I'm unsure why this change is necessary, but without using a relative path I was unable to compile rails-server with a locally linked DS with an absolute path

src/RichTextEditor/actionHandlers.ts
```diff
@@ -1,12 +1,15 @@
-export const createActionHandlers = (editor) => ({
+import type { Editor } from '@tiptap/core';
+
+export const createActionHandlers = (editor: Editor) => ({
   link: () => {
```
kyleshike, line 4: According to TS, this function expects a boolean return.  Since it was previously just returning undefined, I just swapped them it to return false.

src/RichTextEditor/actionHandlers.ts
```diff
@@ -1,12 +1,15 @@
...
     // TODO: use DS components for link prompt
     const previousUrl = editor.getAttributes('link').href;
+    // eslint-disable-next-line no-alert
```
kyleshike, line 7:
```suggestion
```

src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -43,11 +45,55 @@ const ExtendedLink = Link.extend({
...
+   If none provided then it uses sanitize-html's defaults
+   https://github.com/apostrophecms/sanitize-html#default-options
+  */
+  allowedTags?: string[];
+  /**
+    ARIA attributes to pass along to the text editor field
+    for accessibility purposes. The keys of the object should be valid
+    ARIA attributes.
+  */
+  ariaAttributes?: AriaAttributes;
```
kyleshike, line 66: Love that react provides this out of the box

src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -43,11 +45,55 @@ const ExtendedLink = Link.extend({
...
+   Which actions to include in the taskbar. Available as constants.
+   Current options are BOLD, ITALIC, LINK, UNLINK, UNORDERED_LIST and ORDERED_LIST
+  */
+  availableActions?: typeof RichTextEditorActions[keyof typeof RichTextEditorActions][];
+  characterLimit?: number;
+  className?: string;
+  /**
+   https://tiptap.dev/docs/editor/guide/custom-extensions
+  */
+  extensions?: (Extension | TipTapNode | Mark)[];
```
kyleshike, line 77: Open to suggestions here.  I could see it being confusing for devs to reconcile `allowedExtensions`  and `extensions`.  Perhaps `customExtensions` instead of `extensions`?

kyleshike:
I will be PTO for the next two weeks.  I have a draft PR (linked in the description above) that _should_ be a plug & play update to accommodate the changes here but erring on the side of caution, and even if/when this PR is approved, I think it will be best not to merge this PR into develop until I return so that it won't block any releases between now & when I come back.

gabescarbrough:
This is exciting! Looking forward to checking it out

src/RichTextEditor/RichTextEditor.stories.jsx
gabescarbrough: That makes sense to me

src/RichTextEditor/RichTextEditor.tsx
gabescarbrough: Agreed. This is awesome.

src/RichTextEditor/RichTextEditor.tsx
gabescarbrough: **suggestion (blocking)**: 💡 

Lets go with `customExtensions` since it describes the intent better.

If we find ourselves wanting to add extensions/marks/nodes that TipTap has existing extensions for (like underline, strike, color, etc) then we should add them via `RichTextEditorActions` and expose them through the `availableActions` prop. Most of the time you shouldn't have to register extensions yourself when using RichTextEditor.

This prop should be an escape hatch for when you're doing something truly custom or that is specific to app code beyond the text editor itself.

src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -43,11 +45,55 @@ const ExtendedLink = Link.extend({
...
+  /**
+   Which actions to include in the taskbar. Available as constants.
+   Current options are BOLD, ITALIC, LINK, UNLINK, UNORDERED_LIST and ORDERED_LIST
+  */
+  availableActions?: typeof RichTextEditorActions[keyof typeof RichTextEditorActions][];
+  characterLimit?: number;
+  className?: string;
+  /**
+   https://tiptap.dev/docs/editor/guide/custom-extensions
+  */
```
gabescarbrough, line 76: **suggestion (blocking)**: 💡

I think this deserves more clarification. Something like this but feel free to tweak the wording.

```suggestion
Custom extensions that provide functionality related to app features beyond the responsibility of the text editor itself. An example of such an extension would be template variables as the requirements of implementation would rely on other code processing them. Not yet supported extensions which relate solely to the text editor should be added via a pull request and used via `availableActions`.

  /**
   https://tiptap.dev/docs/editor/guide/custom-extensions
  */
```

src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -108,15 +145,15 @@ const RichTextEditor = ({
   ].filter((extension) => availableActions.includes(extension.name))
     .map((extension) => extension.config);
 
-  const extensions = [
+  const editorExtensions = [
```
gabescarbrough, line 148: **suggestion (non-blocking)**: 💡

For some reason GH isn't letting me do a multi-line suggestion here

I think this reads more clearly if we switch it to order of necessity (required > optional > custom)

Switching the name to `customExtensions` above also helps with readability

```ts
  const editorExtensions = [
    ...requiredExtensions,
    ...optionalExtensions,
    ...oneLineExtension,
    ...customExtensions,
  ];
```

gabescarbrough (commented):
This is awesome! Excited to see all the TS files.

A few minor tweaks and this will be good to go.

src/RichTextEditor/RichTextEditor.tsx
kyleshike: [dc4c9a8](https://github.com/user-interviews/ui-design-system/pull/1121/commits/dc4c9a800b63f4d2722abf7c256fc30c5861f37c)

src/RichTextEditor/RichTextEditor.tsx
kyleshike: [dc4c9a8](https://github.com/user-interviews/ui-design-system/pull/1121/commits/dc4c9a800b63f4d2722abf7c256fc30c5861f37c)

src/RichTextEditor/RichTextEditor.tsx
kyleshike: [dc4c9a8](https://github.com/user-interviews/ui-design-system/pull/1121/commits/dc4c9a800b63f4d2722abf7c256fc30c5861f37c)

gabescarbrough (approved):
Looks great!

Thanks for incorporating those changes

PR #1143, opened 2024-02-02
remove duplicate JS files RTE

jasonbasuil:


gabescarbrough (approved):
Thanks!

spec/__snapshots__/Storyshots.test.js.snap
```diff
@@ -3209,7 +3209,7 @@ Array [
   </button>,
   " ",
   <button
-    className="Button btn btn-brand-google btn-md"
+    className="Button btn btn-brand-google"
```
kyleshike, line 3212: This was interesting to me.  react-bootstrap doesn't actually provide a medium size: 

Screenshot 2024-02-05 at 9 21 47 AM


Instead it just assumes 'md' as default and offers two variants instead.  An interesting way to simplify the interface that I _think_ I like.  (ie: sane defaults while still offering customization).  I think though it might be _best_ to have all three sizes available and just have the props default to `md`  instead.

src/EmptyState/EmptyState.stories.tsx
```diff
@@ -26,7 +27,7 @@ export const Default: Story = {
 
 export const PrimaryAction: Story = {
   args: {
-    primaryAction: <Button leadingIcon={faPlus} variant="primary">New project</Button>,
+    primaryAction: <Button leadingIcon={faPlus as IconDefinition} variant="primary">New project</Button>,
```
kyleshike, line 30: I'd love to tackle this issue soon.  ASAIK, we have to cast this type because we have mismatched versions of all the fortawesome libraries we use. So long as there aren't any significant breaking changes in the updates, we just should move all these to the latest. Icons can't change too much, right?

kyleshike (approved):
Looks great, thanks for getting this done

PR #1150, opened 2024-02-06
Allow prop type node for Card title

domschab23:
Allowing prop `title` from `Card` component to be have `node` type

### DIFF
--- a/src/Card/Card.jsx
+++ b/src/Card/Card.jsx
@@ -91,7 +91,7 @@ Card.propTypes = {
   noPadding: PropTypes.bool,
   size: PropTypes.string,
   subTitle: PropTypes.string,
-  title: PropTypes.string,
+  title: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
 };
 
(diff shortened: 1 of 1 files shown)


jasonbasuil (approved):
👍

src/Card/Card.jsx
```diff
@@ -91,7 +91,7 @@ Card.propTypes = {
   noPadding: PropTypes.bool,
   size: PropTypes.string,
   subTitle: PropTypes.string,
-  title: PropTypes.string,
+  title: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
```
kyleshike, line 94: strings are technically nodes, so assuming prop-types is accurate, this can just be:

```suggestion
  title: PropTypes.node,
```

But probably worth double checking

domschab23: Done

kyleshike (approved):

domschab23:
@kyleshike / @jasonbasuil can you merge this PR (I don't have permissions)

PR #1151, opened 2024-02-07
IconButton: expose loading props

eric2523:
mentioned in our slack [here](https://userinterviews.slack.com/archives/C06J56K92EL/p1707244680141449?thread_ts=1707243526.321079&cid=C06J56K92EL) -- exposing the loading props for the `IconButton` component and updated the storybook with examples


eric2523:
was working off the wrong branch oops

PR #1152, opened 2024-02-07
IconButton: expose loading props

exian-ut:
mentioned in our slack [here](https://userinterviews.slack.com/archives/C06J56K92EL/p1707244680141449?thread_ts=1707243526.321079&cid=C06J56K92EL) -- exposing the loading props for the `IconButton` component and updated the storybook with examples

### DIFF
--- a/src/IconButton/IconButton.jsx
+++ b/src/IconButton/IconButton.jsx
@@ -57,7 +57,7 @@ export const IconButtonActions = {
 };
 
 const IconButton = ({
- action, ariaLabel, className, icon, isLoading, loadingText, size, variant, ...props
+ action, ariaLabel, className, icon, isLoading, size, variant, ...props
 }) => {
   const getAriaLabel = () => {
     if (action) {
@@ -71,7 +71,7 @@ const IconButton = ({
       aria-label={getAriaLabel()}
       className={classnames('IconButton', className)}
       isLoading={isLoading}
-      loadingText={loadingText}
+      loadingText=""
       size={size}
       variant={variant}
       {...props}
@@ -92,7 +92,6 @@ IconButton.propTypes = {
   className: PropTypes.string,
   icon: PropTypes.object.isRequired,
   isLoading: PropTypes.bool,
-  loadingText: PropTypes.string,
   size: PropTypes.oneOf(Object.values(ButtonSizes)),
   variant: PropTypes.string,
 };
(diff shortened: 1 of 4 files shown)


src/IconButton/IconButton.jsx
```diff
@@ -89,13 +91,17 @@ IconButton.propTypes = {
...
+  loadingText: PropTypes.string,
   size: PropTypes.oneOf(Object.values(ButtonSizes)),
   variant: PropTypes.string,
 };
 
 IconButton.defaultProps = {
   action: undefined,
   className: undefined,
+  isLoading: false,
+  loadingText: undefined,
```
jasonbasuil, line 104: ❓ wondering if `loadingText` should default to an empty string `""` since an `IconButton` doesn't have visible text to begin with, it might make sense to not have the default "Loading..." be present.

src/IconButton/IconButton.stories.jsx
```diff
@@ -37,6 +37,14 @@ export const AriaLabel = () => (
   </>
 );
 
+export const Loading = () => (
+  <>
+    <IconButton action="ADD" isLoading />
+    <IconButton action="ADD" isLoading loadingText="Adding participant..." />
+    <IconButton action="ADD" isLoading loadingText="" />
```
jasonbasuil, line 44: ❓ Curious if we should even allow `loadingText` to be given actual text. I could see a situation where a lot of content shifting could happen which is usually pretty jarring for users? I think it could be pretty self explanatory from the icon what action is happening when the loading spinner is present.

jasonbasuil (commented):
Thanks for bringing this in! Sorry for just getting to it, was out for a couple days. Had a couple questions and also adding @gabescarbrough for his opinion on this and I can also get some design input later today on this as well.

src/IconButton/IconButton.stories.jsx
gabescarbrough: Agreed. I don't think we should offer `loadingText` as an option for IconButton. As you said, I think the spinner is enough indication of what is going on.

src/IconButton/IconButton.jsx
gabescarbrough: I don't think we should even offer `loadingText` as a prop on IconButton. We should pass down `""` as the `loadingText` to Button to avoid the default text. Since IconButton is pretty much always used for some sort of action it is obvious that action caused a loading state indicated by the spinner.

gabescarbrough (approved):
🔥 thanks for exposing the `isLoading` prop. Will make it a lot more obvious this is an option.

If we remove `loadingText` then I think this good to go

src/IconButton/IconButton.stories.jsx
eric2523: makes sense to me. will update

jasonbasuil (approved):
Looks great! 🔥

PR #1154, opened 2024-02-12
Fix DropdownToggle focus

### LINKED ISSUE #1153: [UIDS] DropdownToggle focused when closed
## Description

If you click on the DropdownToggle to open the Dropdown then then it is `active`. If you click it again then it closes but has `focus` so it is still grey. If you click away then it loses focus.

Sav and I's expectation was that clicking it to close it would lose focus.

Interestingly, react-bootstrap's [Dropdown works how we expected
](https://react-bootstrap.netlify.app/docs/components/dropdowns) so its possible this is due to our styles

This is most obvious with the `transparent` variant but the same behavior happens for the others too. There is just less color different for the the `focus` styles.

## User Stories

- As a user, I want the DropdownToggle to close as expected so that I am not confused.

## Acceptance Criteria

- When the DropdownToggle is closed it is not focused

https://github.com/user-interviews/ui-design-system/assets/20173797/00e03556-187c-47da-b496-73fe2f6d1742

gabescarbrough:
Closes #1153

If you click on the `DropdownToggle` to open the `Dropdown` then the button is `active`, which our CSS styles to be a different color. If you click it again then it closes but has focus so it is still the `active` color. If you click away then it loses `focus` and goes back to the original color.

As [this goofy JSFiddle](https://jsfiddle.net/eo7fhaL6/) demonstrates that is the expected behavior for buttons when they are clicked twice since clicking it gives it focus. You generally don't notice this since something usually happens to takes focus away from the button when it is clicked (like a form submit) and button focus states aren't styled that often.

It feels really wonky for a dropdown toggle, however. So this PR changes the styles to what we expected, which is that if you click the button to open the dropdown and then click it to close it the button goes back to the original color.

### Before

https://github.com/user-interviews/ui-design-system/assets/20173797/56ece63e-0e45-4715-aeb0-be034e8a2cc4

https://github.com/user-interviews/ui-design-system/assets/20173797/746e874a-1ca9-4328-b333-6bb61e7330ea

### After

https://github.com/user-interviews/ui-design-system/assets/20173797/a9575071-7d91-42f9-9f68-6709f06a00b6

https://github.com/user-interviews/ui-design-system/assets/20173797/5ff1bcf3-c97b-4f74-acfe-adfdab2a4e74

### DIFF
--- a/src/Dropdown/DropdownToggle.scss
+++ b/src/Dropdown/DropdownToggle.scss
@@ -4,18 +4,6 @@ $primary: $ux-emerald-600;
 $danger: $ux-red;
 $warning: $ux-yellow-400; 
 
-.Synthesized {
-  .DropdownToggle {
-    &.btn-primary {
-      @include synthesis-btn-primary;
-    }
-  
-    &.btn-outline-primary {
-      @include synthesis-btn-outline-primary;
-    }
-  }
-}
-
 .DropdownToggle {
   @include synth-font-type-30--bold;
 
@@ -99,5 +87,17 @@ $warning: $ux-yellow-400;
 
   &.btn-outline-transparent {
     @include btn-outline-transparent;
+  }
+}
+
+.Synthesized {
+  .DropdownToggle {
+    &.btn-primary {
+      @include synthesis-btn-primary;
+    }
+
+    &.btn-outline-primary {
+      @include synthesis-btn-outline-primary;
+    }
   }
 }
--- a/src/Dropdown/DropdownToggle.scss
+++ b/src/Dropdown/DropdownToggle.scss
@@ -11,10 +11,6 @@ $warning: $ux-yellow-400;
     all: unset;
   }
 
-  &--unstyled {
-    all: unset;
-  }
-
   i, svg {
     &.icon-left {
       margin-right: $synth-spacing-2;
--- a/src/Dropdown/DropdownToggle.scss
+++ b/src/Dropdown/DropdownToggle.scss
@@ -39,6 +39,14 @@ $warning: $ux-yellow-400;
 
   &.btn-primary {
     @include btn-primary;
+
+    &:focus {
+      background: $ux-emerald-600;
+    }
+
+    &.show:focus {
+      background: $ux-emerald-700;
+    }
   }
 
   &.btn-outline-primary {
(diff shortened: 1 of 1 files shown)


src/Dropdown/DropdownToggle.scss
```diff
@@ -4,29 +4,13 @@ $primary: $ux-emerald-600;
...
-    &.btn-primary {
-      @include synthesis-btn-primary;
-    }
-  
-    &.btn-outline-primary {
-      @include synthesis-btn-outline-primary;
-    }
-  }
-}
-
```
gabescarbrough, line 18: Moved this down to the bottom so it takes precedence once the non-Synthesized styles

You'll notice in the screen recording that the focus state for the Synthesized primary button was wrong before this was moved

src/Dropdown/DropdownToggle.scss
```diff
@@ -4,29 +4,13 @@ $primary: $ux-emerald-600;
...
   @include synth-font-type-30--bold;
 
   &--unstyled {
     all: unset;
   }
 
-  &--unstyled {
-    all: unset;
-  }
-
```
gabescarbrough, line 29: This was duplicate code that came in via a merge

src/Dropdown/DropdownToggle.scss
```diff
@@ -39,6 +23,14 @@ $warning: $ux-yellow-400;
...
   &.btn-primary {
     @include btn-primary;
+
+    &:focus {
+      background: $ux-emerald-600;
+    }
+
+    &.show:focus {
+      background: $ux-emerald-700;
+    }
```
gabescarbrough, line 33: Treats focus styles different based on if the Dropdown is showing the options or not

The reason this is here, as opposed to changing it on Button is that this behavior only really makes sense in the context of the Button toggling something

src/Dropdown/DropdownToggle.scss
```diff
@@ -63,9 +71,29 @@ $warning: $ux-yellow-400;
...
+      background: $ux-gray-400;
+    }
   }
 
   &.btn-outline-transparent {
     @include btn-outline-transparent;
   }
 }
+
+.Synthesized {
```
gabescarbrough, line 89: The new styles weren't needed for the Synthesized DropdownToggle because the focused button isn't a different color

kyleshike (approved):
Nice catch, love that this is a CSS only solution 🚀

jasonbasuil (approved):
Sweet 🙌 thanks for fixing this!

PR #1158, opened 2024-02-23
remove warning and danger Button variants from docs

### LINKED ISSUE #1157: [UIDS] - remove danger and warning Button variants from docs
---
title: "[UIDS]"
assignees: ''

---

## Description

The `danger` and `warning` variants are no longer being used intentionally by designers, so getting ahead by not publishing them on Storybook just in case devs think they can still use them

jasonbasuil:
closes #1157 

We noticed during potluck yesterday that the `danger` variant of `Button` was still being introduced, though from the Design side it was an intentional decision that we won't be using `danger` or `warning` variants moving forward. This PR just unpublishes the stories from Storybook. 

We should probably get around to doing cleanup on RS, but at least this helps get ahead of any devs who might still be inclined to use them.


gabescarbrough:
Would it make sense to mark them as deprecated in the docs as opposed to removing the examples before the variants themselves are removed?

It feels more explicit to me to say "don't use this" than to not mention it

gabescarbrough (approved):

PR #1160, opened 2024-02-26
Converts Card to typescript

kyleshike:


src/Card/Card.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { LoadingSkeleton } from 'src/LoadingSkeleton';
+
+import './Card.scss';
+
+export const CardSizes = {
+  EXTRA_SMALL: 'xs',
+  SMALL: 'sm',
+  MEDIUM: 'md',
+  LARGE: 'lg',
+} as const;
```
kyleshike, line 18: I've kept this here for backward compatibility's sake, but it's not really necessary in Typescript

src/Card/Card.tsx
```diff
@@ -0,0 +1,98 @@
...
+  children?: ReactNode;
+  className?: string;
+  divided?: boolean;
+  elementType?: string;
+  helperText?: ReactNode;
+  isLoading?: boolean;
+  loadingSkeleton?: ReactNode;
+  loadingSkeletonParagraphCount?: number;
+  noPadding?: boolean;
+  size: 'xs' | 'sm' | 'md' | 'lg';
```
kyleshike, line 32: Instead, we can explicitly state that `size` can _only_ one of these four values

src/Card/Card.tsx
```diff
@@ -0,0 +1,98 @@
...
+import './Card.scss';
+
+export const CardSizes = {
+  EXTRA_SMALL: 'xs',
+  SMALL: 'sm',
+  MEDIUM: 'md',
+  LARGE: 'lg',
+} as const;
+
+type ElementProps = DetailedHTMLProps<HTMLAttributes<HTMLElement>, HTMLElement>;
```
AlexAlt, line 20: what is happening here with this syntax?

kyleshike: Yea so this is using what are called [generics](https://www.typescriptlang.org/docs/handbook/2/generics.html), which we won't be getting to for a few weeks in the course.

In essence, this type evaluates to the set of _all_  possible props that can be passed to an HTML element (as far as React is concerned). So really, it's the firehose of HTML props, including all even listeners, all aria attributes, etc, etc, _etc_.

The **only** reason that I'm using it here is because the card component is spreading all its props into the element that it renders.  This is an [anti-pattern](https://aurelio.me/blog/Why-I-stopped-spreading-props-react/) IMHO, but I didn't want to change the API of this component in this work, just convert it to TS:

https://github.com/user-interviews/ui-design-system/blob/c11d2f0b50148db507171832305864294c512ca7/src/Card/Card.tsx#L50

https://github.com/user-interviews/ui-design-system/blob/c11d2f0b50148db507171832305864294c512ca7/src/Card/Card.tsx#L65-L68

kyleshike:
@jasonbasuil @gabescarbrough mind giving this a look?

src/Card/Card.jsx
```diff
@@ -1,111 +0,0 @@
...
-      ),
-    },
-    cardChildren,
-  );
-};
-
-Card.propTypes = {
-  className: PropTypes.string,
-  divided: PropTypes.bool,
-  elementType: PropTypes.oneOfType([PropTypes.func, PropTypes.string]),
```
jasonbasuil, line 86: honestly weird that we accepted `PropTypes.func` here but seems to have been a long remnant of the deprecated Card when it was brought into the DS years ago.

src/Card/Card.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+type CardProps = {
+  children?: ReactNode;
+  className?: string;
+  divided?: boolean;
+  elementType?: string;
+  helperText?: ReactNode;
```
jasonbasuil, line 27: `helperText` was previously `string`, tho I do recall people wanting that bit of flexibility by accepting node here as well as the other text based props in order to work around some layout issues, but I think that's for another discussion about the `Card` so this works fine.

jasonbasuil (approved):
👍

PR #1165, opened 2024-03-11
Create DrawerFooter component with dropdown toggle for primary action

eric2523:
### Description
Us at rx-fast want to extend functionality of the `DrawerFooter` to also allow for a `DropdownToggle` as it's primary action. This adds a new `additionalPrimaryActions` prop that is the interface for rendering the individual `DropdownItems`. 

The `DrawerFooter` now conditionally renders a single primary action or dropdown toggle depending on `additionalPrimaryActions`

This also adds the ability to add a trailing icon to the `DropdownItem`.

### Screenshot
image

### DIFF
--- a/src/Drawer/Drawer.mdx
+++ b/src/Drawer/Drawer.mdx
@@ -67,6 +67,10 @@ you want the drawer to be the full width of the viewport but you want to disable
 
 If additional components are required to be included, pass them as children to the `DrawerFooter`. This example also doesn't have a title or a border on the header, and the background overlay is turned off to allow interaction with outside elements.
 
+### Additional Primary Actions
+
+If additional primary actions are required, you can utilize the `primaryAdditionalActions` prop. This will render the primary action as a dropdown toggle.
+
 ### Empty
 
 While using the corresponding child components is recommended, Drawer is fully functional on its own and be closed by clicking outside of it or pressing the escape key. This is intended as an escape hatch to allow fully custom contents. In fully custom cases it is encouraged to include a close button. As the child components all work independent of each other it is also possible to use the header and body, but not the footer, for instance. This is useful for drawers which only include information and not an action.
--- a/src/Drawer/DrawerFooter.jsx
+++ b/src/Drawer/DrawerFooter.jsx
@@ -64,7 +64,6 @@ const DrawerFooter = ({
               {additionalPrimaryActions.map((action) => (
                 <DropdownItem
                   disabled={action.disabled}
-                  href={action.href}
                   key={action.text}
                   leadingIcon={action.leadingIcon}
                   trailingIcon={action.trailingIcon}
--- a/src/Drawer/DrawerFooter.jsx
+++ b/src/Drawer/DrawerFooter.jsx
@@ -2,11 +2,15 @@ import React from 'react';
 import * as propTypes from 'prop-types';
 
 import Button from 'src/Button';
+import {
+ Dropdown, DropdownToggle, DropdownItem, DropdownMenu,
+} from 'src/Dropdown';
 
 import './DrawerFooter.scss';
 
 const DrawerFooter = ({
   children,
+  additionalPrimaryActions,
   isPrimaryActionLoading,
   isSecondaryActionLoading,
   primaryActionDisabled,
@@ -19,13 +23,16 @@ const DrawerFooter = ({
   secondaryActionText,
   onPrimaryAction,
   onSecondaryAction,
-}) => (
-  <div className="DrawerFooter">
-    <div>
-      {children}
-    </div>
-    <div className="DrawerFooter__actions">
-      {onSecondaryAction && (
+}) => {
+  const isDropdownToggle = !!additionalPrimaryActions.length;
+
+  return (
+    <div className="DrawerFooter">
+      <div>
+        {children}
+      </div>
+      <div className="DrawerFooter__actions">
+        {onSecondaryAction && (
         <Button
           disabled={secondaryActionDisabled}
           isLoading={isSecondaryActionLoading}
@@ -37,24 +44,53 @@ const DrawerFooter = ({
           {secondaryActionText}
         </Button>
       )}
-      {onPrimaryAction && (
-        <Button
-          disabled={primaryActionDisabled}
-          isLoading={isPrimaryActionLoading}
-          leadingIcon={primaryActionIcon}
-          loadingText={primaryActionLoadingText}
-          type="button"
-          variant={primaryActionVariant}
-          onClick={onPrimaryAction}
-        >
-          {primaryActionText}
-        </Button>
-      )}
+        {!isDropdownToggle && onPrimaryAction && (
+          <Button
+            disabled={primaryActionDisabled}
+            isLoading={isPrimaryActionLoading}
+            leadingIcon={primaryActionIcon}
+            loadingText={primaryActionLoadingText}
+            type="button"
+            variant={primaryActionVariant}
+            onClick={onPrimaryAction}
+          >
+            {primaryActionText}
+          </Button>
+        )}
+        {isDropdownToggle && (
+          <Dropdown>
+            <DropdownToggle>{primaryActionText}</DropdownToggle>
+            <DropdownMenu>
+              {additionalPrimaryActions.map((action) => (
+                <DropdownItem
+                  disabled={action.disabled}
+                  href={action.href}
+                  key={action.text}
+                  leadingIcon={action.leadingIcon}
+                  trailingIcon={action.trailingIcon}
+                  onClick={action.onClick}
+                >
+                  {action.text}
+                </DropdownItem>
+              ))}
+            </DropdownMenu>
+          </Dropdown>
+        )}
+      </div>
     </div>
-  </div>
-);
+  );
+};
 
 DrawerFooter.propTypes = {
+  additionalPrimaryActions: propTypes.arrayOf(
+    propTypes.shape({
+      disabled: propTypes.bool,
+      leadingIcon: propTypes.object,
+      text: propTypes.string.isRequired,
+      trailingIcon: propTypes.object,
+      onClick: propTypes.func.isRequired,
+    }),
+  ),
   children: propTypes.node,
   isPrimaryActionLoading: propTypes.bool,
   isSecondaryActionLoading: propTypes.bool,
@@ -72,6 +108,7 @@ DrawerFooter.propTypes = {
 
 DrawerFooter.defaultProps = {
   children: undefined,
+  additionalPrimaryActions: [],
   isPrimaryActionLoading: false,
   isSecondaryActionLoading: false,
   primaryActionDisabled: false,
--- a/spec/__snapshots__/Storyshots.test.js.snap
+++ b/spec/__snapshots__/Storyshots.test.js.snap
@@ -5927,6 +5927,101 @@ Array [
 ]
 `;
 
+exports[`Storyshots Components/Drawer Additional Primary Actions 1`] = `
+Array [
+  <button
+    className="Button btn btn-primary"
+    disabled={false}
+    onClick={[Function]}
+    type="button"
+  >
+    Open
+  </button>,
+  <div
+    className="DrawerBackgroundOverlay"
+    onClick={[Function]}
+    onKeyDown={[Function]}
+    role="presentation"
+  />,
+  <div
+    className="Drawer Drawer--right Drawer--sm Drawer--behind-nav"
+  >
+    <div
+      className="Drawer__header Drawer__header--bordered"
+    >
+      <div
+        className="Drawer__title"
+      >
+        Title goes here
+      </div>
+      <button
+        aria-label="Close"
+        className="Drawer__header-action"
+        onClick={[Function]}
+        type="button"
+      >
+        <svg
+          aria-hidden="true"
+          className="svg-inline--fa fa-xmark "
+          data-icon="xmark"
+          data-prefix="fas"
+          focusable="false"
+          role="img"
+          style={Object {}}
+          viewBox="0 0 384 512"
+          xmlns="http://www.w3.org/2000/svg"
+        >
+          <path
+            d="M342.6 150.6c12.5-12.5 12.5-32.8 0-45.3s-32.8-12.5-45.3 0L192 210.7 86.6 105.4c-12.5-12.5-32.8-12.5-45.3 0s-12.5 32.8 0 45.3L146.7 256 41.4 361.4c-12.5 12.5-12.5 32.8 0 45.3s32.8 12.5 45.3 0L192 301.3 297.4 406.6c12.5 12.5 32.8 12.5 45.3 0s12.5-32.8 0-45.3L237.3 256 342.6 150.6z"
+            fill="currentColor"
+            style={Object {}}
+          />
+        </svg>
+      </button>
+    </div>
+    <div
+      className="Drawer__body"
+    >
+      <p>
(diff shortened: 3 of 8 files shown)


src/Drawer/DrawerDropdownToggleFooter.jsx
```diff
@@ -0,0 +1,71 @@
...
+            loadingText={secondaryActionLoadingText}
+            type="button"
+            variant="transparent"
+            onClick={onSecondaryAction}
+          >
+            {secondaryActionText}
+          </Button>
+        )}
+        <Dropdown>
+          {dropdownToggle}
```
eric2523, line 43: with another thought it might be better to call the `DropdownToggle` here and just have a `primaryActionText` prop. Forces this component to stay true to being a dropdown toggle

kyleshike (commented):
What do you think about just extending `DrawerDropdownFooter` to be able to accept either a primary action or a primary dropdown rather than a new component with its own interface? I think that it might help reduce complexity in rails-server (eg: number of components to import for a drawer, etc)

If could be something like:
```jsx
DrawerFooter.propTypes = {
  // ...etc
  primaryActionDropdownItems: propTypes.arrayOf(proptypes.shape({
    text: propTypes.node.isRequired,
    onClick: propTypes.func.isRequired,
  })),
};
```

Then in the component, just do something like:
```jsx
    {onPrimaryAction && (
        <Button
          disabled={primaryActionDisabled}
          isLoading={isPrimaryActionLoading}
          leadingIcon={primaryActionIcon}
          loadingText={primaryActionLoadingText}
          type="button"
          variant={primaryActionVariant}
          onClick={onPrimaryAction}
        >
          {primaryActionText}
        </Button>
      )}
      {primaryActionDropdownItems && (
        <Dropdown>
          <DropdownToggle variant={primaryActionVariant}>
            {primaryActionText}
          </DropdownToggle>
          <DropdownMenu>
            {primaryActionDropdownItems.map((item) =>
              <DropdownItem href="#" onClick={item.onClick}>{item.text}</DropdownItem>)}
          </DropdownMenu>
        </Dropdown>
      )}
```

Curious what y'all think

eric2523:
Yea I was a bit torn on this. My OG thought process was if we add an additional prop to `DropdownFooter` we start getting into the world where we are telling the component what to do vs getting the thing we need off the bat. _Buuuttt_ that was based off my idea that either the `Dropdown` or `PrimaryAction` would render and _not_ both. With that I think the first question to answer is if it makes sense to allow the footer to have both types of buttons displayed together. That will decide how opinionated we want this thing to be.

Imo I feel like we'd want to enforce a secondary, primary and _not_ tertiary button? At least for now it seems like a lot of our drawers are forms when they have these actions and so idk if i see a picture/world of having both. 

@kyleshike

src/Drawer/DrawerDropdownToggleFooter.jsx
```diff
@@ -0,0 +1,71 @@
...
+import { Dropdown, DropdownMenu } from 'src/Dropdown';
+import PropTypes from 'prop-types';
+
+import './DrawerDropdownToggleFooter.scss';
+
+export function DrawerDropdownToggleFooter({
+  children,
+  isSecondaryActionLoading,
+  renderDropdownToggle,
+  renderDropdownItems,
```
gabescarbrough, line 12: Do these need to be render functions? Couldn't we just use DropdownToggle and DropdownItem within this component?

src/Drawer/DrawerDropdownToggleFooter.jsx
gabescarbrough: Yeah I think avoiding render functions is a good idea

gabescarbrough:
> What do you think about just extending DrawerDropdownFooter to be able to accept either a primary action or a primary dropdown rather than a new component with its own interface? I think that it might help reduce complexity in rails-server (eg: number of components to import for a drawer, etc)

If allowing multiple primary actions was a requirement at the time I was deciding the props for the Drawer components I think I would have approached it similar to what you outlined instead of taking one primary action. If the array only had one action then you could display it as a button, and if there was more than one you could display it as a select.

```js
DrawerFooter.propTypes = {
  // ...etc
  primaryActions: propTypes.arrayOf(proptypes.shape({
    text: propTypes.node.isRequired,
    onClick: propTypes.func.isRequired,
  })),
};
```

I tried to keep the Drawer's child components decoupled enough that you could write a custom version of any of them easily. So having a different DrawerFooter doesn't bother me, but I do get the argument that this is very similar to the existing footer so it might make sense to just have one.

gabescarbrough:
It does kind of bug me that this isn't a split button. My expectation would be that clicking "Save" would save and clicking the down-arrow icon would show additional options. Having it as a dropdown feels off to me, but we don't have a split button right now so that would be something else we'd have to add first.

gabescarbrough:
> Imo I feel like we'd want to enforce a secondary, primary and not tertiary button? At least for now it seems like a lot of our drawers are forms when they have these actions and so idk if i see a picture/world of having both.

For what its worth, the original use-case for the Drawer (bulk emails) has the preview button over on the left as a tertiary button which was done by passing another Button as `children` to the DrawerFooter. The original idea was that anything you needed to add over to the left could be passed as `children` and be more use-case specific whereas the primary and secondary buttons would generally always be the same. If you needed something fully custom you could build your own DrawerFooter and use it with the Drawer. I still prefer have tertiary items over to the left and am not a big fan of dropdowns used like this but I understand that folks think it looks cleaner.

eric2523:
> If the array only had one action then you could display it as a button, and if there was more than one you could display it as a select.

Hmm imo this feels like a [primitive obsession](https://medium.com/the-sixt-india-blog/primitive-obsession-code-smell-that-hurt-people-the-most-5cbdd70496e9) smell where we are trying to using an array here to define a type and use it to control a lot of logic. Which bc of that I do lean a towards kyle's approach with TS and using type predicates to distinguish between the two diff action buttons. 

> For what its worth, the original use-case for the Drawer (bulk emails) has the preview button over on the left as a tertiary button which was done by passing another Button as children to the DrawerFooter.

Yea on the new workspace side of things we are trying to align the way we handle saves and previews. I'm not sure if this was discussed amongst the designers whether we wanna move behavior towards the dropdown tho

gabescarbrough:
> Hmm imo this feels like a [primitive obsession](https://medium.com/the-sixt-india-blog/primitive-obsession-code-smell-that-hurt-people-the-most-5cbdd70496e9) smell where we are trying to using an array's length to define a type and use it to control a lot of logic.

It doesn't feel like a lot of logic to me. Either it is one action and you display it as one thing (a button) or it is multiple actions and you display it as multiple grouped things (a select). It's not so much defining a type, as displaying correctly for the amount of actions given. It doesn't matter much now though because switching to that prop interface would be a breaking change so its a no-go anyway.

I'm not opposed to enforcing one or the other with TS. Then it is clear that you're meant to use a button _or_ a select for the primary action and not both. Having the necessary props for either feels kind of ugly to me. That's what makes me wish you could just provide an array of actions and have the component worry about how to display it. But having props for either gets you back to having one DrawerFooter component so I'm not opposed if that is a priority for folks.

> Yea on the new workspace side of things we are trying to align the way we handle saves and previews. I'm not sure if this was discussed amongst the designers whether we wanna move behavior towards the dropdown tho

Yeah it was. I made the case that I prefer the tertiary button over to the side but I'm the outlier 👀

natkasman:
Scoooching in here to add some context and rationale behind this! 

I 1000% agree that a split button is more appropriate for this than a dropdown. 

So currently we have:
(1) Preview participant view for scheduling
(2) Preview screener
(3) Preview emails in workspace
(4) Preview email in Hub table

While all different, they are similar in requiring the user to "Save" before previewing. Currently, we do what was suggested, and display the Preview button on the left. We got feedback that it was unclear what the preview experience was going to be—were users going to see what they typed pre-save, or did they need to save first in order for the preview to work? 

image

We changed the button language to "Save and send preview" but it's quite lengthy for button text, and makes it more obvious that there are 2 "Save"-related actions in the action bar. Given that our app is already quite text-heavy, we thought this approach would be more win-win because it makes it obvious for users that you need to Save first before you can preview, and [minimizes the options that users can take](https://lawsofux.com/hicks-law/).

Wrt the question of "are we doing this for all previews" - I think it makes sense for the experience to be consistent (I think the above 4 are the only preview experiences we have, but lmk if i'm missing anything). If I'm missing a preview experience, and it doesn't make sense to do this in that scenario, happy to explore!

jasonbasuil:
> It does kind of bug me that this isn't a split button. My expectation would be that clicking "Save" would save and clicking the down-arrow icon would show additional options. Having it as a dropdown feels off to me, but we don't have a split button right now so that would be something else we'd have to add first.

➕ fwiw haven't looked too close but looks like DropdownToggle supports a split button https://react-bootstrap.netlify.app/docs/components/dropdowns#split-button-dropdowns

eric2523:
> Having the necessary props for either feels kind of ugly to me. That's what makes me wish you could just provide an array of actions and have the component worry about how to display it.

Mm i see wym. Like you gotta know that potentially prop A can't be used with prop B or something like that. Im curious if we pass an array of actions how will that alleviate the "either this prop or that prop" problem? Expose only the props/behaviors that applies to both a Button or DropdownToggle? If so, what if theres a control that only DropdownToggle uses but is important enough that we wanna expose it.

gabescarbrough:
After thinking about this some more I think the way to go is

```js
DrawerFooter.propTypes = {
  // ...etc
  isPrimaryActionLoading: propTypes.bool,
  primaryActionDisabled: propTypes.bool,
  primaryActionIcon: propTypes.object,
  primaryActionLoadingText: propTypes.string,
  primaryActionText: propTypes.string,
  primaryActionVariant: propTypes.string,
  onPrimaryAction: propTypes.func,
  additionalPrimaryActions: propTypes.arrayOf(proptypes.shape({
    text: propTypes.node.isRequired,
    onClick: propTypes.func.isRequired,
  })),
};
```
If you provide `additionalPrimaryActions` then the primary action is displayed as a split button with the additional actions in the dropdown. If you don't then it is displayed as a button like now. The `DrawerFooter` owns how the button is displayed as opposed to taking a render function.

This is backwards compatible. The existing props relate to the primary button and can still work the same. `isPrimaryActionLoading`, `primaryActionDisabled`, etc are relevant whether there are additional actions or not. It is explicit which action is intended for the button vs the select sections of the split button.

This pattern doesn't require props that are exclusive of each other. Converting to TS would be great but I think having props that don't work with each other is confusing.

We don't need to have multiple DrawerFooter components and I think its pretty intuitive what `additionalPrimaryActions` is for, especially if you have seen a split button example.

> Expose only the props/behaviors that applies to both a Button or DropdownToggle? If so, what if theres a control that only DropdownToggle uses but is important enough that we wanna expose it.

Expose only the props we want devs to use. Since the goal here is consistency it's actually better to not make this too flexible. The existing props should play nice with either a button or a split button. Some of the existing props we should probably actually remove in a future major version, like `primaryActionVariant` was intended to allow you to use a `danger` button for a destructive action or similar and we don't recommend that anymore.

eric2523:
One thing i'm thinking about is the `onPrimaryAction` prop for a toggle and whether we should make that exclusive but I can see the benefit of having some tracking passed to that primary action which is okay. Like if you choose to do some weird behavior thats on you lol. 

I think I like it. This checks off: 
- [x] restricting the type of primary button that renders. 
- [x] Having one component
- [x] Not falling into primitive obsession as the `additionalPrimaryActions` is like the type that drives render logic instead of the array length 
- [x] Not having exclusive props exposed that only pertains to one or the other

This is great i like it and i'm down to move on with this solution!

src/Drawer/DrawerFooter.jsx
```diff
@@ -37,24 +44,52 @@ const DrawerFooter = ({
...
+            onClick={onPrimaryAction}
+          >
+            {primaryActionText}
+          </Button>
+        )}
+        {isDropdownToggle && (
+          <Dropdown>
+            <DropdownToggle>{primaryActionText}</DropdownToggle>
+            <DropdownMenu>
+              {additionalPrimaryActions.map((action) => (
```
kyleshike, line 64: I might be misreading this, but what happens when both `onPrimaryAction` is present and `isDropdownToggle` is true?  It looks like it won't render the primary action in that case.  Is that  intentional?

src/Drawer/Drawer.mdx
```diff
@@ -67,6 +67,10 @@ you want the drawer to be the full width of the viewport but you want to disable
 
 If additional components are required to be included, pass them as children to the `DrawerFooter`. This example also doesn't have a title or a border on the header, and the background overlay is turned off to allow interaction with outside elements.
 
+### Additional Primary Actions
+
+If additional primary actions are required, you can utilize the `primaryAdditionalActions` prop. This will render the primary action as a dropdown toggle.
```
gabescarbrough, line 72:
```suggestion
If additional primary actions are required, you can utilize the `additionalPrimaryActions` prop. This will render the primary action as a dropdown toggle.
```

src/Drawer/DrawerFooter.jsx
gabescarbrough: @eric2523 my intention was that when `additionalPrimaryActions` is passed then the primary action should be a split button, not a dropdown. I think having a typical dropdown for a primary action is confusing. I don't think I've ever seen that in an app before. Non-split dropdowns are better suited to menus of actions as opposed to primary actions.

@kyleshike if the primary action is a split button then `onPrimaryAction` is what is called when you click the primary action, and the `additionalPrimaryActions` `onClick` functions would be called for the additional actions.

Using this random split button example I found on NN/g, clicking "Insert" would call `onPrimaryAction`. Clicking the downward arrow then clicking "Insert above" would call the `onClick` associating with that action.

image

As Jason pointed out React Bootstrap's `Dropdown` component supports split buttons so ours (which wraps it) should be able to as well: https://react-bootstrap.netlify.app/docs/components/dropdowns/#split-button-dropdowns

My preference would be to add a `SplitButton` component so we could use it elsewhere easily as well. I could see an argument for supporting split buttons as part of `Button` or `Dropdown` but it's probably more discoverable if it is its own thing. Also the props could be neater.

natkasman: I think we preferably want to get this in soon bc we need this for the "Preview participant's view" functionality for Automated scheduling. But I agree that a split button makes the most sense here and will be more flexible for future usage. If it doesn't take too long to add and use that instead, I'm on board.

gabescarbrough: I don't think it would be a big lift to switch to using a split button but I also get that balancing responding to DS PR feedback to end up with an ideal general implementation and shipping fast on the product side can be at odds at times

In my opinion, we shouldn't merge this current implementation to the DS if what we want long term is a split button. We should get to the implementation we want then include that in an upcoming DS release. Both because semver means that we're committing to not breaking the interface to the component until the next major version, but also because when we announce it devs are going to get used to and expect what we announce.

If shipping fast is a priority on the product side to the point where the ticket can't wait on this PR, it doesn't have to though. This implementation of `DrawerFooter` could be included in `rails-server` in the `design_system_components` folder and used until the DS version has been finalized and rolled out. At that point you'd just swap in the DS version.

gabescarbrough: That said, this is most of the way there so either way I don't think this PR is far off of the ideal general implementation. The sticking point is really the split button.

eric2523: Hmm i think im gonna bring this into rails-server to avoid this from spiraling too much as well as making it were able to iterate it on it before the DS. One of my concerns is that passing the `split` prop to the DropdownToggle will require some style changes and potentially changes to how were encapsulating that component atm since you'll still need to put a `Button` component next to the split toggle. 

```jsx
<Button variant="success">Split Button</Button>
      <Dropdown.Toggle split variant="success" id="dropdown-split-basic" />
```

jeffbax:
take it this one can be closed?

eric2523:
super old so closing

PR #1168, opened 2024-03-25
Provides CSS properties of commonly used variables

kyleshike:
Providing CSS properties to rails-server will allow us to start moving away from Sass and toward CSS, which in turn will unburden our Webpack compiler.

This PR adds css properties for:
```
- borders
- box_shadow
- colors
- elevations
- spacing
- typography
- z_stack
```


bin/link-rs
```diff
@@ -10,4 +10,4 @@ fi
 yarn build
 yalc publish
 cd $RAILS_SERVER_PATH
-yalc link @user-interviews/ui-design-system
+yalc add @user-interviews/ui-design-system
```
kyleshike, line 13: link was giving me grief.  Add is a bit more explicit, and doesn't introduce flakey symlinks when linking/unlinking

> When you run yalc add my-package in your project it pulls package content into .yalc in the current folder and injects a file:/link: dependency into package.json. Alternatively, you may use yalc link my-package which will create a symlink to the package content in node_modules and will not touch package.json (like npm/yarn link does), or you even may use it with Pnmp/Yarn/Npm workspaces.

https://github.com/wclr/yalc

scss/css_properties.scss
kyleshike, line 1: I opted to not be overly opinionated here about what I brought into this file, instead opting to just bring in the things that I know to be common needs in Rails server (typography, spacing, color).

kyleshike:
https://github.com/user-interviews/rails-server/pull/24480/commits/f698df6f0e12fbe987f7db41dd918cdc7c6abfdc highlights what this transition will look like

scss/css_properties.scss
jasonbasuil: This also just reminds me how much I want to get our variables renamed and better organized, especially if we're creating CSS properties for all of them and bringing them here.

jasonbasuil (approved):
Thanks for getting this started! 

Linked locally and seems good. I think this is a good place to start just to begin the process of moving away from SCSS. I think we'll probably want to create a milestone or something to track more of what this cutover might look like + getting a start on renaming / reorganizing our current variable names to be more in line with the `$synth-` naming we started.

scss/css_properties.scss
kyleshike: Yea 100%.  I'm trying not to try and do too much at the same time here even though I know we have a lot we want to do there

PR #1172, opened 2024-03-26
Cleans up a subset of stories

kyleshike:
In preparation for Storybook 8, which does not support knobs, this PR begins the work to remove knobs from our stories and replace them with controls. I've also made what I feel are improvements to the noise of the previous stories by relying on these controls more heavily.  For example, rather than enumerating all an alert's types, we can simply wire up the primary story with the proper variants, and it becomes rather self documenting.


src/Container/Container.stories.jsx
kyleshike, line 1: I don't think these stories worked previously.  But this component is only used in four places in rails-server and is just a very thin wrapper of Bootstrap's `container`, so I didn't give it much priority, just got rid of the knobs

src/CardStack/CardStack.stories.jsx
```diff
@@ -1,48 +1,43 @@
 import React from 'react';
 
 import Card, { CardSizes } from 'src/Card';
-import { withKnobs, select } from '@storybook/addon-knobs';
+// import { withKnobs, select } from '@storybook/addon-knobs';
```
kyleshike, line 4:
```suggestion
```

.storybook/preview.js
```diff
@@ -1,6 +1,14 @@
...
 
 const preview = {
+  decorators: [
+    (Story) => (
+      <div className="Synthesized">
+        {/* 👇 Decorators in Storybook also accept a function. Replace <Story/> with Story() to enable it  */}
+        <Story />
+      </div>
+    ),
+  ],
```
gabescarbrough, line 11: **thought**: 💭

I wonder if it is worth having the wrapper be a component with a toggle for whether `Synthesized` is set or not

Given parts of our app use the old styles for the same components still it would be nice for the DS docs to allow you to see how the components will appear in either context

Not blocking to this PR but def worth discussing

kyleshike: Yea definitely a good idea.  I really wish we didnt' have to continue supporting non Synth styles, but also understand that the effort to update those views is low priority from a business perspective.

package.json
```diff
@@ -192,4 +191,4 @@
   "readme": "https://github.com/user-interviews/ui-design-system#readme",
   "homepage": "https://github.com/user-interviews/ui-design-system",
   "_id": "@user-interviews/ui-design-system@1.32.0"
-}
+}
```
kyleshike, line 194:
```suggestion
}

```

kyleshike:
@gabescarbrough @jasonbasuil Alright this is a _lot_ cleaner now & ready for review

gabescarbrough (approved):
Looks good! 🚀

PR #1173, opened 2024-03-27
Update RichTextEditor to be a forwardRef and expose setContent through useImperativeHandle

eric2523:
Us at rx-fast want a way to programmatically set the content within the RichTextEditor. Specifically for the team email templates page, we have an action to reset the editor content to a default. This exposes a `setContent` function that uses `setContent` from the editor object returned by the `useEditor` hook. Nat and I thought it was best to not have this be an extension as the behavior for email templates are pretty specific and so a generic reset icon within the actions toolbar might be ambiguous. 

https://github.com/user-interviews/ui-design-system/assets/65802187/18a36d31-1d2c-4cf2-9bed-10d8a194ab38

### DIFF
--- a/src/RichTextEditor/RichTextEditor.tsx
+++ b/src/RichTextEditor/RichTextEditor.tsx
@@ -4,7 +4,7 @@ import type { Extension, Node as TipTapNode, Mark } from '@tiptap/core';
 
 import './RichTextEditor.scss';
 
-import React from 'react';
+import React, { forwardRef, type ForwardedRef, useImperativeHandle } from 'react';
 
 import classNames from 'classnames';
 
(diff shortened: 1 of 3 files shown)


src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -94,21 +94,28 @@ export type RichTextEditorProps = {
...
-  id,
-  initialValue,
-  isOneLine,
-  onChange,
-  placeholder,
-  customExtensions = [],
-}: RichTextEditorProps) {
+export type RichTextEditorRef = {
+  setContent: (content: string) => void;
+}
```
kyleshike, line 99: Great call only exposing this particular function of the editor

kyleshike (approved):

PR #1176, opened 2024-03-27
Fixes Card import resolution

kyleshike:


src/Card/Card.tsx
```diff
@@ -6,7 +6,7 @@ import React, {
 } from 'react';
 import classNames from 'classnames';
 
-import { LoadingSkeleton } from 'src/LoadingSkeleton';
+import { LoadingSkeleton } from '../LoadingSkeleton';
```
kyleshike, line 9: I'm not entirely sure why, but the module-aliasing between babel, eslint, and typescript differ to such an extent that the previous import statement is broken when consuming this in rails-server.  Updating it to a relative import should fix it.

jasonbasuil (approved):

PR #1179, opened 2024-03-29
Upgrade Storybook to v8.0.5

kyleshike:
This completes the upgrade from Storybook 7 to 8.  A few things were done:
- removes storyshots & snapshot tests as they aren't supported any longer
- slight reconfiguration of webpack & babel tooling
- updates node version of the DS to 18
- minor miscellaneous changes

most of this diff is the snapshots going away and yarn.lock.  the changes are actually not that crazy


spec/Storyshots.test.js
kyleshike, line 1: Storyshots are no longer supported

stories/Intro.stories.jsx
kyleshike, line 1: Defining stories _within_ `*.mdx` files is no longer supported, so I moved this into jsx.  Not exactly ideal, but I figure we can address this when we address documentation more broadly

package.json
```diff
@@ -34,16 +34,16 @@
   },
   "scripts": {
     "build": "NODE_ENV=production babel src --out-dir lib --copy-files --extensions '.js,.jsx,.ts,.tsx' && tsc",
-    "build-storybook": "storybook build -s public",
-    "build-storybook-docs": "storybook build -s public --docs",
+    "build-storybook": "storybook build public",
```
gabescarbrough, line 37: **question (non-blocking)**: ❓

What's the difference between `storybook build -s` and `storybook build`? Is `-s` not supported in v8?

.node-version
```diff
@@ -1 +1 @@
-16.20.0
+18.18.2
```
gabescarbrough, line 1: **praise**: 🙌

.storybook/preview.js
```diff
@@ -1,4 +1,6 @@
 import colors from '../src/Styles/colors/palette';
+import '../scss/global.scss';
+import 'bootstrap/dist/css/bootstrap.min.css';
```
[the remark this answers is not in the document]
gabescarbrough: This order means the Bootstrap styles will override ours right?

That is true to `rails-server`, so this does keep us honest by having Storybook reflect how the components will look in-app. It also makes my eye twitch since ideally we should be able to override Boostrap's styles easily.

gabescarbrough (approved):
Looks good to me. Thanks for tackling this.

package.json
kyleshike: Yea that's right.  -s isn't supported in v8:

> With the release of Storybook 8, the -s CLI flag was removed. We recommend using the [static directory](https://storybook.js.org/docs/configure/images-and-assets#serving-static-files-via-storybook) instead if you need to serve static files.

I'm actually going to dig on this a bit and make sure I didn't break something here, because I don't believe I've tested what serving this output actually renders

.storybook/preview.js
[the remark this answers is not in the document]
kyleshike: I would love to prioritize some general stylesheet cleanup this year and getting bootstrap out of rails server altogether & use the DS as the only entry point into Booststrap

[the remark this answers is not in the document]
gabescarbrough: 💯

jasonbasuil (approved):
LGTM! Thanks for getting on this 🚀

package.json
kyleshike: Hosted the static build locally and everything looks to be in order 👍

PR #1180, opened 2024-04-04
Updates node to 20, React to 18, FA peer dep to latest

kyleshike:
This PR does the following:
- Upgrades Node to 20
- Upgrade4s React to 18
- Fixers all the dependency warnings where feasible
- minor cleanup


.eslintrc
```diff
@@ -26,6 +26,7 @@
       "parser": "@typescript-eslint/parser",
       "plugins": [ "babel", "react", "react-hooks" ],
       "rules": {
+        "@typescript-eslint/explicit-module-boundary-types": "off",
```
kyleshike, line 29: Fixing the dep warning by updating tseslint makes us need to manually turn this rule off

src/InputLabel/InputLabel.jsx
```diff
@@ -8,7 +8,6 @@ import 'scss/forms/input_label.scss';
 
 const InputLabel = ({
   className,
-  elementType,
```
kyleshike, line 11: this prop isn't used in the component, so I just removed it

src/Layout/SidebarView/SidebarView.jsx
```diff
@@ -145,7 +145,6 @@ export default class SidebarView extends Component {
     const sidebarClassName = classNames(
       this.props.className,
       {
-        // eslint-disable-next-line @typescript-eslint/camelcase
```
kyleshike, line 148: updating TS lint made this obsolete

kyleshike:
This is going to be merged into the **DS 3.0** branch, not develop.  We can use this feature branch leading up to tech debt week.  We can then use part of TD week to merge it, make the release, and update rails-server

kyleshike:
@jasonbasuil @gabescarbrough bumping this. Nothing crazy going on here just some major version updates, but I want to get the ball rolling on DS 3.0

.node-version
```diff
@@ -1 +1 @@
-18.18.2
+20.12.1
```
gabescarbrough, line 1: 🥳

src/InputLabel/InputLabel.jsx
gabescarbrough: This prop isn't being used anywhere in `rails-server` either so that's good

package.json
```diff
@@ -70,12 +68,12 @@
     "bootstrap": "5.1",
     "classnames": "^2.2.5",
     "prop-types": "^15.6.1",
-    "react": "^16.12.0",
+    "react": "^18.2.0",
```
gabescarbrough, line 71: 🎉

gabescarbrough:
> Fixers all the dependency warnings where feasible

Is the only non-feasible one `react-toggle@4.1.1" has incorrect peer dependency "react@^15.3.0 || ^16.0.0"` or are there others too?

Given react-toggle is unmaintained and has [an issue from two years ago asking about React 18 with no replies](https://github.com/aaronshaf/react-toggle/issues/177) I'd say we should look into moving away from that dependency

kyleshike:
Yea I believe the main offender was react toggle.  IMO we can probably just replace it with some react code to do the same:
https://docs.userinterviews.com/?path=/docs/components-toggleinput--docs

PR #1184, opened 2024-04-11
Improves hover state of tabs

kyleshike:
https://github.com/user-interviews/ui-design-system/assets/3161151/8b2b9627-918b-4d08-a496-932c4808156e

Previously, hovered tab items would have their text bolded. In some situations, this can be problematic as bolded text is larger than non-bolded text (all else being equal). This difference causes some rather unfortunate content shifting.

Now, we use a lighter color on inactive tab items, and indicate the hover state by using text black on hover.

### DIFF
--- a/scss/colors/synthesis-palette.scss
+++ b/scss/colors/synthesis-palette.scss
@@ -19,6 +19,7 @@ $synth-header-neutral: #fbfbfb;
 $synth-page-neutral: #fdfdfd;
 $synth-text-neutral: #1b1b1b;
 $synth-unselected-neutral: #5b5b5b;
+$synth-unselected-tab: #c4c4c4;
 
 // main schema - visual hierarchy
 $synth-accent-green: #158d71;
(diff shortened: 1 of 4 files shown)


scss/colors/synthesis-palette.scss
```diff
@@ -19,6 +19,7 @@ $synth-header-neutral: #fbfbfb;
 $synth-page-neutral: #fdfdfd;
 $synth-text-neutral: #1b1b1b;
 $synth-unselected-neutral: #5b5b5b;
+$synth-unselected-tab: #c4c4c4;
```
gabescarbrough, line 22: This color fails WCAG color contrast guidelines

image

Looks like `#707070` is about as light as you can get and pass against white

Here's the link to check color combos https://webaim.org/resources/contrastchecker/

kyleshike: [41edc0c](https://github.com/user-interviews/ui-design-system/pull/1184/commits/41edc0c95339fb4d9876049e76710496f46d7a7d)

gabescarbrough (approved):
🚢

PR #1190, opened 2024-04-19
Upgrades to Yarn 4.1.1

kyleshike:


.github/workflows/chromatic.yml
kyleshike, line 1: each of these workflows were _slightly_ different from one another.  I had to make some changes to get yarn 4 working correctly so I just took some extra time to shore things up here as well

jeffbax (approved):

PR #1195, opened 2024-04-26
Add default containerClassName to loading skeleton

### LINKED ISSUE #1194: [UIDS] Update LoadingSkeleton to have the exact height that we specify
## Description

We use `react-loading-skeleton` underneath our `LoadingSkeleton` component. There's a known issue with react-loading-skeleton having container height slightly off https://www.npmjs.com/package/react-loading-skeleton#the-height-of-my-container-is-off-by-a-few-pixels. 

## User Stories

## Acceptance Criteria

- [Criteria 1]
- [Criteria 2]
- [Criteria 3]

domschab23:
[//]: # (
See https://www.notion.so/Expectations-for-Code-Reviews-4c261233caf64f12ab459ba090be2c50 for more best practices around how we do Code Review.
This is a default pull request template that teams can use as a starting point.
You can edit this here: docs/pull_request_template.md
)
### Related Issue(s)

Fix #1194

### Description
[//]: # (
Provide a brief description of the intent of this pull request.
Example: Allow users to search for a specific product by name.
)

Addressing https://www.npmjs.com/package/react-loading-skeleton#the-height-of-my-container-is-off-by-a-few-pixels

### Reviewers
[//]: # (
Indicate who your primary and secondary reviewers are, and make sure they know.
Do this AFTER you have done your own self-review.
)

Primary:  @kyleshike 

secondary: @jasonbasuil / @gabescarbrough  

cc: @user-interviews/rx-fast 

### Summary of Changes
[//]: # (
Provide a list of the major areas of change or screenshots and provide context that makes it easier to understand what is going on when first reading the code.
Think big picture, don't just replicate the *Files changed* tab or feel the need to have an item for each file changed if there are natural groupings, like adding a controller action and a route, or a view and a component.
The purpose is to call out anything you found confusing while developing, or that might not be obvious. What would you want a heads up about if you were the reviewer? If you were pairing what would you call out at the start of the call?
If there is something you're unsure about or need clarification on, then include your questions alongside the relevant topic.
)

Adding default `containerClassName` prop to `LoadingSkeleton` which sets `line-height: 1`

### Manual Test plan
[//]: # (
How did you test this manually? How can your reviewer test the workflow?
If this is a bug, how can you reproduce the bug?
)

- Start DS locally on this branch and verify that loading states with skeletons behave as expected (did that, but don't want to be including too many recordings since it's a public repo)


src/LoadingSkeleton/LoadingSkeleton.scss
```diff
@@ -8,6 +8,11 @@
   }
 }
 
+// Per https://github.com/dvtng/react-loading-skeleton/issues/23#issuecomment-939231878
+.loadingSkeletonContainer {
```
domschab23, line 12: Not adding display block or inline-block as it seems like we don't need it (plus it messes up some of our current usages of loading skeleton in other places like `Card` with default loading skeletons)

Before

https://github.com/user-interviews/ui-design-system/assets/54962889/013b1912-d362-4d52-8f84-c59a29425b36




After

https://github.com/user-interviews/ui-design-system/assets/54962889/f2ef8d79-6d07-401d-83d9-03f38070e509

The loading skeleton from a vid is getting a `height` prop with `38` value. In after video, you can see that the loading skeleton has the exact height of 38px. However, in before vid it still has 38px, but its container is slightly taller which results in more space being occupied than expected height prop. More explanation/context here https://github.com/dvtng/react-loading-skeleton/issues/23#issuecomment-939231878

src/LoadingSkeleton/LoadingSkeleton.scss
```diff
@@ -8,6 +8,11 @@
...
+  line-height: 1;
```
domschab23, line 13: Vid demonstrating height change when toggling this css prop

https://github.com/user-interviews/ui-design-system/assets/54962889/f97d9a3e-d27f-46a4-978b-01826bfbe3c1

src/LoadingSkeleton/LoadingSkeleton.jsx
```diff
@@ -8,10 +8,11 @@ import colors from '../Styles/colors/palette';
 
 import './LoadingSkeleton.scss';
 
-const LoadingSkeleton = ({ className, ...props }) => (
+const LoadingSkeleton = ({ className, containerClassName, ...props }) => (
   <SkeletonTheme baseColor={colors.UX_GRAY_300}>
     <Skeleton
       className={classNames('LoadingSkeleton', className)}
+      containerClassName={classNames('loadingSkeletonContainer', containerClassName)}
```
kyleshike, line 15: I sorta feel like we shouldn't expose containerClassName as a prop here but keep the fix.  In an ideal world, folks in in our private repo shouldn't have to worry about this inner container any longer:
```suggestion
const LoadingSkeleton = ({ className, ...props }) => (
  <SkeletonTheme baseColor={colors.UX_GRAY_300}>
    <Skeleton
      className={classNames('LoadingSkeleton', className)}
      containerClassName={'loadingSkeletonContainer'}
```

Curious what others think?

kyleshike (approved):
Approved as this clearly solves the issue, just one nit about exposing containerClassName in the props interface.

src/LoadingSkeleton/LoadingSkeleton.jsx
gabescarbrough: I agree with Kyle that it makes more sense to not expose `containerClassName` as a prop. I'd rather only expose what we need from `react-loading-skeleton`. If we were to ever want to switch to another third-party package for loading skeletons, we don't want to be locked into supporting their prop interface or bumping a major version.

domschab23: Well, we already expose that prop with `{...props}` (line 11 and 34 this file) so if we do 

```
containerClassName={'loadingSkeletonContainer'}
```

then you can still override the prop and you may unintentionally not use this default class. Plus, because we say that `containerClassName` by default is `undefined` on line 64, it looks like the 

```
containerClassName={'loadingSkeletonContainer'}
```

will actually never be applied (wasn't able to get the default applied locally with suggested change).

But if you do this

 ```
containerClassName={classNames('loadingSkeletonContainer', containerClassName)}
```

then you can just apply your containerClassName (if you need/want to) while keeping the default class. 

For components that are separate packages that are very lightly wrapped by our DS, I'd vote to keep flexibility with `{...props}` (or naming those props, but still the same idea). Otherwise, we can't use what's available to us.

Lastly, we already use that prop in RS in e.g. `app/javascript/researcher/participant_profile/components/project_workspace_sessions/components/session_edit_form.jsx:43`

Comment (domschab23) [reply src/LoadingSkeleton/LoadingSkeleton.jsx:15]: If you flipped the order to be

```
    <Skeleton
      className={classNames('LoadingSkeleton', className)}
      {...props}
      containerClassName="loadingSkeletonContainer"
    />
```

then this would work, but then again you're losing ability to use this prop even though we have advertised that prop to be working since the inception of this component. I think that this

```
containerClassName={classNames('loadingSkeletonContainer', containerClassName)}
```

is the best solution

gabescarbrough: All good points Dom

> Well, we already expose that prop with {...props} (line 11 and 34 this file)

You're right. Due to this we can't remove `containerClassName` as a prop without bumping a major version anyway. Like you said: they would be able to use it anyway, its in the propTypes, it is already used in RS.

> For components that are separate packages that are very lightly wrapped by our DS, I'd vote to keep flexibility with {...props} (or naming those props, but still the same idea).

I disagree with spreading props in a JS file since then we're accepting anything. That's a mistake we made that has locked us in to supporting all props of the libraries we wrap. Naming them or converting it to TS so they are restricted by the type system I agree with you.

> Otherwise, we can't use what's available to us.

That is intended. We should expose the props that we plan to support and encourage, not all the props that a third-party component has. Just because we're wrapping that component now, doesn't mean we'll use that third-party component forever. We should be free to replace third-party components as implementation details without affecting our own prop interface.

None of this is blocking for this PR. I'm talking about the "ideal world" Kyle's comment referred to and what I'd prefer in the future. As far as fixing the bug without introducing any breaking changes this PR nails it.

gabescarbrough (approved):

src/LoadingSkeleton/LoadingSkeleton.jsx
domschab23: > Otherwise, we can't use what's available to us.

> That is intended. We should expose the props that we plan to support and encourage, not all the props that a third-party component has. Just because we're wrapping that component now, doesn't mean we'll use that third-party component forever. We should be free to replace third-party components as implementation details without affecting our own prop interface.

If we wrap the 3rd party component like in this example and just pass all the props without really doing anything in between, then do we actually have our own prop interface for this component? Food for thought for the future ideal world. 

Seems like in those cases you want to be flexible (enabling using as many props as possible) since we're already making a decision to 'offload' this component completely to a 3rd party package. Plus, if you want to disable usage of certain number of props from a 3rd party component, then maybe it's a sign that we shouldn't pick that 3rd party dependency to begin with. 

I'm only mentioning this because this would make me worried that can we just slow each other down when we want/need to use a prop that's provided by a 3rd party.

kyleshike: Let's just remove `containerClassName` from the props for now then since it's already exposed to rails-server via the spread props. That way at the least we aren't exposing multiple ways to control the same attribute.

Once that's done I'll merge this in.

Regarding the discussion regarding what of 3rd party code we should or should not expose as props, like Gabe I'm of the opinion that the less is best.  Lets move that convo to slack :)

PR #1197, opened 2024-04-29
Converts most of the components to TS

kyleshike:
This excludes a few of our trickier components (popper, layout, selects) that will be handled individually.


kyleshike:
I did some initial QA:  the storybook builds just fine and rails server can _mostly_  compile.  There are a few errors over there that can and should be fixed there (like missing required props, etc).  That'll come once 3.0 is released and consumed

src/Container/Col.tsx
```diff
@@ -53,46 +105,41 @@ const columnPropTypes = PropTypes.oneOfType([
 ]);
 
 Col.propTypes = {
```
gabescarbrough, line 107: We should delete `Col.propTypes` right?

src/Container/Row.tsx
```diff
@@ -41,46 +93,41 @@ const rowColumns = PropTypes.oneOfType([
 ]);
 
 Row.propTypes = {
```
gabescarbrough, line 95: **chore**: 🧹

Delete `Row.propTypes`

src/ControlButtonGroup/ControlButtonGroup.tsx
```diff
@@ -1,26 +1,37 @@
...
 export const ORIENTATIONS = {
   COLUMN: 'column',
   ROW: 'row',
+} as const;
+
+type ControlButtonGroupProps = {
+  childChecked: (...args: unknown[]) => boolean;
+  children: React.ReactNode;
+  handleChangeValue: (...args: unknown[]) => unknown;
+  orientation?: 'column' | 'row';
```
gabescarbrough, line 12:
```suggestion
  orientation?:typeof ORIENTATIONS[keyof typeof ORIENTATIONS];
```

src/CopyToClipboardButton/CopyToClipboardButton.tsx
```diff
@@ -1,22 +1,33 @@
...
 export const ButtonVariants = {
   NEUTRAL: 'neutral',
   SECONDARY: 'secondary',
+} as const;
+
+type CopyToClipboardButtonProps = {
+  copyText?: string;
+  displayText?: string;
+  trackingEvent: string;
+  variant?: 'neutral' | 'secondary';
```
gabescarbrough, line 22:
```suggestion
  variant?: typeof ButtonVariants[keyof typeof ButtonVariants];
```

gabescarbrough (commented):
Still making my way through this. Here's the feedback I have so far.

src/Avatar/Avatar.tsx
```diff
@@ -1,12 +1,22 @@
...
 
 import './Avatar.scss';
 
-function Avatar(props) {
+type AvatarProps = {
+  ariaHidden?: boolean;
+  colorId?: string | number;
+  image?: string;
+  initials: string;
+  large?: boolean;
```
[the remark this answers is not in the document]
kyleshike: yea agreed.  I just used this component and was confused how to size it correctly

src/Dropdown/Dropdown.stories.tsx
```diff
@@ -59,8 +58,8 @@ export const Sizes = () => (
     </Dropdown>
     <br />
     <Dropdown>
-      <DropdownToggle size="md" variant="outline-primary">
-        Medium toggle
+      <DropdownToggle size="lg" variant="outline-primary">
+        Large toggle
```
gabescarbrough, line 62: Why was this switched from medium to large?

src/Dropdown/Dropdown.tsx
```diff
@@ -85,40 +45,58 @@ Dropdown.propTypes = {
...
+  className,
+  drop,
+  flip,
+  focusFirstItemOnShow,
+  navbar,
+  onSelect,
+  onToggle,
+  show,
+  bsPrefix,
+  ...props
```
gabescarbrough, line 81: Can't we remove the prop spreading here since we're extending the React Bootstrap props on line 65?

Or are there places where we're relying on arbitrary attributes being passed?

src/Dropdown/DropdownItem.tsx
```diff
@@ -1,26 +1,65 @@
...
+    HTML href attribute corresponding to a.href.
+   */
+  href?: string;
+  leadingIcon?: IconDefinition;
+  /**
+    Callback fired when the menu item is clicked.
+   */
+  onClick?: RBDropdownProps['onClick'];
+  children?: React.ReactNode;
+  variant?: 'danger'
```
gabescarbrough, line 47:
```suggestion
```

The danger variant was removed

src/Dropdown/DropdownMenu.tsx
```diff
@@ -1,105 +1,88 @@
...
-  onSelect: PropTypes.func,
+  rootCloseEvent?: 'click' | 'mousedown';
+  /**
+   Controls the visibility of the Dropdown menu.
+   */
+  show?: boolean;
+  /**
+   Menu color variant.
+   */
+  variant?: string;
```
gabescarbrough, line 51: **thought**: 💭

We should make this type more specific for better discoverability at some point

Doesn't look like `react-bootstrap` export a type for this unfortunately. They just use `string` too

src/FormControlLabel/FormControlLabel.tsx
```diff
@@ -1,10 +1,26 @@
...
-import 'scss/forms/form_control_label.scss';
+import './form_control_label.scss';
 
-const FormControlLabel = React.forwardRef(({
+type FormControlLabelProps = {
+  bordered?: boolean;
+  checked?: boolean;
+  className?: string;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  Control: React.ComponentType<any>;
```
gabescarbrough, line 11: **question (non-blocking)**: ❓

Why is any needed here?

src/FormGroup/FormGroup.tsx
```diff
@@ -0,0 +1,144 @@
...
+function renderErrors(errors) {
+  if (typeof errors === 'string') {
+    return errors;
+  }
+
+  return (
+    <ol className="FormGroup__invalid-feedback__list">
+      {
+        // eslint-disable-next-line react/no-array-index-key
+        errors.map((e, idx) => <li key={idx}>{e}</li>)
```
gabescarbrough, line 18:
```suggestion
        errors.map((e, idx) => <li key={`${idx}-${e}`}>{e}</li>)
```

src/IconButton/IconButton.tsx
```diff
@@ -54,11 +52,32 @@ export const IconButtonActions = {
...
+export type IconButtonProps = (
+  {
+    action: keyof typeof IconButtonActions;
+    ariaLabel?: string;
+    icon?: never;
+  } | {
+    action?: never;
+    ariaLabel: string;
+    icon: IconDefinition;
+  }
```
gabescarbrough, line 66: **question (non-blocking)**: ❓

What's up with this? I would have thought it would be

```ts
  {
    action: keyof typeof IconButtonActions;
    ariaLabel?: string;
    icon: IconDefinition;
  }
```

src/Modal/ModalHeader.tsx
```diff
@@ -1,16 +1,26 @@
...
 import { faExclamationTriangle } from '@fortawesome/pro-solid-svg-icons';
 
-export default class ModalHeader extends Component {
+type ModalHeaderProps = {
+  children?: React.ReactNode;
+  closingIsDisabled?: boolean;
+  subtitle?: string;
+  title?: React.ReactNode;
+  titleClass?: string;
+  titleId?: string;
```
gabescarbrough, line 14: I assume we're making this optional now on purpose. Kind of surprised it was required before given title is optional.

src/RadioButtonGroup/RadioButtonGroup.tsx
```diff
@@ -1,24 +1,32 @@
...
+  value?: number | string | boolean;
+  onChange?: (...args: unknown[]) => unknown;
+};
+
 export default function RadioButtonGroup({
   children,
   fullWidth,
-  name,
-  orientation,
+  orientation = 'column',
```
gabescarbrough, line 19:
```suggestion
  orientation = ORIENTATIONS.COLUMN,
```

src/Dropdown/Dropdown.stories.tsx
kyleshike: Our use of "md" for a lot of Bootstrap button sizes was previously invalid.  tl;dr is that medium is the default, and size can only be "sm" or "lg":

```ts
// node_modules/react-bootstrap/esm/Button.d.ts
export interface ButtonProps extends BaseButtonProps, Omit<BsPrefixProps, 'as'> {
    active?: boolean;
    variant?: ButtonVariant;
    size?: 'sm' | 'lg';
}
```

gabescarbrough (approved):
🔥 🔥 🔥 

If a suggestion makes the types stricter then I suggest we do it now

If a suggestion is minor then feel free to say we won't do it now and we can revisit. Better to have a record of little improvements for later than not.

src/Dropdown/Dropdown.stories.tsx
gabescarbrough: Interesting. That's not how I would have written that type. I'd have allowed `md` but made it optional. The more you know.

src/Dropdown/DropdownMenu.tsx
kyleshike: They actually have a type for it but it doesn't do anything:

```ts
// node_modules/react-bootstrap/esm/DropdownMenu.d.ts
export type DropdownMenuVariant = 'dark' | string;
```
Screenshot 2024-05-03 at 11 14 24 AM

src/FormControlLabel/FormControlLabel.tsx
kyleshike: In order to make this narrower, we'd have to provide the prop types for the Control component here in such a way that it is compatible with the existing uses.  I started trying to narrow it but after running into a few downstream type errors I switched it back to the any type.  

Definitely room for improvement here.  Probably falls into the general category of "fix our form elements"

src/FormGroup/FormGroup.tsx
kyleshike: looks like the eslint  doesn't want an index used _at all_ in keys: 
Screenshot 2024-05-03 at 11 21 37 AM

src/IconButton/IconButton.tsx
kyleshike: The way this component works, you can pass an action and (eg: add, edit, delete etc) and it will use the corresponding icon & aria label.  This means that 

```tsx
<IconButton action="CLOSE" />
<IconButton action="CLOSE" ariaLabel="Custom aria label" />
<IconButton icon={faCircle} ariaLabel="Another" />
```

src/Modal/ModalHeader.tsx
kyleshike: yea seemed a little strange

jasonbasuil (approved):
🚀 excited for all of this and thanks for leading the charge on getting this converted! I imagine there are components to cleanup with the conversions but overall I think we're on the right path 📈

PR #1202, opened 2024-05-01
Remove Sass vars represented as CSS Properties

kyleshike:
This PR replaces all Sass vars with their equivalent CSS property and removes the corresponding Sass var _unless_ it is in use by a Sass mixin argument, which doesn't play well with native CSS properties.


package.json
```diff
@@ -197,5 +197,6 @@
   "volta": {
     "node": "20.12.1",
     "yarn": "4.1.1"
-  }
+  },
+  "packageManager": "pnpm@9.0.4+sha1.b198ac6d38244fd829253720f9daafd6a606834d"
```
kyleshike, line 201:
```suggestion
  }
```

scss/forms/form_group.scss
```diff
@@ -10,8 +9,8 @@
   margin: 8px 0;
 
   &--bordered {
-    border: 1px solid $ux-gray-400;
-    border-radius: $ux-border-radius;
+    border: 1px solid var(--ux-gray-400);
+    border-radius: 4px;
```
kyleshike, line 13:
```suggestion
    border-radius: var(--ux-border-radius);
```

src/Table/Table.mdx
kyleshike, line 1: not quite sure how this got in, but this is a result of autoformatting in accordance with md styleguides

scss/colors/palette.scss
```diff
@@ -1,146 +1,39 @@
-// When updating this file, please export the color variable in variables.scss
-// so it is available in src/Styles/palette.js
+// DEPRECATED -- use css_properties.scss
+// The only reason these variables still exist is because they are currently in use by Sass mixins.
+// They can be iteratively removed as we continue to kill Sass.
```
jasonbasuil, line 3: 👍

jasonbasuil (approved):
Seems straightforward enough 🚀

gabescarbrough (approved):
🔥 

Not blocking to this PR at all but we should discuss which `ux` variables are supported by Synthesis (or need an equivalent) and move them over to the `synth` prefix

kyleshike:
> Not blocking to this PR at all but we should discuss which ux variables are supported by Synthesis (or need an equivalent) and move them over to the synth prefix

100%.  I feel like this can be one of the things we tackle next week

PR #1214, opened 2024-05-03
Dependency sweep

kyleshike:
Mainly tooling and eslint

Confirmed:
- local storybook build
- rails-server compiler works against this branch


package.json
```diff
@@ -197,5 +196,6 @@
   "volta": {
     "node": "20.12.1",
     "yarn": "4.2.1"
-  }
+  },
+  "packageManager": "pnpm@9.0.4+sha1.b198ac6d38244fd829253720f9daafd6a606834d"
```
jeffbax, line 200: is this expected?

I think this requires checking in `pnpm` doesn't it?

jeffbax (approved):

package.json
kyleshike: I don't know why my local keeps adding this to be honest.  Committed by accident

PR #1217, opened 2024-05-06
Fixes CheckboxButtonGroup sass imports

kyleshike:
Not sure how this flew under the radar


gabescarbrough (approved):
Ran `yarn storybook` and worked as expected

PR #1218, opened 2024-05-07
Update Tiptap to 2.3.1

gabescarbrough:
Updates Tiptap and its associated packages to the latest

This is a minor version bump so there aren't breaking changes in the package's API

Mostly getting bug fixes. Looking at their changelog there are some new features but I didn't see any changes to what we use.


src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -115,7 +115,7 @@ const RichTextEditor = forwardRef((
     placeholder,
     customExtensions = [],
   }: RichTextEditorProps,
-  ref: ForwardedRef<RichTextEditorRef> = null,
+  ref: ForwardedRef<RichTextEditorRef>,
```
gabescarbrough, line 118: This fixes a warning about `forwardRef` accepting exactly two params

This was the error:
```
Warning: forwardRef render functions accept exactly two parameters: props and ref. Did you forget to use the ref parameter?
```

src/RichTextEditor/RichTextEditor.test.tsx
```diff
@@ -12,17 +12,19 @@ describe('<RichTextEditor />', () => {
...
+
     expect(asFragment()).toMatchSnapshot();
   });
 
   describe('given an initial value', () => {
-    it('deserializes value correctly', () => {
+    it('deserializes value correctly', async () => {
       render(<Setup initialValue="<p>hello world</p>" />);
 
-      expect(screen.getByText('hello world')).toBeInTheDocument();
```
kyleshike, line 25: Will this have downstream implications in rails-server?

kyleshike (approved):

src/RichTextEditor/RichTextEditor.test.tsx
gabescarbrough: What kind of implications are you thinking? This change was just to get the test to not fail due to using `getByText` while the loading state was still showing.

gabescarbrough: I'm about to open a PR to update Tiptap on the `rails-server` side as well

kyleshike: Yea that's exactly what I mean.  This seems to imply that the editor loads async now, where previously that was not the case. If so, will this have any impact on specs in rails-server, or the rendered UI itself (eg: does it pop-in, have a loading state, etc) ?

gabescarbrough: RichTextEditor has [always had a loading state](https://github.com/user-interviews/ui-design-system/blob/develop/src/RichTextEditor/RichTextEditor.tsx#L202) while the `editor` initialized

I'm guessing that the newer version takes slightly longer to initialize so the test was checking too early. But the test really should have been async before.

Once this lands in a DS release, in the PR to update rails-server to use that new DS version there might be some tests to fix if they weren't awaiting the RichTextEditor loading

kyleshike: Okay cool.  Makes sense

PR #1219, opened 2024-05-09
Enable Dependabot Version Upgrades

jeffbax:
Allows Dependabot to open up PRs with library version upgrades, using a "widen" strategy by default which should be most compatible


kyleshike (approved):
nice

PR #1220, opened 2024-05-10
Add DateTimePicker

gabescarbrough:
Ports the `DateTimePicker` over from rails-server

Converted to TypeScript and simplified the props a little bit in the process

image


gabescarbrough:
Getting one TS error I haven't been able to figure out. Was able to find other folks with the same error but none of the solutions worked. @kyleshike any idea what is going on here?

```
Error: src/DateTimePicker/PickerEnforcedInput.tsx(36,9): error TS2322: Type 'ForwardedRef<HTMLElement>' is not assignable to type 'LegacyRef<HTMLInputElement> | undefined'.
  Type 'MutableRefObject<HTMLElement | null>' is not assignable to type 'LegacyRef<HTMLInputElement> | undefined'.
    Type 'MutableRefObject<HTMLElement | null>' is not assignable to type 'RefObject<HTMLInputElement>'.
      Types of property 'current' are incompatible.
        Type 'HTMLElement | null' is not assignable to type 'HTMLInputElement | null'.
          Type 'HTMLElement' is missing the following properties from type 'HTMLInputElement': accept, align, alt, autocomplete, and 53 more.
```

kyleshike:
Taking a look now!

src/DateTimePicker/PickerEnforcedInput.tsx
```diff
@@ -0,0 +1,51 @@
...
+type PickerEnforcedInputProps = {
+  disabled?: boolean;
+  inputClassName?: string;
+  name?: string;
+  startDate?: string;
+  value?: string;
+  onClick?: (...args: unknown[]) => unknown;
+};
+
+export const PickerEnforcedInput = forwardRef<HTMLElement, PickerEnforcedInputProps>(({
```
kyleshike, line 18: So the crux of this error is: 
```
Types of property 'current' are incompatible.
        Type 'HTMLElement | null' is not assignable to type 'HTMLInputElement | null'.
```
`'HTMLElement | null' is not assignable to type 'HTMLInputElement | null'.

Using a more specific html element as the generic property fixes the issue.  If the forwarded ref were a form, for example you'd want to do like `forwardRef<HTMLFormElement,`
```suggestion
export const PickerEnforcedInput = forwardRef<HTMLInputElement, PickerEnforcedInputProps>(({
```

package.json
```diff
@@ -19,8 +19,11 @@
     "@tiptap/pm": "^2.3.1",
     "@tiptap/react": "^2.3.1",
     "@tiptap/suggestion": "^2.3.1",
+    "date-fns": "^3.6.0",
+    "date-fns-tz": "^3.1.3",
```
kyleshike, line 23: should we put these in as peer deps as well?  We may want to lock this in btwn here & rails-server to prevent version

kyleshike (approved):
Looks good.  Just one q about deps and a few sass vars that can potentially use css properties

src/DateTimePicker/DateTimePicker.scss
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  .react-datepicker__triangle {
+    display: none;
+  }
+
+  // Override form-control's default greying of read only inputs
+  input:read-only, .form-control[readonly] {
+    background-color: var(--ux-white);
+  }
```
jasonbasuil, line 84: Suggestion: could we add an override to the bootstrap focus input here to match the other ones?

Screenshot 2024-05-15 at 9 36 43 AM

Screenshot 2024-05-15 at 9 37 00 AM

something like 
```
  .form-control:focus {
    box-shadow: none;
    border: 1px solid $input-focus-border-color;
  }
  ```

jasonbasuil (approved):
So good! Been wanting to get this one in for a while 🚀  thanks!

package.json
gabescarbrough: Good call

PR #1222, opened 2024-05-21
Adds borderedMultiValue prop to Creatable select

kyleshike:
This was added to the normal select, but we need it on the creatable select as well

### DIFF
--- a/src/Select/CreatableSelect.jsx
+++ b/src/Select/CreatableSelect.jsx
@@ -69,7 +69,7 @@ const CreatableSelect = ({
 CreatableSelect.propTypes = {
   'aria-label': PropTypes.string,
   'aria-labelledby': PropTypes.string,
-  borderedMultiValue: propTypes.bool,
+  borderedMultiValue: PropTypes.bool,
   className: PropTypes.string,
   components: PropTypes.any,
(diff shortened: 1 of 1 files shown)


kyleshike:
@jasonbasuil @gabescarbrough please review this today

src/Select/CreatableSelect.jsx
```diff
@@ -4,11 +4,14 @@ import PropTypes from 'prop-types';
...
 
-import { defaultTheme, defaultStyles, SELECT_SIZES } from './styles';
+import {
+ borderedMultiValueStyles, defaultTheme, defaultStyles, SELECT_SIZES,
+} from './styles';
 
 const CreatableSelect = ({
   'aria-label': ariaLabel,
   'aria-labelledby': ariaLabelledBy,
+  borderedMultiValue,
```
gabescarbrough, line 14: **thought**: 💭

I would have preferred `isBorderedMultiValue` to be more obvious it is a boolean, but since SingleSelect already uses this name I won't argue against consistency

gabescarbrough (approved):
🚢

jasonbasuil (approved):
🚀

PR #1223, opened 2024-05-22
Adds responsive FlexContainer component

kyleshike:
For each breakpoint xs, sm, md, lg, xl, xxl, we provide optional flex props. meaning that for any breakpoint, the following are configurable:
```
alignItems?: 'stretch' | 'center' | 'flex-start' | 'flex-end' | 'baseline' | 'initial' | 'inherit';
flexDirection?: 'column' | 'column-reverse' | 'row' | 'row-reverse';
flexWrap?: 'wrap' | 'no-wrap' | 'wrap-reverse';
justifyContent?: 'flex-start' | 'flex-end' | 'space-between' | 'space-around' | 'space-evenly' | 'center' | 'initial' | 'inherit'
gap?: 1 | 2 | 3 | 4 | 5;
width?: 'xs' | 'sm' | 'md' | 'lg';
```
This props are of course also just available to the component directly to facilitate the most common use case in which breakpoints aren't necessary at all.

### DIFF
--- a/src/FlexContainer/FlexContainer.module.css
+++ b/src/FlexContainer/FlexContainer.module.css
@@ -1,0 +1,767 @@
+.flexContainer {
+  display: flex;
+  max-width: 100%;
+}
+
+.alignItems_stretch {
+  align-items: stretch;
+}
+
+.alignItems_center {
+  align-items: center;
+}
+
+.alignItems_flex-start {
+  align-items: flex-start;
+}
+
+.alignItems_flex-end {
+  align-items: flex-end;
+}
+
+.alignItems_baseline {
+  align-items: baseline;
+}
+
+.alignItems_initial {
+  align-items: initial;
+}
+
+.alignItems_inherit {
+  align-items: inherit;
+}
+
+.flexDirection_column {
+  flex-direction: column;
+}
+
+.flexDirection_column-reverse {
+  flex-direction: column-reverse;
+}
+
+.flexDirection_row {
+  flex-direction: row;
+}
+
+.flexDirection_row-reverse {
+  flex-direction: row-reverse;
+}
+
+.flexWrap_wrap {
+  flex-wrap: wrap;
+}
+
+.flexWrap_no-wrap {
+  flex-wrap: no-wrap;
+}
+
+.flexWrap_wrap-reverse {
+  flex-wrap: wrap-reverse;
+}
+
+.justifyContent_flex-start {
+  justify-content: flex-start
+}
+
+.justifyContent_flex-end {
+  justify-content: flex-end
+}
+
+.justifyContent_space-between {
+  justify-content: space-between
+}
+
+.justifyContent_space-around {
+  justify-content: space-around
+}
+
+.justifyContent_space-evenly {
+  justify-content: space-evenly
+}
+
+.justifyContent_center {
+  justify-content: center
+}
+
+.justifyContent_initial {
+  justify-content: initial
+}
+
+.justifyContent_inherit {
+  justify-content: inherit
+}
+
+.gap_1 {
+  gap: var(--synth-spacing-1);
+}
+
+.gap_2 {
+  gap: var(--synth-spacing-2);
+}
+
+.gap_3 {
+  gap: var(--synth-spacing-3);
+}
+
+.gap_4 {
+  gap: var(--synth-spacing-4);
+}
+
+.gap_5 {
+  gap: var(--synth-spacing-5);
+}
+
+.width_xs {
+  max-width: 240px;
+}
+
+.width_sm {
+  max-width: 512px;
+}
+
+.width_md {
+  max-width: 640px;
+}
+
+.width_lg {
+  max-width: 896px;
+}
+
+/* https://getbootstrap.com/docs/5.0/layout/breakpoints/#max-width */
+/* X-Small devices (portrait phones, less than 576px) */
+@media (max-width: 575.98px) {
+  .xs_alignItems_stretch {
+    align-items: stretch;
+  }
+
+  .xs_alignItems_center {
+    align-items: center;
+  }
+
+  .xs_alignItems_flex-start {
+    align-items: flex-start;
+  }
+
+  .xs_alignItems_flex-end {
+    align-items: flex-end;
+  }
+
+  .xs_alignItems_baseline {
+    align-items: baseline;
+  }
+
+  .xs_alignItems_initial {
+    align-items: initial;
+  }
+
+  .xs_alignItems_inherit {
+    align-items: inherit;
+  }
+
+  .xs_flexDirection_column {
+    flex-direction: column;
+  }
+
+  .xs_flexDirection_column-reverse {
+    flex-direction: column-reverse;
+  }
+
+  .xs_flexDirection_row {
+    flex-direction: row;
+  }
+
+  .xs_flexDirection_row-reverse {
+    flex-direction: row-reverse;
+  }
+
+  .xs_flexWrap_wrap {
+    flex-wrap: wrap;
+  }
+
+  .xs_flexWrap_no {
+    flex-wrap: no-wrap;
+  }
+
+  .xs_flexWrap_wrap {
+    flex-wrap: wrap-reverse;
+  }
+
+  .xs_justifyContent_flex-start {
+    justify-content: flex-start
+  }
+
+  .xs_justifyContent_flex-end {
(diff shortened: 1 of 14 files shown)


kyleshike:
@jasonbasuil @gabescarbrough please review this today

gabescarbrough:
@kyleshike will do! Going to block out some time to check this out today

gabescarbrough:
Can you add the video you had in the last PR? It is 404ing there now

kyleshike:
Appreciate it.  I'd love to get 3.1 released today if possible as there a few things folks are waiting on, and I'd also love to get this into that release because it's awesome

kyleshike:
@gabescarbrough: https://www.loom.com/share/f960d1d3e15649eb8fcb3759895e9933?sid=ed463711-6cc0-454d-979e-52c46703c7de

src/FlexContainer/FlexContainer.module.css
```diff
@@ -0,0 +1,767 @@
...
+  gap: var(--synth-spacing-3);
+}
+
+.gap_4 {
+  gap: var(--synth-spacing-4);
+}
+
+.gap_5 {
+  gap: var(--synth-spacing-5);
+}
```
jasonbasuil, line 112: Not sure if this was discussed already with designers, but wondering if we'll need spacing higher than 5?

src/FlexContainer/FlexContainer.module.css
```diff
@@ -0,0 +1,767 @@
...
+  max-width: 512px;
+}
+
+.width_md {
+  max-width: 640px;
+}
+
+.width_lg {
+  max-width: 896px;
+}
```
jasonbasuil, line 128: is it possible to use any variables that we have for Card widths if we're using it here?

src/FlexContainer/FlexContainer.module.css
```diff
@@ -0,0 +1,767 @@
...
+    flex-direction: row-reverse;
+  }
+
+  .sm_flexWrap_wrap {
+    flex-wrap: wrap;
+  }
+
+  .sm_flexWrap_no {
+    flex-wrap: no-wrap;
+  }
```
jasonbasuil, line 310: ❓ suggestion: `sm_flexWrap_no-wrap`

jasonbasuil: goes for the other ones as well

src/FlexContainer/FlexContainer.module.css
jasonbasuil: also for the other other width values below

src/FlexContainer/FlexContainer.tsx
```diff
@@ -0,0 +1,60 @@
...
+    'flex-start' |
+    'flex-end' |
+    'space-between' |
+    'space-around' |
+    'space-evenly' |
+    'center' |
+    'initial' |
+    'inherit';
+  gap?: 1 | 2 | 3 | 4 | 5;
+  width?: 'xs' | 'sm' | 'md' | 'lg';
```
jasonbasuil, line 24: Could possibly add some comments on here to describe the props and show up in the docs:

```
/**
    Sets the gaps (gutters) between rows and columns using Synthesis spacing values
  */
  gap?: 1 | 2 | 3 | 4 | 5;
  /**
    Sets the width of the FlexContainer using similar values to Card sizes
  */
  width?: 'xs' | 'sm' | 'md' | 'lg';
```

src/FlexContainer/FlexContainer.tsx
```diff
@@ -0,0 +1,60 @@
...
+  md?: FlexProps;
+  lg?: FlexProps;
+  xl?: FlexProps;
+  xxl?: FlexProps;
+}
+
+export function FlexContainer({
+  children,
+  className,
+  ...props
```
jasonbasuil, line 41: ❓ our older `Flex` component has an `as` [prop](https://github.com/user-interviews/ui-design-system/blob/77c5d41dd0cb13b0cf4465701891189d83f8000a/src/Flex/Flex.tsx#L12). Would it be helpful to have that flexibility to define the element that is used as the `FlexContainer` as well (or a set of elements we want to allow)?

src/FlexContainer/index.ts
```diff
@@ -0,0 +1 @@
+export * from './FlexContainer';
```
jasonbasuil, line 1: Need to export from top level `index.ts` as well

jasonbasuil (commented):
Loving it! The gap property is slick and definitely an upgrade from the `Flex` component. A few comments and questions but looking good.

src/FlexContainer/FlexContainer.module.css
kyleshike: oops!

src/FlexContainer/FlexContainer.module.css
kyleshike: yea good call

src/FlexContainer/FlexContainer.module.css
kyleshike: I chatted w Kiley about this and did some grepping of rails-server.  It seems like this should be sufficient for now:
https://userinterviews.slack.com/archives/G01P02LNJTF/p1715010523934299

src/FlexContainer/FlexContainer.tsx
kyleshike: Looking at rails-server, that prop isn't used at all on the Flex component.  I think we should leave it out for now until we come upon a legit need for it

src/FlexContainer/index.ts
kyleshike: https://github.com/user-interviews/ui-design-system/pull/1223/commits/ca1ea95bff4d1b58f9e9429d0a374d1e8ca8166c

src/FlexContainer/FlexContainer.module.css
kyleshike: [ce479e4](https://github.com/user-interviews/ui-design-system/pull/1223/commits/ce479e494ca706151f0afd079f1ca91f3cfb40ea)

src/FlexContainer/FlexContainer.module.css
kyleshike: ce479e4

src/FlexContainer/FlexContainer.tsx
kyleshike: eabe050

src/FlexContainer/FlexContainer.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+export type FlexContainerProps = FlexProps & {
+  className?: string;
+  children: ReactNode;
+  xs?: FlexProps;
+  sm?: FlexProps;
+  md?: FlexProps;
+  lg?: FlexProps;
+  xl?: FlexProps;
+  xxl?: FlexProps;
```
gabescarbrough, line 41: **praise**: 🙌

This is a neat pattern

src/FlexContainer/FlexContainer.module.css
```diff
@@ -0,0 +1,767 @@
+.flexContainer {
+  display: flex;
+  max-width: 100%;
+}
+
```
gabescarbrough, line 5: **thought**: 💭

The classes below look a lot like Tailwind-style utility classes

An alternative approach would have been to add these utility classes as global styles and then any component could use the flex properties via `className`

I'm not advocating switching to that. You've wrapped this in a tidy component interface that is discoverable via TS types. Just pointing out that solving "generic styles" leads you to the same styles regardless of how you expose it to devs.

src/FlexContainer/FlexContainer.module.css
```diff
@@ -0,0 +1,767 @@
...
+.alignItems_stretch {
+  align-items: stretch;
+}
```
gabescarbrough, line 8: **nitpick**: 💅 

```suggestion
.align-items_stretch {
  align-items: stretch;
}
```

Should the class names be kebab-case like the CSS properties? That would be my instinct.

gabescarbrough (approved):
🚀 nice job. This is going to be a powerful tool to not have to write custom layout styles per component.

src/FlexContainer/FlexContainer.tsx
kyleshike: Right?  I'm really interested to see if this works out well.  If it does, then we should be able to do a similar thing to any variant that we want to support breakpoints

PR #1227, opened 2024-05-23
Remove Drawer--open when drawer unmounts

domschab23:


src/Drawer/Drawer.test.tsx
```diff
@@ -270,5 +286,19 @@ describe('Drawer', () => {
         });
       });
     });
+
+    describe('when user unmounts the drawer that is open (e.g navigation to a different page)', () => {
+      it('body tag does not have Drawer--open', async () => {
```
domschab23, line 291: This spec would previously fail if it was added

PR #1235, opened 2024-05-24
Fix default props

kyleshike:
I dropped a bunch of default props in the TS merge.  this puts them back


src/Input/Input.tsx
```diff
@@ -40,7 +40,7 @@ const Input = React.forwardRef<HTMLInputElement, InputProps>((props, ref) => {
     trailingIconOnClick,
```
jasonbasuil, line 40: a few `''` should be included?

```
Input.defaultProps = {
  disabled: false,
  leadingIcon: undefined,
  placeholder: '',
  trailingIcon: undefined,
  trailingIconLabel: '',
  trailingIconOnClick: undefined,
  trailingIconOnClickSubmit: false,
  trailingText: undefined,
  type: 'text',
  value: undefined,
  onChange: undefined,
};
```

src/LoadingOverlay/LoadingOverlay.tsx
```diff
@@ -23,7 +23,7 @@ const LoadingOverlay = ({
   header,
```
jasonbasuil, line 23: a few missing? 

```
LoadingOverlay.defaultProps = {
  contentCenterOverflow: false,
  contentTop: false,
  dataTestid: 'LoadingOverlay',
  header: undefined,
  text: undefined,
  textClassName: undefined,
  visible: true,
};
```

src/RadioButton/RadioButton.tsx
```diff
@@ -19,7 +19,7 @@ const RadioButton = React.forwardRef<HTMLInputElement, RadioButtonProps>(({
   className,
   disabled,
```
jasonbasuil, line 20: `disabled` ?
```
RadioButton.defaultProps = {
  checked: undefined,
  className: undefined,
  disabled: false,
  name: '',
  value: undefined,
  onChange: undefined,
};

jasonbasuil (approved):
just a few I think were missed? but looks good 👍

PR #1238, opened 2024-05-28
Don't include css properties in theme

kyleshike:


scss/theme.scss
```diff
@@ -3,7 +3,6 @@
 @import './bootstrap';
 @import './buttons';
 @import './cards';
-@import './css_properties.scss';
```
kyleshike, line 6: Including these styles here resulted in the properties being declared way too many times in the bundled output of our DS consumer.  More likely than not, the error is on the consumer end, but this resolves the immediate issue without needing to unpack a nontrivial amount of sass debt.

gabescarbrough (approved):

PR #1247, opened 2024-05-30
Set RichTextEditor value to empty string if editor is empty

domschab23:
More context here https://github.com/ueberdosis/tiptap/issues/154

Changing this to empty string if tiptap deems its content to be empty instead of saving `<p></p>` which could create some unexpected behavior when checking if string is empty 

primary: @kyleshike 

seconday: @gabescarbrough @jasonbasuil

### DIFF
--- a/src/RichTextEditor/RichTextEditor.tsx
+++ b/src/RichTextEditor/RichTextEditor.tsx
@@ -171,7 +171,7 @@ const RichTextEditor = forwardRef((
     extensions: editorExtensions,
     content: initialValue,
     onUpdate: ({ editor: ttEditor }) => {
-      const html = ttEditor.getHTML();
+      const html = ttEditor.isEmpty ? '' : ttEditor.getHTML();
 
       // if allowAttributes or allowedTags aren't passed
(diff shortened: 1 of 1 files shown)


src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -171,7 +171,7 @@ const RichTextEditor = forwardRef((
     extensions: editorExtensions,
     content: initialValue,
     onUpdate: ({ editor: ttEditor }) => {
-      const html = ttEditor.getHTML();
+      const html = ttEditor.isEmpty ? '' : ttEditor.getHTML();
```
domschab23, line 174: This would previously leave `<p></p>` which could get a little messy when saving it on the backend per https://github.com/ueberdosis/tiptap/issues/154

Now, we're changing this to empty string if we delete everything.

This will still save multiple line or spaces because tiptap does not deem to be empty (and I think I agree that this is a good approach)


https://github.com/user-interviews/ui-design-system/assets/54962889/8f3b174f-d3cf-4576-82fb-4407df38e874

kyleshike (approved):

PR #1252, opened 2024-06-05
Render disabled select option as disabled

jacklieblich:


savuserinterviews (approved):
looks great Jack!

src/Select/styles.js
```diff
@@ -93,14 +93,21 @@ const defaultStyles = ({ menuWidth, size }) => ({
...
         ...styles,
-        backgroundColor: isSelected || isFocused ?
-          systemColors.SYNTH_HOVER_STATE : styles.backgroundColor,
-        color: systemColors.UX_GRAY_900,
+        backgroundColor: getBackgroundColor(),
+        color: isDisabled ? systemColors.SYNTH_UNSELECTED_NEUTRAL : systemColors.UX_GRAY_900,
         fontWeight: fontWeights.light,
         fontSize: '0.875rem',
         cursor: 'pointer',
+        pointerEvents: isDisabled ? 'none' : 'auto',
```
kyleshike, line 110: Is it possible to have the disabled option have a default cursor (as opposed to `cursor: pointer`)?

Screenshot 2024-06-05 at 4 25 33 PM

jacklieblich: That screenshot is actually a weird case where the select is rendered within a button so it's inheriting the buttons cursor. This is how it was being rendered in storybook.

https://github.com/user-interviews/ui-design-system/assets/19777770/94838549-215e-4c03-9387-01215745d701

There was actually another issue there, where clicking the disabled option would select the previously hovered option. Anyway, latest commit fixes both:


https://github.com/user-interviews/ui-design-system/assets/19777770/0cd6ac7a-c526-41a7-90df-8a184c5f20ea

jacklieblich: also, the way I'm handling disabled now, is in line with how react-select does it. `isDisabled` controls `cursor` and `background-color`, although they set `cursor: not-allowed`.

https://github.com/JedWatson/react-select/blob/06e34882638d1526b9f5a1238bb567a3e9460ce5/storybook/stories/StyledMulti.stories.tsx#L17

kyleshike (approved):

PR #1256, opened 2024-06-21
Fixes active state for buttons

kyleshike:
Demo: https://www.loom.com/share/ef5b38bc922146d38ad3a94e1c3bd39d

### DIFF
--- a/scss/buttons.scss
+++ b/scss/buttons.scss
@@ -131,11 +131,11 @@ $primary: $ux-emerald-600;
     $btn-outline-primary-hover-border,
     $btn-outline-primary-hover-color,
   );
-  border-color: var(--btn-outline-primary-border);
+  border-color: $btn-outline-primary-border;
 
   &:active {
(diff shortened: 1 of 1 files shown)


scss/buttons.scss
```diff
@@ -131,11 +131,11 @@ $primary: $ux-emerald-600;
     $btn-outline-primary-hover-border,
     $btn-outline-primary-hover-color,
   );
-  border-color: var(--btn-outline-primary-border);
+  border-color: $btn-outline-primary-border;
```
kyleshike, line 134: When I converted these styles over to css variables, I was over-eager on some of them.  The changes in this file revert a subset of the variables that don't exist as css properties, but are provided the mixin as arguments.

It'd be great to reduce the complexity of these styles down to simple css again at some point, time permitting.

jasonbasuil (approved):
ooof. Good catch

PR #1258, opened 2024-06-28
Adds description to `Option`

tjbarber:
This PR allows you to use a description that's passed through the `data` prop so you get something that looks like this:

image

Since the `Option` component and its accessible props seem dictated by React Select, it looked like `data` was the best place to access this. Let me know if that's wrong and if you think there's a better way to do this.

### DIFF
--- a/src/Select/Option.jsx
+++ b/src/Select/Option.jsx
@@ -14,7 +14,7 @@ import './Option.scss';
 
 /* eslint-disable react/prop-types */
 const Option = forwardRef(({ indeterminate, ...props }, ref) => {
-  const description = props.data?.description;
+  const description = props.data.description;
   return (
     <components.Option {...props}>
       <div className="Option">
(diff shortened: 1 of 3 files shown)


jasonbasuil (commented):
I think this is the step in the right direction. 

It's probably worth noting that we have the `OptionWithDescription` component available (seen below). Perhaps it would be worth either making a checkbox available as a prop in that component or the other idea was to do another custom component like `OptionWithDescriptionAndCheckbox` for our use case which basically is a combination of what `Option` and `OptionWithDescription` are. 

Lmk if you have questions. Happy to help!

Screenshot 2024-07-02 at 9 17 53 AM

jasonbasuil (approved):
Overall looks good. I think just one more thing to consider is we probably want to be able to control for making the `label` either bold or not. We'll probably want the label to be bold if a description is present just to have that visual hierarchy (and to match the Figma design we're going for), otherwise it can just be regular weight as is.
Screenshot 2024-07-02 at 2 19 50 PM

PR #1261, opened 2024-07-02
noMargin prop for FormGroup

jasonbasuil:
add `noMargin` prop for `FormGroup`. Wasn't sure if `variant` made sense to use here or even `spacing="none"`. Open to suggestions 

Or maybe even make a `FormGroupContainer` that has spacing={ 1 | 2 | 3 etc.) and uses the `<FlexContainer>`

Screenshot 2024-07-02 at 3 02 27 PM

### DIFF
--- a/scss/forms/form_group.scss
+++ b/scss/forms/form_group.scss
@@ -54,6 +54,10 @@
     }
   }
 
+  &--no-margin {
+    margin: 0;
+  }
+
   &__helper-text {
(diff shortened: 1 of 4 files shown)


kyleshike (approved):
I think this does the job for now.  Love the idea of making a `FormGroupContainer` that controls a flex container.  We could probably even get away with setting our desired gap property for form groups and have the component accept no props

PR #1268, opened 2024-07-15
REC-25 add a new accordion alert component

magicmarie:
Adds a new component accordion alert.
Screenshot 2024-05-29 at 2 01 38 PM


jasonbasuil (approved):
Overall looks good for our use case! Just one suggestion on prop name. Also if you wouldn't mind adding a quick piece of documentation to `src/Accordion/Accordion.mdx`

PR #1271, opened 2024-07-22
move prepare-release steps into create-new-release

jasonbasuil:
closes REC-121

We're planning on switching back to `main` as our default branch for the Design System. This PR just moves some of the steps from `prepare-release` into `create-new-release` which should just create a PR for bumping the package.json version from `main`. This also should resolve the gnarly commit history that the previous gitflow process was plaguing us with and all the merge conflicts that usually needed to be resolved when merging develop into main.

So we should just be able to run the action `create-new-release` which will create a new package release and then just merging the version bump PR created.

To update the default branch, if all looks good, I just need to make the switch in settings for this repo.


kyleshike (approved):
Looks good.  I think this is a great choice

PR #1282, opened 2024-08-20
Update github pages workflow to only publish on release

Dchyk:
https://linear.app/user-interviews/issue/REC-296/update-create-new-release-to-handle-gh-pages-deploy-once-during

### DIFF
--- a/.github/workflows/gh-pages-deploy.yml
+++ b/.github/workflows/gh-pages-deploy.yml
@@ -1,9 +1,8 @@
 name: Deploy to github pages
 on:
-  workflow_dispatch:
-  push:
-    branches:
-      - main
+  release:
+    types:
+      - published
 
 # Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
 permissions:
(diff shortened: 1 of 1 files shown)


.github/workflows/gh-pages-deploy.yml
```diff
@@ -1,9 +1,7 @@
 name: Deploy to github pages
 on:
-  workflow_dispatch:
-  push:
-    branches:
-      - main
+  release:
+    types: [published]
```
Dchyk, line 4: I think this will do it ([source](https://github.com/user-interviews/ui-design-system/pull/1282/files)), but not sure how to test

kyleshike: the link points to this url lol

Dchyk: lol it's supposed to be this https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#release

jasonbasuil (approved):

kyleshike (approved):

PR #1286, opened 2024-09-05
Add support for arrays of active event keys to Accordion

Dchyk:
I noticed that adding the `alwaysOpen` prop to the Accordion caused the chevrons to stop rotating when the accordions were opened. This PR resolves that by allowing for arrays of `activeKeys` as well as single strings.

https://github.com/user-attachments/assets/34974478-7ff2-4b31-a1ba-1db4f4d73de1

### DIFF
--- a/src/Accordion/Accordion.stories.tsx
+++ b/src/Accordion/Accordion.stories.tsx
@@ -64,6 +64,48 @@ Default.args = {
   title: 'Accordion Toggle',
 };
 
+export function AlwaysOpen(args) {
+  return (
+    <Accordion alwaysOpen>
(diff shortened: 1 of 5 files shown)


Dchyk:
I'm also including some new misc styling options here for efficiency (added as props to the `AccordionToggle` specifically that Nat is requesting for use in the ODT Column Config component I'm working on - an always-there bottom border, and removal of most horizontal padding: 

https://github.com/user-attachments/assets/24359957-ecfe-4809-a1d9-827a72a00cfe

jasonbasuil (approved):

PR #1288, opened 2024-09-13
add deprecation warnings to components

jasonbasuil:
closes ENG-226

- Adds `useDeprecationWarning` hook and adds them to the following components under a Deprecated storybook folder
- Deprecated components: `Flex`, `Container`, `Col`, `Row`, `IconCell`
- Moves the `Layout` component under `Components` just for ease of discovery amongst all components until / if we want to organize the folders another way.
- Updates story sort to ensure all components shown are now alphabetical in Storybook nav.

Should only warn in dev environments
Screenshot 2024-09-17 at 9 29 34 AM

No longer have that funky non-alphabetical sort that was happening with the remaining JSX files
Screenshot 2024-09-17 at 9 46 52 AM

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -35,6 +35,7 @@ const preview = {
     },
     options: {
       storySort: {
+        method: 'alphabetical',
         order: ['Foundations', 'Components'],
       },
     },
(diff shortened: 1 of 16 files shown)


.storybook/preview.js
```diff
@@ -35,6 +35,7 @@ const preview = {
     },
     options: {
       storySort: {
+        method: 'alphabetical',
```
kyleshike, line 38: wow amazing it was that simple

kyleshike (approved):
Awesome thank you for doing this

PR #1289, opened 2024-09-24
Reduce bundle size fortawesome

domschab23:
I looked at our bundle sizes and I've noticed that our fortawesome is occupying 2 top spots for dependencies 

From prod

Screenshot 2024-09-24 at 4 55 07 PM

We're already guarding against it the rs repo, but looks like we never did that for DS

Locally before changes 
Screenshot 2024-09-24 at 4 55 52 PM

Locally after changes
Screenshot 2024-09-24 at 4 57 28 PM

Shedding off about 5.4 MB

### DIFF
--- a/src/Accordion/Accordion.stories.tsx
+++ b/src/Accordion/Accordion.stories.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 
-import { faCreditCard } from 'src/font_awesome/regular';
+import { faCreditCard } from '../font_awesome/regular';
 import Card from '../Card';
 
 import {
--- a/src/Accordion/Accordion.stories.tsx
+++ b/src/Accordion/Accordion.stories.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 
-import { faCreditCard } from '@fortawesome/pro-regular-svg-icons';
+import { faCreditCard } from 'src/font_awesome/regular';
 import Card from '../Card';
 
 import {
--- a/src/Container/Col.tsx
+++ b/src/Container/Col.tsx
@@ -4,7 +4,7 @@ import {
   type ColProps as ReactBootstrapColProps,
 } from 'react-bootstrap';
 
-import { useDeprecationWarning } from 'src/utils';
+import { useDeprecationWarning } from '../utils';
 
 export type ColProps = {
   /**
--- a/src/font_awesome/brands.ts
+++ b/src/font_awesome/brands.ts
@@ -6,17 +6,10 @@
 
 /* eslint-disable no-restricted-imports */
 
-import { library } from '@fortawesome/fontawesome-svg-core';
-
 import { faGoogle } from '@fortawesome/free-brands-svg-icons/faGoogle';
 import { faFacebook } from '@fortawesome/free-brands-svg-icons/faFacebook';
 import { faLinkedin } from '@fortawesome/free-brands-svg-icons/faLinkedin';
 import { faTwitter } from '@fortawesome/free-brands-svg-icons/faTwitter';
-
-library.add(faGoogle);
-library.add(faFacebook);
-library.add(faLinkedin);
-library.add(faTwitter);
 
 export {
   faGoogle,
--- a/src/font_awesome/brands.ts
+++ b/src/font_awesome/brands.ts
@@ -1,0 +1,18 @@
+import { library } from '@fortawesome/fontawesome-svg-core';
+
+import { faGoogle } from '@fortawesome/free-brands-svg-icons/faGoogle';
+import { faFacebook } from '@fortawesome/free-brands-svg-icons/faFacebook';
+import { faLinkedin } from '@fortawesome/free-brands-svg-icons/faLinkedin';
+import { faTwitter } from '@fortawesome/free-brands-svg-icons/faTwitter';
+
+library.add(faGoogle);
+library.add(faFacebook);
+library.add(faLinkedin);
+library.add(faTwitter);
+
+export {
+  faGoogle,
+  faFacebook,
+  faLinkedin,
+  faTwitter,
+};
--- a/src/font_awesome/brands.ts
+++ b/src/font_awesome/brands.ts
@@ -1,3 +1,11 @@
+// Keep this eslint-disable here. FontAwesome imports should only be happening
+// in this file and nowhere else in the codebase.
+//
+// Any direct import from '@fortawesome/free-brands-svg-icons' will create bloat in
+// our overall bundle size.
+
+/* eslint-disable no-restricted-imports */
+
 import { library } from '@fortawesome/fontawesome-svg-core';
 
 import { faGoogle } from '@fortawesome/free-brands-svg-icons/faGoogle';
--- a/src/font_awesome/regular.ts
+++ b/src/font_awesome/regular.ts
@@ -1,3 +1,11 @@
+// Keep this eslint-disable here. FontAwesome imports should only be happening
+// in this file and nowhere else in the codebase.
+//
+// Any direct import from '@fortawesome/pro-regular-svg-icons' will create bloat in
+// our overall bundle size.
+
+/* eslint-disable no-restricted-imports */
+
 import { library } from '@fortawesome/fontawesome-svg-core';
 
 import { faSpinnerThird } from '@fortawesome/pro-regular-svg-icons/faSpinnerThird';
--- a/src/font_awesome/regular.ts
+++ b/src/font_awesome/regular.ts
@@ -1,0 +1,98 @@
+import { library } from '@fortawesome/fontawesome-svg-core';
+
+import { faSpinnerThird } from '@fortawesome/pro-regular-svg-icons/faSpinnerThird';
+import { faCreditCard } from '@fortawesome/pro-regular-svg-icons/faCreditCard';
+import { faFileAlt } from '@fortawesome/pro-regular-svg-icons/faFileAlt';
+import { faCaretDown } from '@fortawesome/pro-regular-svg-icons/faCaretDown';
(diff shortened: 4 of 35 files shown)


src/Accordion/Accordion.stories.tsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 
-import { faCreditCard } from '@fortawesome/pro-regular-svg-icons';
+import { faCreditCard } from '../font_awesome/regular';
```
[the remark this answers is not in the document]
domschab23: We can look at absolute imports without reexporting this as a nice to have in the follow up

src/Container/Col.tsx
```diff
@@ -4,7 +4,7 @@ import {
   type ColProps as ReactBootstrapColProps,
 } from 'react-bootstrap';
 
-import { useDeprecationWarning } from 'src/utils';
+import { useDeprecationWarning } from '../utils';
```
domschab23, line 7: This actually failed for me as well locally so using relative (looks like a new addition that wasn't released yet to rs repo)
```
[webpack-cli] ModuleNotFoundError: Module not found: Error: Can't resolve 'src/utils' in '/Users/domschab/Work/rails-server/node_modules/@user-interviews/ui-design-system/lib/Container'
```

When building rs shakapacker

src/Accordion/Accordion.stories.tsx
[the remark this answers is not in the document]
domschab23: I think we can address the best course of actions for relative imports in a follow up PR; I was reading up about it and this may be tricky to get done. Adding alias in RS shakapacker config doesn't seem like a good solution

src/font_awesome/regular.ts
```diff
@@ -0,0 +1,104 @@
+// Keep this eslint-disable here. FontAwesome imports should only be happening
+// in this file and nowhere else in the codebase.
+//
+// Any direct import from '@fortawesome/pro-regular-svg-icons' will create bloat in
+// our overall bundle size.
+
+/* eslint-disable no-restricted-imports */
+
+import { library } from '@fortawesome/fontawesome-svg-core';
```
jeffbax, line 9: So the old app did this originally because we were still using `<i class='fa fa-fooo' />` elements and this would make things available outside of webpack context for us... I would expect tree shaking to take care of not importing things that were not used (if may be a bit slow)

Are we totally sure the prod bundle is affected here? 5.4mb is certainly massive savings so I don't want to undermine what could be a huge win... I just want to make sure the bad things we think are happening are happening.

The thing I would generally like to see if possible, is for the DS to provide FA to the rails app so they are always in sync and managed in one spot.

I don't know if this might be digging us deeper on a mechanism we don't want to rely on.... but maybe that is not the case.

I guess my other concern in the rails app was that these "libraries" didn't tend to get things removed when they were not used any longer. Which is why tree shaking is useful, it should get what people don't clean up.

jeffbax: All that said, being deliberate also might mean a faster build for us to not depend on tree shaking. I'm mostly just saying I'm not totally familiar with what has changed for FA in the past few years.

❓ Do we know tree-shaking is in fact turned on for DS?

src/font_awesome/brands.ts
```diff
@@ -0,0 +1,26 @@
+// Keep this eslint-disable here. FontAwesome imports should only be happening
+// in this file and nowhere else in the codebase.
+//
+// Any direct import from '@fortawesome/free-brands-svg-icons' will create bloat in
+// our overall bundle size.
+
+/* eslint-disable no-restricted-imports */
+
+import { library } from '@fortawesome/fontawesome-svg-core';
```
jeffbax, line 9: Is library needed if we are exporting things? In the past this was for putting things into the global scope... notes in another comment.

src/Accordion/Accordion.stories.tsx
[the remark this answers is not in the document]
jeffbax: is DS wired up to respect `ts` files? I'm assuming it is... but if it only had TSX until now then that could be why its not resolving

src/font_awesome/regular.ts
domschab23: Tree shaking is turned on for shakapacker when in production mode by default, but we're serving our DS components as a package that are transpiled with babel into commonJS and webpacker/shakapacker does not support tree shaking for commonJS modules https://webpack.js.org/guides/tree-shaking/#conclusion 

> Ensure no compilers transform your ES2015 module syntax into CommonJS modules (this is the default behavior of the popular Babel preset @babel/preset-env - see the [documentation](https://babeljs.io/docs/en/babel-preset-env#modules) for more details).

It's very likely that we're having unnecessary ~5MB of fortawesome in prod based on few tests I run

Shakapcker locally with prod setting with current DS from main
Shaka with prod mode with current DS

Shakapacker locally with prod setting and DS from this branch. 4.9MB less
shaka with prod setting wihth DS local

Bundles locally with prod setting and DS from this branch

Screenshot 2024-09-24 at 9 19 11 PM

Bundles locally with prod setting with current DS from main
Screenshot 2024-09-24 at 9 28 36 PM

You can see we're importing all of `@fortawesome` icons and that matches with screenshot from prod I added in PR description.

src/Accordion/Accordion.stories.tsx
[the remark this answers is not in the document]
domschab23: > is DS wired up to respect ts files?

I think so; we've had `.ts` files here for a while. Based on what I was reading, it might seem like webpack/shaka might also have be aware of the correct alias configuration for bundling/resolving modules (which that seems weird to add an alias in shaka config in rs repo). So it may not be enough to specify an alias in `.babelrc` file in DS repo. Happy to look more into this in a separate PR

src/font_awesome/regular.ts
domschab23: Font awesome also mentions this in their docs

https://docs.fontawesome.com/apis/javascript/tree-shaking

> Known Issues with Tree-shaking
Production Mode
Webpack only performs tree-shaking when the [--mode=production](https://webpack.js.org/concepts/mode/#mode-production) flag is used. This enables the UglifyJsPlugin which performs the optimization. To nerd out for a second; this plugin is actually doing “dead code elimination” which is a slower process than the canonical tree-shaking but the results are similar.
Rollup enables tree-shaking by default but it does so [only with ES modules](https://rollupjs.org/guide/en#tree-shaking). Font Awesome provides ES modules in all of its NPM packages so you should be good-to-go.

Which seems to confirm that currently we're not getting benefits of tree shaking for `@fortawesome` and that aligns with screenshot above comparing bundle sizes

src/font_awesome/brands.ts
domschab23: In general, using `library.add` seems like it registers icons globally as well so you don't need to import icons all the time and can instead do this

```
<FontAwesomeIcon icon="bullhorn" />
```

vs
```
import { faBullhorn } from '@fortawesome/pro-solid-svg-icons/faBullhorn';
.
.
.
<FontAwesomeIcon icon={faBullhorn} />
```

and that seems to nicely support the idea of minimizing the bundle size because only those icons that are added via library are included in the final bundle (assuming you're not having unnecessary imports elsewhere). Fine with deleting it and fine with keeping it as I don't have strong feelings if I like adding them to the lib or not.

src/font_awesome/regular.ts
jeffbax: I think my question is whether we are building the DS "in production" -- if we build it in production with tree shaking, then when it gets to the rails server it would presumably already have removed the unused icons?

src/font_awesome/brands.ts
jeffbax: I do not think we want them to be available globally, the first method makes it harder to understand whether the icon is used at all which gets me to my previous point that rails server probably has a ton of vestigial icons its including b/c nobody cleaned up library.

src/Accordion/Accordion.stories.tsx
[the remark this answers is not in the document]
jeffbax: is that an error with rails server, or building DS?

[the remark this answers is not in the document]
domschab23: rails server; there's no error building DS

src/font_awesome/regular.ts
domschab23: > I think my question is whether we are building the DS "in production" -- if we build it in production with tree shaking, then when it gets to the rails server it would presumably already have removed the unused icons?

It doesn't seem like that's what happens today per screenshots above (you can see icons taking much more space than needed in screenshots above).

When you transpile your code into commonJS (what we're doing today with DS components) then treeshaking for that portion of the code is not working per 

> webpacker/shakapacker does not support tree shaking for commonJS modules https://webpack.js.org/guides/tree-shaking/#conclusion

> Ensure no compilers transform your ES2015 module syntax into CommonJS modules (this is the default behavior of the popular Babel preset @babel/preset-env - see the [documentation](https://babeljs.io/docs/en/babel-preset-env#modules) for more details).

src/font_awesome/brands.ts
domschab23: Done

src/Container/Col.tsx
domschab23: cc @jeffbax 

(and this is example from a different PR from a couple of days ago that would make a rs build fail if we tried to merge newest DS to rs)

with relative imports to utils / icons in DS, RS build works. The problem here is defining aliases for internal of DS package that will also work with RS build. That could be looked at when we incorporate rollup as a follow on, but the PR in current states works with rs build.

jeffbax (approved):
lets take the short term win here if its working, but lets try to check out Rollup in the DS working group as I do think ideally we don't have to do this type of thing and its hopefully a straightforward change to package.json

PR #1296, opened 2024-10-14
Adds semibold font weight

kyleshike:


kyleshike:
@jasonbasuil any idea why CI might be failing here?

PR #1298, opened 2024-10-31
ControlButtonGroup do not set bordered if child is a DOM element

domschab23:
Fixing console warning that prevented us from adding more frontend specs (since warnings make specs fail)

before
Screenshot 2024-10-31 at 2 48 30 PM

After
Screenshot 2024-10-31 at 2 43 10 PM

### DIFF
--- a/src/ControlButtonGroup/ControlButtonGroup.tsx
+++ b/src/ControlButtonGroup/ControlButtonGroup.tsx
@@ -36,7 +36,9 @@ function ControlButtonGroup({
       onChange: handleChangeValue,
     } : {};
 
-    if (row && child.props.bordered === undefined) {
+    const isDOMElement = typeof child.type === 'string';
+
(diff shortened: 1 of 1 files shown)


src/ControlButtonGroup/ControlButtonGroup.tsx
```diff
@@ -36,7 +36,9 @@ function ControlButtonGroup({
       onChange: handleChangeValue,
     } : {};
 
-    if (row && child.props.bordered === undefined) {
+    const isDOMElement = typeof child.type === 'string';
+
+    if (!isDOMElement && row && child.props.bordered === undefined) {
```
domschab23, line 41: Problem here was that we tried to set `bordered` on any child which may not necessarily be a react element; In this case we tried setting this up for a div and hence this warning

frankhock (approved):

PR #1300, opened 2024-11-15
TDW: upgrade packages, remove unnecessary resolutions

jasonbasuil:
- ran `yarn upgrade interactive` and upgraded packages to latest 
- `css-loader` and `eslint` are the only remaining packages that need to be upgraded but weren't straightforward 
- removed resolutions that Yarn should be able to resolve now and not include version mismatches 🤞 
- removed the `Layout` component sub directory since we never used it and removed `react-router-dom` as a dependency


src/DateTimePicker/DateTimePicker.tsx
```diff
@@ -178,14 +178,14 @@ function DateTimePicker({
         adjustDateOnChange
         allowSameDay
         className={showPickerEnforcedInput ? '' : 'date-time-picker__input-group'}
-        customInput={showPickerEnforcedInput && (
+        customInput={showPickerEnforcedInput ? (
```
jasonbasuil, line 181: addresses this
Screenshot 2024-11-15 at 11 37 12 AM

package.json
```diff
@@ -130,18 +129,18 @@
     "babel-plugin-transform-react-remove-prop-types": "^0.4.24",
     "babel-preset-react": "^6.24.1",
     "bootstrap": "5.1.3",
-    "chromatic": "^6.24.1",
+    "chromatic": "^11.18.1",
     "classnames": "^2.5.1",
     "css-loader": "^6.11.0",
     "eslint": "^8.57.1",
```
jasonbasuil, line 135: wish I could get eslint 9 in here. Seems like @brianCollinsUI you have experience with this upgrade? 👀 https://github.com/user-interviews/rails-server/pull/27389

kyleshike (approved):
Nice

package.json
```diff
@@ -161,20 +160,14 @@
...
   "resolutions": {
-    "@babel/helper-create-regexp-features-plugin": "^7.22.15",
-    "cookie": "^0.7.0",
-    "express": "^4.21.0",
     "istanbul-lib-instrument": "^6.0.0",
-    "jackspeak": "2.1.1",
     "micromatch": "^4.0.8",
-    "normalize-package-data": "^3.0.0",
-    "path-to-regexp@npm:^1.7.0": "^1.9.0",
     "semver": "^7.5.3"
```
jeffbax, line 171: are these 3 still stuck? think these were ones I recently fixed (micromatch and semver at least)

package.json
```diff
@@ -19,19 +19,18 @@
     "@tiptap/pm": "^2.7.4",
     "@tiptap/react": "^2.7.4",
     "@tiptap/suggestion": "^2.7.4",
-    "date-fns": "^3.6.0",
+    "date-fns": "^4.1.0",
```
jeffbax, line 22: I think we'll have to update rails server to match these when we release the new DS

jeffbax (approved):
LGTM but will defer to Kyle

package.json
jasonbasuil: I can double check, but I tried removing these and ran `yarn why` for them and saw some packages resolved to versions lower than what's set as the resolution

jasonbasuil: I assume if these resolutions come from dependabot security updates, we could remove them all and see if any pop back up again? Not sure if that's the right approach tho?

jeffbax: no if they are resolving to lower then we still need these for sure

jeffbax: the older versions don't get patched for the fixes

package.json
brianCollinsUI: Yeah happy to help out here- is there an issue to track or can I make one? 👍

PR #1305, opened 2024-12-17
Add isWithinModal option to DateTimePicker

domschab23:
Without specifying `isWithinModal` or setting it to `false`

https://github.com/user-attachments/assets/78b43bfe-4d2b-4e8e-8689-5df24ca46482

With setting it to `true`

https://github.com/user-attachments/assets/e1f2fdda-a954-419e-9a76-e20310964046

it should behave in the same way as single select.

Took an inspiration from this recommended solution https://github.com/Hacker0x01/react-datepicker/issues/2545

### DIFF
--- a/src/DateTimePicker/DateTimePicker.scss
+++ b/src/DateTimePicker/DateTimePicker.scss
@@ -1,6 +1,10 @@
 @import 'react-datepicker/dist/react-datepicker.css';
 @import '../../scss/theme.scss';
 
+.react-datepicker__popper-container--modal {
+  z-index: var(--z-index-popover);
+}
+
 .date-time-picker {
   align-items: flex-start;
   display: flex;
@@ -26,13 +30,12 @@
     color: var(--ux-gray-300);
   }
 
-
   .date-time-picker {
     &__input-group {
       background-color: var(--ux-white);
       border-radius: var(--ux-border-radius);
       border: thin solid var(--ux-gray-400);
-      padding: .46875rem .75rem;
+      padding: 0.46875rem 0.75rem;
       justify-content: space-between;
       width: inherit;
     }
@@ -58,10 +61,10 @@
   }
 
   .react-datepicker__time-container
-  .react-datepicker__time
-  .react-datepicker__time-box
-  ul.react-datepicker__time-list
-  li.react-datepicker__time-list-item--selected {
+    .react-datepicker__time
+    .react-datepicker__time-box
+    ul.react-datepicker__time-list
+    li.react-datepicker__time-list-item--selected {
     background: var(--ux-blue-500);
   }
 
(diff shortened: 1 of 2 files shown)


src/DateTimePicker/DateTimePicker.scss
```diff
@@ -58,10 +61,10 @@
   }
 
   .react-datepicker__time-container
-  .react-datepicker__time
-  .react-datepicker__time-box
-  ul.react-datepicker__time-list
-  li.react-datepicker__time-list-item--selected {
+    .react-datepicker__time
```
domschab23, line 64: some lint fixes below; but nothing changed

src/DateTimePicker/DateTimePicker.tsx
```diff
@@ -195,8 +202,9 @@ function DateTimePicker({
         minDate={minDate}
         name={name}
         placeholderText={getDateFormat().toUpperCase()}
+        popperClassName={isWithinModal ? 'react-datepicker__popper-container--modal' : ''}
+        popperContainer={isWithinModal ? popperContainerDocumentBody : undefined}
         selected={dateFromString()}
-        showMonthDropdown={showMonthAndYearSelects}
```
DevStarks, line 199: Was this unused or incorrect?

src/DateTimePicker/DateTimePicker.tsx
```diff
@@ -32,6 +33,10 @@ const localeMap = {
 const STANDARD_TIME_FORMAT_FNS = 'hh:mm aa';
 const ISO_DATE_FORMAT_FNS = 'yyyy-MM-dd';
 
+const popperContainerDocumentBody = ({ children }: { children?: React.ReactNode }) => (
+  createPortal(children, document.body)
+);
```
DevStarks, line 38: I ran into a similar issue when rendering the button below in a modal. Setting overflow on the specific modal fixed it. Would that work here?

```
.ProjectSelectModal, .ProjectSelectModal :global(.ModalFooter) {
  overflow: visible;
}
```
IMO a component should avoid having parent specific props especially in the DS -- otherwise we'd end up with isWithinX, isWithinY, etc. I think fixing with css only would be great otherwise maybe renaming the prop to renderPopperInPortal or something more general? wdyt?

https://github.com/user-attachments/assets/4af45463-946b-423d-93c1-92f5b14906b5

src/DateTimePicker/DateTimePicker.tsx
domschab23: Whoopsie; bringing it back!

src/DateTimePicker/DateTimePicker.tsx
domschab23: This could also _probably_ be a default behavior for a date picker, but probably want to roll this out slowly to avoid any regressions. I can add a comment to look at making it a default separately so you don't have to specify the prop for modals

DevStarks (commented):
lgtm -- will let @kyleshike or @jasonbasuil or anyone else with strong DS opinions make the final call here!

jasonbasuil (approved):
👍

PR #1303, opened 2024-12-04
Fix Dependabot #124 `cross-spawn`, upgrade CSS-Loader@7

jeffbax:
Started fixing https://github.com/user-interviews/ui-design-system/security/dependabot/124

But kept pushing the rest of our dependencies and got everything to work except I didn't try the latest eslint given it is usually a huge pain in the butt.

PR #1307, opened 2024-12-19
Update date-fns and date-fns-tz peer deps

domschab23:
Was running into a peer dep warning when upgrading RS locally 
Screenshot 2024-12-19 at 12 13 41 PM


package.json
```diff
@@ -69,8 +69,8 @@
     "@popperjs/core": "^2.5.3",
     "bootstrap": "5.1",
     "classnames": "^2.2.5",
-    "date-fns": "^3.6.0",
-    "date-fns-tz": "^3.1.3",
+    "date-fns": "^4.1.0",
```
domschab23, line 72: We already specify these to be in these versions in "dependencies" but we didn't update the "peer-dependencies"

Going back to https://github.com/user-interviews/ui-design-system/pull/1300/files#r1844420644

jeffbax: you should probably update this in sync with the DS rather than just on its own

jeffbax: your PR is updating the DS so we should handle the dependencies and warnings on the same branch together IMO

domschab23: > you should probably update this in sync with the DS rather than just on its own

Aren't they already in sync? https://github.com/user-interviews/ui-design-system/blob/3accd22a9984436bd894cdaa2aa966997100ab8b/package.json#L22

We updated `date-fns` and `date-fns-tz` here in DS https://github.com/user-interviews/ui-design-system/pull/1300/files and here https://github.com/user-interviews/ui-design-system/pull/1303/files but didn't update peer deps for them. 

unless I don't follow then please rephrase. The warning from PR description can only happen on RS when you install npm deps on RS

jeffbax: I don't know that the DS version that changed the TZ gem ever got deployed to RS

That, or this did not get done as part of it https://github.com/user-interviews/ui-design-system/pull/1300/files#r1844420644

So the version of DS pulling in the new version of TZ should update the RS TZ at the same time that version of DS is added.

domschab23: It wasn't deployed.

I have a PR open now to deploy new DS in RS hence I wanted to update this peer dep in DS to be accurately reflecting what DS is expecting. 

I'm planning to update `date-fns` in RS (which is actually already updated in my PR in RS hence I stumbled upon this warning 
Screenshot 2024-12-19 at 12 13 41 PM so now coming back to the DS to update peer dep to be in sync with DS and RS dep. 

What am I missing? Should I follow a different update flow? If yes, then please explain what I should do step by step.

jeffbax: Yeah I get that part. But if you merge this TZ separately, it will be misaligned between the two for however much time right?

So you should upgrade TZ in tandem with RS in the same PR that upgrades the DS bringing along its upgraded version. Its a major upgrade, so it would be best to try to keep them aligned with an atomic commit rather than disjointed.

jeffbax: ok, my mistake -- I thought this was the RS 🥲

jeffbax (approved):

jeffbax:
just make sure the calendar components actually still work

PR #1364, opened 2025-12-12
Update Text to use CSS modules and clean up stories

bsaris-ut:
This updates how we use Text to use CSS modules and moves the documentation to reference the CSS variables intead.


jeffbax (approved):
lint error

jbaxendale-ut:
@bsaris-ut is this PR still relevant? do `synthesis` classes still apply? not sure where its landed, but its old and conflicted

PR #1506, opened 2026-05-26
Clean up SASS warning spam in Storybook build

jbaxendale-ut:
Ran this pre-release against Percy on a rails server branch, so pretty confident in this https://github.com/user-interviews/rails-server/actions/runs/26466519004

https://github.com/user-interviews/rails-server/pull/35264/changes

---

The current build output is full of spam related to sass and peer dependency warnings:

image

So this is addressing as many of the root causes as possible, with a final custom silencer for the boostrap/sass 3.0 upgrade warnings giving us clean output that should help make things more noticeable:

image

> • Here’s the accurate Sass-warning cleanup summary for this branch, ignoring the roundabout commit subjects:
> 
>   We cleaned up the remaining Storybook Sass noise by separating warnings we can fix from Bootstrap compatibility warnings we intentionally tolerate.
> 
>   First, we moved our own Sass toward the module system. Local SCSS imports were converted from @import to @use/@forward, and the button Sass was split into focused files: button-classes,
>   button-mixins, and button-variants. That let us avoid repeatedly pulling Bootstrap-heavy button Sass through broader theme imports while keeping public classes like btn-outline-
>   secondary and btn-link--neutral available.
> 
>   Second, we isolated Bootstrap Sass usage behind three small facade files: scss/bootstrap-variables.scss:1, scss/bootstrap-buttons.scss:1, and scss/bootstrap-breakpoints.scss:1. Those
>   files intentionally still use Bootstrap 5.1 @imports because Bootstrap’s Sass is not safely consumable through @use here, and upgrading Bootstrap has to be coordinated with the Rails
>   repo due to display risk.
> 
>   Third, Storybook’s Sass loader now uses quietDeps: true, which removes the flood of Bootstrap-internal Dart Sass deprecations without hiding warnings from our own code. We also fixed
>   our own newly visible Sass deprecations by replacing local darken() calls in button-variants.scss with sass:color / color.adjust.
> 
>   Finally, we added a very targeted Storybook ignoreWarnings filter for only the known Bootstrap facade @import deprecations. This does not silence every local @import warning; it only
>   suppresses warnings whose message points at the three Bootstrap facade files.
> 

---
<a href="https://app.blacksmith.sh/user-interviews/codesmith/ui-design-system/pr/1506"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782405766&installation_id=127447467&pr_number=1506&repository=user-interviews%2Fui-design-system&return_to=https%3A%2F%2Fgithub.com%2Fuser-interviews%2Fui-design-system%2Fpull%2F1506&signature=3477e7b06810ed97a3116f79cf478f4a066b484e736f25ac7856e281e7cbf6e3"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need. Autofix is disabled.</sup>


.storybook/webpack.config.js
```diff
@@ -1,6 +1,13 @@
 const path = require('path');
 const { codecovWebpackPlugin } = require('@codecov/webpack-plugin');
 
+const bootstrapSassFacadeImportWarningPattern =
```
jbaxendale-ut, line 4: we can't do anything about so I left a note and figure we know we have to deal with boostrap at some point anyway... this will help us spot others

scss/colors/inputs.scss
```diff
@@ -1,4 +1,4 @@
-@import './palette';
+@use './palette' as *;
```
jbaxendale-ut, line 1: basically swapping over to `@use` is the fix for most things

scss/button-classes.scss
```diff
@@ -0,0 +1,17 @@
+.btn-link--neutral {
```
jbaxendale-ut, line 1: basically because we were always importing "buttons" these classes were getting redefined over and over and over again, so just breaking them out into a single file

scss/theme.scss
```diff
@@ -1,13 +1,12 @@
-@import './avatars';
-@import './bootstrap';
-@import './buttons';
```
jbaxendale-ut, line 3: basically because we import "theme" everywhere, buttons were always needing to get recompiled in every file which is one of the original sins of our previously slow webpack build and bloated CSS output

(fixing this, we would probably have never had to nuke sass for the flat files a year or so ago)

scss/buttons.scss
```diff
@@ -1,414 +1,3 @@
-@import './colors';
```
jbaxendale-ut, line 1: this file was the scss devil

.storybook/webpack.config.js
```diff
@@ -25,7 +32,14 @@ module.exports = function ({ config }) {
             },
           },
         },
-        'sass-loader',
+        {
+          loader: 'sass-loader',
+          options: {
+            sassOptions: {
+              quietDeps: true,
```
jbaxendale-ut, line 39: This hides warnings emitted from within other dependencies that we can't control (namely, bootstrap's own sass warnings for its pinn'd bootstrap)

.storybook/webpack.config.js
```diff
@@ -35,5 +49,12 @@ module.exports = function ({ config }) {
     },
   );
 
+  config.ignoreWarnings = [
+    ...(config.ignoreWarnings || []),
+    // Keep other local @import deprecations visible; these three files are the
```
jbaxendale-ut, line 54: narrowly quiets down import warnings only when emitted from the known bootstrap warning

scss/button-classes.scss
jbaxendale-ut, line 1: double checking some things to make sure they are not still getting re-emitted over and over

src/CopyToClipboardButton/CopyToClipboardButton.scss
```diff
@@ -1,5 +1,4 @@
-@import '../../scss/theme';
-@import '../../scss/buttons';
+@use '../../scss/button-classes';
```
jbaxendale-ut, line 1: investigating whether this is needed

jbaxendale-ut (commented):
Going to cut a preview release to run in the RS in Cypress

scss/colors/inputs.scss
bcollins-ut: Image

interesting handle

bcollins-ut (approved):
lgtm - working locally as expected

scss/colors/inputs.scss
jbaxendale-ut: haha, need to wrap that in ``

PR #1651, opened 2026-08-25
Fix stale entities lockfile locator

jbaxendale-ut:
## Summary

Remove the stale `entities@^4.4.0` locator reintroduced when several Dependabot lockfile updates merged from overlapping base commits.

## Verification

- `yarn install --immutable --check-cache`
- `git diff --check`

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -6,6 +6,7 @@ name: "Chromatic build"
 # Event for the workflow
 on:
   pull_request:
+  merge_group:
   workflow_dispatch:
     inputs:
       dry_run:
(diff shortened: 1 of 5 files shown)


.github/workflows/chromatic.yml
```diff
@@ -6,6 +6,7 @@ name: "Chromatic build"
 # Event for the workflow
 on:
   pull_request:
+  merge_group:
```
jbaxendale-ut, line 9: going to make this repo use a merge group queue, that way rapid dependabot's won't trip on each other and cause busted lockfiles like this fix is for

dschab-ut (approved):

PR #1652, opened 2026-08-25
Validate merge queue entries in required workflows

jbaxendale-ut:
## Summary

- run each required GitHub Actions workflow for `merge_group` events
- keep Chromatic dry-run-only for synthetic queue commits

This is temporarily stacked on #1651 so the broken lockfile is repaired before merge-queue support lands. It will target `main` after #1651 merges.

## Verification

- `actionlint -config-file .github/actionlint.yaml`
- `zizmor --offline --min-severity high --no-progress .github`


bcollins-ut (commented):
👍

PR #1713, opened 2026-09-09
Bump nwsapi and move browser-dependent component coverage to Storybook

jbaxendale-ut:
## Summary

  - Update the transitive `nwsapi` dependency from 2.2.24 to 2.2.27, along with the related package refresh.
  - Move DateTimePicker interaction coverage that is unreliable in JSDOM into Storybook Chromium play tests.
  - Add hidden Chromium interaction suites for Drawer dismissal, expansion, lifecycle cleanup, multiple-drawer scroll
  locking, and Tabs keyboard/focus accessibility.
  - Keep public Drawer stories manual and interactive; document non-dismissible overlays and multiple mounted Drawers.
  - Fix Drawer body-scroll locking when multiple overlay-backed Drawers are open
  - Preserve and test the titleless Additional Actions example.
  - Correct Drawer breakpoint CSS ordering so `sm`/`md`/`lg` widths apply above their breakpoints; expanded Drawers
  remain 90% wide.
image
^ used to be busted, all stories would be effectively full screen


src/DateTimePicker/DateTimePicker.stories.tsx
```diff
@@ -1,24 +1,101 @@
 import React from 'react';
 
+import { expect, fn, within } from 'storybook/test';
```
jbaxendale-ut, line 3: these are new interaction stories, ported from the specs -- they help avoid the dependency update from colliding into an infinite loop with JSDOM, but are also more appropriate as stories than spec cases

src/Drawer/Drawer.interactions.stories.tsx
```diff
@@ -0,0 +1,203 @@
+import React, { useState } from 'react';
```
jbaxendale-ut, line 1: did a sweep of a few drawer and tab stories to also cut over

src/Drawer/Drawer.scss
```diff
@@ -54,6 +36,24 @@
     padding-top: 48px;
   }
 
+  @media (min-width: $drawer-width-small) {
```
jbaxendale-ut, line 39: this fixes the stories themselves, from being all forced to full width due to media order being wrong (this will take testing the UI app a bit, workspace and cypress)

src/Drawer/Drawer.tsx
```diff
@@ -30,6 +30,12 @@ export const DrawerSizes = {
   LARGE: 'lg',
 } as const;
 
+const openDrawerIds = new Set<symbol>();
```
jbaxendale-ut, line 33: this is a bugfix, where multiple drawers were not tracking their open state and overlay properly

src/Tabs/Tabs.interactions.stories.tsx
```diff
@@ -0,0 +1,56 @@
+import { expect } from 'storybook/test';
```
jbaxendale-ut: just some missing spec/story cases codex filled out

package.json
```diff
@@ -2,23 +2,23 @@
   "name": "@user-interviews/ui-design-system",
   "version": "8.7.2",
   "dependencies": {
-    "@tiptap/core": "^2.27.2",
```
jbaxendale-ut, line 5: bumped everything for what is hopefully the final 8.x release

src/Drawer/Drawer.tsx
dschab-ut: Kinda surprised that we have multiple drawers anywhere that can be open at the same time 🤔

dschab-ut (approved):

jbaxendale-ut:
Doing rails app regression

jbaxendale-ut:
Breakpoints check out as OK

Screenshot 2026-09-09 at 4 10 08 PM

Screenshot 2026-09-09 at 4 13 23 PM

